PR #578, opened 2013-09-21
Added a module to configure behavior when clicking an entry (EntryTitleAction)

ericsubach:
Configure the behavior when clicking the entry's title.

### Options
* Action
  - open the content (reddit's default action)
  - open the comments
* Focus
  - open in foreground
  - open in background

### Testing
Fully tested on Firefox and Chrome. I appreciate any feedback for about its behavior on other browsers.

### Other
* This is similar to the SingleClickOpener module, but I felt it was different enough to warrant its own module.
* The module is disabled by default because **it changes reddit's default behavior**.
* My motivation for making this is the desire to only open the comments (since I can get to the content from there) and to have interaction remain as close to the normal flow as possible (left-clicking on the main link).

### Disclaimer
I am not a web developer and did a lot of copy-pasting based on the existing SingleClickOpener module.

### DIFF
--- a/Chrome/background.js
+++ b/Chrome/background.js
@@ -144,11 +144,11 @@ chrome.extension.onMessage.addListener(
 				}
 				sendResponse({status: "success"});
 				break;
-			case 'mainLinkConfigure':
+			case 'entryTitleAction':
 				var focus = !(request.isBackground);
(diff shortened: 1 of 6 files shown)


mc10:
Closing this (now replaced) PR.

PR #981, opened 2014-04-24
Add demo link in option

matheod:
This is not a real PR, but more a suggestion (which can be declined).

The idea is to allow user to get demo of each option/module. This come in two ways : 
##### Demo link on console

Each module and option can have a `demo` attribute, specify it will add a demo link at the right of the option allowing the user to see the option.
resdemo
##### Demo link on wiki

Demo link will be added on the wiki, and Res will automatically transform them into image (not link) if they are in a correct form (the url must end with `?show=1`, plus an other condition specified just after). User without Res will just see a link.
#### Where will images be stored ?

In order to allow image to be see both in Res and in the wiki, the image can't be stored directly into Res, moreover it will avoid to increase extension size with lots of image. Using reddit upload is not a solution, because there is a limit of 50 images (and a limit size). Using css sprite would lead to force to load all the image just to see one. But these both last problem are not the main problem, the main problem is it's not easy to add a image, because user doesn't have right. Here is my solution : Using github to store image in /`img/demo`.
Why is this good ? Because user can add image with PR which is easier, and more logical, if a user add a demo link to an option/module, it will have to make a pull request, so it can also add directly the image to the PR. About wiki image, image will be only show if they meet this three conditions :
- They come from this github repository in the `/img/demo` folder
- We are on the /r/Enhancement wiki
- The url end with `?show=1`

Wiki example : http://www.reddit.com/r/Enhancement/wiki/demoofdemo


erikdesjardins:
Closing for cleanup

PR #1382, opened 2014-07-04
Fixes for the cog/gear settings dropdown menu (inc. pinHeader)

githue:
This was reverted just before 4.5.0.0 because it didn't work with pinHeader, this new version does.

I've tested a number of subreddits with wild layouts and couldn't find any problems with it. If you can think of anything this could interfere with, namely where `#header-bottom-right` is set to `position: fixed`, then let me know and I can test it. Otherwise this is ready.


githue:
No longer needed.

PR #1506, opened 2014-08-13
Added Image Shrink module

mjkaufer:
Basically, the image shrink module lets users choose a maximum image size for imgur images if they don't want to have to load extremely large images. For instance, a 5000px wide image of a few words doesn't need to be loaded when it can be seen perfectly fine at 1024px. It uses imgur's thumbnails, specified here: https://api.imgur.com/models/image#thumbs. Whenever an image is shrunken, a link to the original image is added to the post as well.


erikdesjardins:
#1520
(cleanup)

PR #1519, opened 2014-08-18
Add support for uploading files on the submit page

SirCmpwn:
We [discussed this](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/770) in February, but I never put in the effort to make it properly modular. And now, here it is, ready to go. I'll repeat some information from a while ago for the sake of this pull request.

First of all, it looks like this: https://mediacru.sh/KkRwHnh-fH_r

This pull request adds the ability to upload files to MediaCrush directly from the submit page. It is modular and should allow for other hosts to come in and add support for their own needs. Take note of [lib/modules/hosts/mediacrush.js](https://github.com/SirCmpwn/Reddit-Enhancement-Suite/blob/master/lib/modules/hosts/mediacrush.js) to see how easily other hosts can get themselves added (they also need to add a little bit of code to [lib/modules/upload.js](https://github.com/SirCmpwn/Reddit-Enhancement-Suite/blob/master/lib/modules/upload.js#L18-L28)).

This pull request has been tested in Firefox and Chrome.

The settings should allow users to switch between hosts when someone else gets around to adding themselves:

We will have to sort something out to select a fair default when the time comes (or put together some UI to make users choose for themselves).

I have evaluated the following hosts:

## MediaCrush

Pros:

1. No API key to upload
2. Rate limited per user, and very generously
3. Highest file size limit (50 MB)
4. Open source and privacy stuff (also forced https)
5. Supports images, audio, video, pretty much all formats
6. Speedy

Cons:

1. This module was written by a MediaCrush dev
2. Not as popular as alternatives
3. Limited support for user accounts

## Imgur

Pros:

1. Ubiquitous across Reddit
2. Speedy
3. HTTPS support (not by default)

Cons:

1. No uploading without an API key (or breaking their terms)
2. Rate limited *per-application*, rather than per-user, so all RES users would be under the same rate limit
3. Pushing their own social network to compete with Reddit
4. Coming under fire for spam problems
5. Reduces quality of uploaded files
6. Has been redirecting users from direct links in a supremely shady manner

## Minus

Pros:

1. Larger file size limit than imgur
2. Lossless compression
3. HTTPS (not by default, triggers mixed content warnings)

Cons:

1. Weird private API
2. Weird social chat app shoehorning
3. Infamous for being slow

## gfycat

Only supports GIFs, not acceptable as a general-purpose host. Disregarded.

- - -

The only one of these I implemented was MediaCrush (of course) because the others have serious concerns from an API perspective.

Will draft procedure for adding new hosts after some peer review.

Open questions:

1. How to support albums and is that desirable?
2. How to notify other hosts that this is happening and how much time to give them to get in?
3. Does this show too favorably upon MediaCrush?

For now, I would say that we can live without albums (making that work in a host-ambiguous way would be a nightmare anyway), and regarding #2, my clearly biased opinion is that we should ship this and let them come add themselves later. As for #3, that's up to individual opinion. If the answer is "yes", then I think MediaCrush is a worthy host to favor.


lib/modules/upload.js
```diff
@@ -0,0 +1,113 @@
...
+				name: host.description,
+				value: hostKey
+			});
+		}
+
+		RESLoadResourceAsText('modules/hosts/hosts.json', function(str) {
+			var hosts = JSON.parse(str);
+			modules['uploader'].options['preferredHost'].values.sort(function(host1, host2) {
+				var hi1 = hosts.indexOf(host1.value);
+				var hi2 = hosts.indexOf(host1.value);
```
JordanMilne: copy / paste error?

SirCmpwn: Thanks!

andytuba: whoops. thanks jordan.

SirCmpwn:
Note: merge conflicts resolved.

SirCmpwn:
:tada:

PR #1520, opened 2014-08-18
Support for shrinking images

mjkaufer:
Added an option to Inline Image Viewer (showImages.js) to allow for shrinking of imgur images.

A modification of this previous pull request: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/1506

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -651,6 +651,22 @@ modules['showImages'] = {
 					console.error.apply(console, arguments);
 				});
 			}
+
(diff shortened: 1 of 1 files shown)


andytuba:
Imgur hashes will always be an odd number length (5 or 7, in the future 9) so we can strip the even-count suffix (which would indicate a specific size).  IMO it would be good to default to to the `g` suffix to preserve bandwidth for user and imgur, and autoload fullsize when dragging to resize up or if maxsize is a large enough value or if user has enabled "always load full reso"

PR #1695, opened 2014-10-21
Change cursor style

benwa:
Align image button's cursor style with Reddit's `.expando-button`.


andytuba:
now they are uniform :dancers:

PR #1738, opened 2014-10-31
Load new tabs in foreground if browser.tabs.loadInBackground == false

jspenguin:
This has been broken for a while, and I finally figured out why: inBackground always gets set to true due to "|| true", and it also gets set in the 'keyboardNav' case.

### DIFF
--- a/XPI/lib/main.js
+++ b/XPI/lib/main.js
@@ -233,7 +233,7 @@ pageMod.PageMod({
 		});
 		worker.on('message', function(data) {
 			let request = data,
-				inBackground = prefs.getBoolPref('browser.tabs.loadInBackground') || true,
+				inBackground = prefs.getBoolPref('browser.tabs.loadInBackground'),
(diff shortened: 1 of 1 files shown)


andytuba:
@mc10, do you remember why you added `|| true`?

mc10:
I believe it was to set `inBackground` to a boolean if it wasn't one, but my code is buggy for sure. Just dropping the `|| true` is fine.

mc10:
I'm not that sure about the `inBackground` removal; that code is necessary for the RES setting to "open in a new tab" to work.

andytuba:
> `inBackground`

sure looks like it. 
https://github.com/jspenguin/Reddit-Enhancement-Suite/blob/master/lib/modules/keyboardNav.js#L1476-L1481

I can imagine the line @jspenguin removed looked like copy-pasta error, but it's just that `button` is a bad property name.  It's a bit more effort to replace that with something better like "focus", though.

Issue #1562, opened 2014-08-29
Give noCtrlF treatment to more buttons


andytuba:
in moderation, though.

PR #1750, opened 2014-11-03
Make more elements not show up in searches (noCtrlF)

erikdesjardins:
As suggested in #1562, this handles comment navigator buttons (no performance impact) and reddit's comment/link buttons (slight performance impact, so disabled by default).

It's not very feasible to apply this to submission times (as was suggested), since they are dynamically updated.


mc10:
This looks fine to me.

PR #1785, opened 2014-11-16
Dailymotion expando support

### LINKED ISSUE #1096: Dailymotion expando support
~~http://www.reddit.com/r/Enhancement/comments/25tbzp/feature_request_expando_support_for_the_still/~~

erikdesjardins:
Resolves #1096.

These expandos work great, *but* they produce about a million [mixed-content warnings and errors](https://cloud.githubusercontent.com/assets/7673145/5063324/ace7bba2-6daf-11e4-814e-b653ddbfba26.PNG) (no loss of functionality, just logging scripts and other unimportant stuff).

### DIFF
--- a/chrome/manifest.json
+++ b/chrome/manifest.json
@@ -111,6 +111,7 @@
 				"modules/quickMessage.js",
 				"modules/hosts/imgur.js",
 				"modules/hosts/xkcd.js",
+				"modules/hosts/dailymotion.js",
 				"modules/hosts/pornbot.js",
 				"modules/hosts/coub.js",
 				"modules/hosts/uploadly.js",
--- a/firefox/index.js
+++ b/firefox/index.js
@@ -565,6 +565,7 @@ PageMod({
 		self.data.url('modules/quickMessage.js'),
 		self.data.url('modules/hosts/imgur.js'),
 		self.data.url('modules/hosts/xkcd.js'),
+		self.data.url('modules/hosts/dailymotion.js'),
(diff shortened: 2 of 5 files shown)


andytuba:
Heck with it, I can set up an AutoMod rule that triggers on "dailymotion [errors|issues|warnings|https]" and explains the situation.

andytuba:
@NykO18 Hmm, I'm not sure what to ask for without knowing the implementation challenges.  

It looks like www.dailymotion.com, logger...dailymotion.com, and dmcdn.net already support ssl, but sometimes things get linked via http://.

I also see a lot of errors from third-party vendors (doubleclick, liverail, etc.) -- they ought to support https, but fixing and testing all of them will take a few hours.

Any low-hanging fruit on that front?

andytuba:
Another alternative (which I'd rather avoid) is that RES provide the video player and dailymotion just provides the URL directly to the content (mp4/aac).  I'd rather avoid that, since it would add maintenance cost to RES and remove dailymotion branding and value-adds.

andytuba:
I think a better middle ground is to add some sort of warning in the settings or expando button title when browsing reddit on https.

NykO18:
I agree, because the content URLs (mp4/aac) can't be retrieved and shared publicly since they're signed for the current user and you can only retrieve your own streams using the API anyway. I wish I could speed up the process on this, but I'm not working on this part of the project.
Realistically, what are you trying to embed in the expando? Which URL/page? Can't you use the embed like so: http://www.dailymotion.com/swf/video/x2fx37n

PR #1804, opened 2014-11-23
NSFW expando concept feature

agentlame:
It's a simple feature: different colored expandos for NSFW links.  While they already have warning flair and (sometimes) thumbnails, I still find myself clicking on NSFW images without forethought.  I know I've seen many comments about people doing this over the years, so it seems like a reasonable feature.

This is rather quick hack for just image links and using an exceptionally ugly sprite.  I wanted to get some feedback and gauge interest before investing too much time on it.

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -1151,7 +1151,7 @@ img {
 	vertical-align: bottom;
 	margin-left: 4px;
 }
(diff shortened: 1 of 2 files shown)


andytuba:
diagonal candy stripes it is.

PR #1837, opened 2014-12-06
Contexts: allow certain functionality under certain conditions

andytuba:
Certain functionality, such as filtering and night mode, are only active in certain contexts: the user is visiting a certain subreddit, frontpage, or all; a post/comment is from a particular subreddit or user; the time of day falls within a certain range, etc.  This PR is the stub of a utility module for centrally managing these contexts in order to address requests like "filter certain subreddits during spoiler-heavy days of the week" as well as adding requested functionality like "toggle switch for certain subreddits or filters".

---

If a context is active, the following must all be true:

1. that context has **at least one** associated toggle enabled.
1. FUTURE: the context is within the scheduled time for **all** associated schedules

---

Toggles can be associated with a new menu item in the dropdown. If any toggle is enabled, then that menu item is marked "active/on". 

---

I've only implemented "context" on keyword filters so far. Before I carry on with implementing it on other features (such as subreddit filters), I was hoping for feedback on the architecture and names (especially "contexts").

---

screen shot 2014-12-06 at 11 11 38 am
screen shot 2014-12-06 at 11 12 26 am
screen shot 2014-12-06 at 11 10 43 am

### DIFF
--- a/Chrome/manifest.json
+++ b/Chrome/manifest.json
@@ -48,6 +48,7 @@
 				"modules/keyboardNav.js",
 				"modules/about.js",
 				"modules/commandLine.js",
+				"modules/contexts.js",
 				"modules/hover.js",
 				"modules/subredditTagger.js",
 				"modules/singleClick.js",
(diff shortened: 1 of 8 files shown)


andytuba:
superseded by #1860 

don't care about the tri-state

PR #1844, opened 2014-12-08
Lower RES Memory usage!

TheRook:
I got a new Microsoft Surface,  and the memory usage on RES is too much for the poor tablet.  So,  I decided to fork RES and see if I can reduce memory overhead for loading new pages.   If this patch goes well,  I'd like to further reduce RES's memory usage. 

This patch could use more testing.  Loading 10 pages with the current RES takes roughly 804M or ram,  after this patch the page consumed 764M with the same dataset.  Still needs more work,  but there is a noticeable reduction in memory usage with normal browsing habits.

Improvements:
- A bloomfilter is the most efficient way of keeping track of dupes.  A bloomfilter doesn't need to store the full string in order to keep track of uniqueness.  Additionally,  insertion into a bloom filter is always O(1), where as dictionaries need to be expanded when they become full.
-  "Hiding" elements with CSS doesn't make them go away.  These elements need to be deleted properly to free up memory. Hiding elements lead to a **nasty** memory leak.  Reddit has enough re-posts,  there is no need for RES to introduce duplicates to the feed.

### DIFF
--- a/Chrome/manifest.json
+++ b/Chrome/manifest.json
@@ -27,6 +27,7 @@
 				"vendor/jquery.tokeninput.js",
 				"vendor/HTMLPasteurizer.js",
 				"vendor/snuownd.js",
+				"vendor/bloomfilter.js",
 				"core/utils.js",
 				"browsersupport.js",
 				"browsersupport-chrome.js",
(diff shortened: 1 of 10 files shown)


andytuba:
Wow, sorry this got orphaned for so long. Should we resurrect it?

andytuba:
The hideDupes: remove and bloom filter aspects of this PR are probably still worth resurrecting.

erikdesjardins:
:-1: on bloom filter unless the existing map/set shows up in a heap or CPU profile.

Even if it were 10x faster, map/set lookups are such a small fraction of our CPU time that the difference won't justify adding another library.

TheRook:
@andy Tuba - There is no doubt that these two changes will reduce memory
usage over many page loads.

PR #1852, opened 2014-12-10
Inline expansion for Dropbox "shared" images

baskervald:
Adds a handler for dropbox "shared" image urls.

Requires no requests other than the fetching of the actual image.

Images shared from the "Public" folders of those who still have them will be unaffected as those are hosted on a different domain.


erikdesjardins:

PR #1908, opened 2014-12-26
Clone username from userbar for "commenting as"

### LINKED ISSUE #1895: mouseover 'comment as {{username}}' attempts to fire user info hover, but throws error instead

erikdesjardins:
Fixes #1895


erikdesjardins:
This works now.

PR #1912, opened 2014-12-27
Replace repeated code with RESUtils.getCacheable

erikdesjardins:
This was only repeated three times, so it might not be worth having its own function.

~~Not sure about `getCacheable` or `maxAge`. (suggestions for better names?)~~

The idea behind splitting the callback into `callback` and `handleData` is to save a tiny bit on storage space (we don't have to store the whole response), and for a tiny bit of performance when using the cache (we don't have to parse it again).


andytuba:
after this and #2019 get merged, then modules.menu.checkForUpdate should use .getCacheable

PR #1944, opened 2014-12-31
More linting

erikdesjardins:
There is definitely more that I could have changed, but I tried to be conservative in some areas.


mc10:
Looks ok for me; any thoughts @andytuba?

andytuba:
it looks okay on the face of it, but I'd like to take a deeper look at the spots where var declarations got moved around.

PR #1948, opened 2014-12-31
Add rts button, save-RES button, and vote handlers to newly loaded comments

### LINKED ISSUE #1929: RES does not add buttons/handlers to comments loaded via "load more comments"
it's happening again in firefox: https://www.reddit.com/r/RESissues/comments/2qkkwp/announcement_res_v454_released_bugfixes/cn9cjmc

i can repro in FF34 OSX

erikdesjardins:
Fixes #1929
I think this is everything those comments were missing.

### DIFF
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -361,7 +361,9 @@ modules['userTagger'] = {
 			this.applyTags();
 			//console.log('after applytags: ' + Date());
 			if (RESUtils.pageType() === 'comments') {
-				RESUtils.watchForElement('newComments', modules['userTagger'].attachVoteHandlers);
+				RESUtils.watchForElement('newComments', function(ele) {
+					modules['userTagger'].attachVoteHandlers(ele.parentNode);
+				});
 				RESUtils.watchForElement('newComments', modules['userTagger'].applyTags);
 			} else {
 				RESUtils.watchForElement('siteTable', modules['userTagger'].attachVoteHandlers);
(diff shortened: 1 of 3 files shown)


andytuba:
Thanks! 

Someday I'll refactor those watchers so it fires if interesting elements exist on page load too, and change remove all the duplication in the modules with "call on page load and call on new elements". That'll help move towards waypoint (call on newly visible elements) as well.

lib/modules/userTagger.js
```diff
@@ -361,7 +361,9 @@ modules['userTagger'] = {
 			this.applyTags();
 			//console.log('after applytags: ' + Date());
 			if (RESUtils.pageType() === 'comments') {
-				RESUtils.watchForElement('newComments', modules['userTagger'].attachVoteHandlers);
+				RESUtils.watchForElement('newComments', function(ele) {
+					modules['userTagger'].attachVoteHandlers(ele.parentNode);
```
andytuba: It looks like, if comments with replies get loaded, then `.attachVoteHandlers()` will affect the replies more than once -- because `ele.parentNode` contains both a single new comment and all its replies.

Lessee, how should we work around this?

erikdesjardins: Replace this line with the following?

    modules['userTagger'].attachVoteHandlers(ele.parentNode.querySelector('.midcol'));

Might not be the most efficient way, though.

andytuba:
i made 4e1a5bc021ef454d898d647c8c242458c56ee3e0 more "thing-centric" with 10ca2496a6727b92573f03f3de0aa153c2d5ecd8 but it still feels odd. that section probably deserves a bit more refactoring.

PR #1949, opened 2015-01-01
console: Swap out forEach for $.each

mc10:
I've used `_` as a convention for the unused index; any thoughts for/against it?


andytuba:
i'm a fan of this style  ($.each and i for index)

PR #1950, opened 2015-01-01
Prevent quick messages from being truncated on some characters + more

erikdesjardins:
Fixes [a bug reported on reddit.](https://www.reddit.com/r/RESissues/comments/2qypen/bug_private_message_get_truncated_when_is_used/) (Would you prefer that I create an issue on github, even if it'll be immediately resolved?)

Also, stop silently shortening the subject to 100 characters and add an error message for it.


andytuba:
PR is enough in my book since it's a superset of issue.

PR #1956, opened 2015-01-03
Flickr fixes

erikdesjardins:
Resolves [an issue from reddit.](https://www.reddit.com/r/Enhancement/comments/2r4fuy/correct_flickr_settings_for_expando/)
Also, checking for `media_url` avoids creating an expando when this would occur:

    Picture by: undefined @ Flickr
    Loading of image undefined failed. Click to view directly

### DIFF
--- a/lib/modules/hosts/flickr.js
+++ b/lib/modules/hosts/flickr.js
@@ -41,7 +41,7 @@ modules['showImages'].siteModules['flickr'] = {
 	handleInfo: function(elem, info) {
 		var def = $.Deferred();
 		var imgRe = /\.(jpg|jpeg|gif|png)/i;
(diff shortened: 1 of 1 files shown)


andytuba:
Is this still relevant?  Maybe flickr fixed it on their side? That [test post](https://www.reddit.com/r/itookapicture/comments/2r49tz/itap_for_my_new_year_card/) works fine on the latest build.

erikdesjardins:
Huh, so it does. I still think updating the regex would be worth it (the old regex matches URLs with 11-digit IDs, but at first glance you wouldn't expect it to), as well as checking for `media_url` since this still happens:
image

andytuba:
sounds good. thanks for rebase!

PR #1957, opened 2015-01-03
Remove autocomplete for command line

gmcclure382:
Removes auto complete suggestions from commandline dialog box. As
selecting autocomplete suggestions don't work in the first place
and even if it did, kind of takes away from the user experience.


andytuba:
just gonna ignore those two stray commits..

andytuba:
thanks!

gmcclure382:
Thanks! Sorry about those two stray commits. I didn't realize pull requests automatically get updated with subsequent commits.

mc10:
That's usually why you use a separate branch for each PR, so you can keep your master branch clean while you mess around as much as you wish in your branches.

gmcclure382:
@mc10 Thank you very much for the tip, I will employ that practice immediately.

andytuba:
Plus you can do follow-up commits on that branch in response to feedback on
the pull request, and the PR gets updated with those commits.

gmcclure382:
Oh that makes a lot of sense, I like that idea of isolating each work item to a branch. Thank you, the both of you, for all your help. We use TFS at work (though the usage is more like SourceSafe), so git is pretty new to me.

Issue #1914, opened 2014-12-28
Show gold in account dropdown switcher


andytuba:
https://www.reddit.com/r/Enhancement/comments/2qmrvv/feature_request_show_if_an_account_has_gold_in/cn7ksxq?context=5

I assume we can get this data in the same request as the karma.  I guess slap on a gilded icon after the karma?   and of course a separate option for gilding.

andytuba:
closed in #2004

PR #1959, opened 2015-01-03
Gold Account Switcher

coreyja:
This is my first public pull request. Fixes issue #1914 

Let me know if there is anything that I need to change or fix in this, any feedback would be appreciated. Thanks

Sorry, closed my last one so I could do this pull request from a branch other than master.

### DIFF
--- a/lib/modules/accountSwitcher.js
+++ b/lib/modules/accountSwitcher.js
@@ -152,11 +152,12 @@ modules['accountSwitcher'] = {
 
 									userDetails = username + ' (' + userInfo.data.link_karma + ' &middot; ' + userInfo.data.comment_karma + ')';
 
+									// If the user has gold display the icon and set the title text to the time remaining till it expires
 									if (userInfo.data.is_gold) {
 										$gildedSpan.css('display', 'inline-block');
 										if (userInfo.data.gold_expiration) {
 											var today = new Date();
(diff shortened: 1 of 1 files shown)


andytuba:
Thanks for implementing this!  With a few little fixes, it'll be good to go.

lib/modules/accountSwitcher.js
```diff
@@ -142,13 +149,33 @@ modules['accountSwitcher'] = {
 
 							RESUtils.getUserInfo(function(userInfo) {
 								var userDetails = username;
+								var $gildedSpan = $('<span>', {
+									'class': 'gilded-icon',
+								});
+								$gildedSpan.css('display', 'none');
```
andytuba: This seems superfluous since the element doesn't even get added to the DOM until later.

lib/modules/accountSwitcher.js
```diff
@@ -142,13 +149,33 @@ modules['accountSwitcher'] = {
 
 							RESUtils.getUserInfo(function(userInfo) {
```
andytuba: This section ought to be refactored instead of tacking on copy-pasta.  Instead of calling RESUtils.getUserInfo here, we ought to call modules['accountSwitcher'].updateUserDetails() after the for loop.  Can you take care of that?

PR #1960, opened 2015-01-03
Update favico.js

benwa:
Updating to latest version (plus three more commits).

This may be a good opportunity to use submodules instead of manually updating each library.


mc10:
Thanks!

PR #1971, opened 2015-01-10
[WIP] RESUtils: save only option values, not their descriptions, etc.

### LINKED ISSUE #1510: RESUtils.setOption should only save values, not options definitions
it'd save a few KB and make settings more readable

erikdesjardins:
Resolves #1510 

Before: 143 KB, after: 45 KB (fresh install)

### DIFF
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -109,17 +109,6 @@ RESUtils.isMatchURL = function(moduleID) {
 
 	return included;
 };
-RESUtils.minifyOptions = function(obj) {
-	var min = {};
-	if (obj) {
-		for (var key in obj) {
-			if ('value' in obj[key]) {
-				min[key] = {value: obj[key].value};
-			}
-		}
-	}
-	return min;
-};
 // gets options for a module...
 RESUtils.getOptionsFirstRun = [];
 RESUtils.getOptions = function(moduleID) {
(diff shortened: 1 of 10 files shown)


lib/core/utils.js
```diff
@@ -109,6 +109,17 @@ RESUtils.isMatchURL = function(moduleID) {
...
+	var min = {};
+	if (obj) {
+		for (var key in obj) {
+			if ('value' in obj[key]) {
+				min[key] = {value: obj[key].value};
+			}
+		}
+	}
+	return min;
+};
```
honestbleeps: hmm, I'm preeeettty sure this is safe, but I'm trying to think of a scenario where we might need to save anything more complex beyond just 'value'.... I don't think there is, and we could always make value a complex object if need be... but @andytuba @mc10 or anyone else, any thoughts on that before we merge?

andytuba: I think value(s) is all that needs to go into storage. Lemme check through to see if any `RESStorage.getItem('RESoptions...'` happen outside the usual "load data from storage and merge into schema / default values" routine.

andytuba: I see a few instances of `RESStorage.getItem('RESoptions.` or `.setItem('RESoptions.`:

* `bitcointip.load` and `.save`, but IIRC that was only used in the standalone build. :white_check_mark: 
* `commentTools.migrateData`: will dump unminified options directly into storage, but they'll be minified next time options are saved. it only affects commentTools and commentPreview options, so not a big concern.  :white_check_mark: however, this ought to be refactored into migrate.js at some point. :warning:
* `filteReddit.checkFilterFormat`: stores unminified settings every pageload if the user has any subreddit filters. worth minifying. :exclamation:  also shouldn't re-set settings every pageload, only if anything changed! :warning:

andytuba: Maybe there ought to be a `RESStorage.setOptions(moduleID, [values])` ?

mc10:
Are we minifying existing options to save space on existing installations?

matheod:
I remember some module (maybe some I created) using more than the value from the storage (i.e. default value maybe ?). But I might be wrong, not sure.

erikdesjardins:
@andytuba Yeah I was thinking that `this` might be an issue but I haven't had a chance to check yet. - and you merged it while I was checking this, oh well I'll still comment.
@matheod I believe we can rule out that possibility as follows:
- RESUtils.getOptions has only ever loaded option values [(on this line).](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/1971/files#diff-9fa41a0df75b0d0555d1c410e214ce0cR130)
- Therefore, a module would have to use `RESStorage.getItem('RESoptions...` to access more than that.
- There are only 3 other instances of `RESStorage.getItem('RESoptions`, all of which are safe:

1. migrate.js (269): `RESStorage.getItem('RESOptionsVersion');` - false positive.
1. bitcointip.js (170): `RESStorage.getItem('RESoptions.bitcoinTip');` - Manually overwrites its own options (looks dangerous - how do you add new options?), but doesn't access anything other than optionName.value. (also it's not currently in use)
1. commentTools.js (405): `RESStorage.getItem('RESoptions.commentPreview')` - Should be replaced by a migrator, and only accesses option values.

andytuba:
bitcointip probably should get removed at this point anyway.

PR #1961, opened 2015-01-04
Added a module to configure entry click behavior (EntryTitleAction) -- fixed

ericsubach:
Added a module to configure entry click behavior (EntryTitleAction).

This allows clicking the main title of a link and opening either the link or the comments in a background tab.

This fixes the broken functionality and merge conflicts since the code refactor (replacing merge request #578).

### Options
* Action
  - open the content (reddit's default action)
  - open the comments
* Focus
  - open in foreground
  - open in background

### Testing
Fully tested on Firefox and Chrome. I appreciate any feedback for about its behavior on other browsers.

### Other
* This is similar to the SingleClickOpener module, but I felt it was different enough to warrant its own module.
* The module is disabled by default because **it changes reddit's default behavior**.
* My motivation for making this is the desire to only open the comments (since I can get to the content from there) and to have interaction remain as close to the normal flow as possible (left-clicking on the main link).

### Disclaimer
I am not a web developer and did a lot of copy-pasting based on the existing SingleClickOpener module.

PR #1975, opened 2015-01-12
SingleClickOpener can hijack post title link to open comments or l+c/l=c

andytuba:
also a good chunk of refactoring

also fixed bug in chrome/chropera where opening an l+c, then opening an l=c puts the second link's comments tab in between the first post's links and comments tabs.

closes #1961

### DIFF
--- a/Chrome/background.js
+++ b/Chrome/background.js
@@ -145,14 +145,14 @@ chrome.runtime.onMessage.addListener(
 				if (request.openOrder === 'commentsfirst') {
 					// only open a second tab if the link is different...
 					if (request.linkURL !== request.commentsURL) {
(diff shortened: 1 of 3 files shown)


Chrome/background.js
```diff
@@ -145,14 +145,14 @@ chrome.runtime.onMessage.addListener(
 				if (request.openOrder === 'commentsfirst') {
 					// only open a second tab if the link is different...
 					if (request.linkURL !== request.commentsURL) {
-						chrome.tabs.create({url: request.commentsURL, selected: button, index: newIndex, openerTabId: sender.tab.id});
+						chrome.tabs.create({url: request.commentsURL, selected: button, index: newIndex++, openerTabId: sender.tab.id});
```
mc10: Should this be `newIndex++` or `++newIndex`? In any case, I think it'd be much clearer if we moved the increment right before or after the line.

PR #1977, opened 2015-01-12
Added steamusercontent.com domain support to showImages Steam module.

JonBons:
The steam module only support links on the steampowered.com domain but I'm seeing a lot of links coming out of steamusercontent.com now.

This updates the domains and the detection regex on this module.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -3074,7 +3074,7 @@ modules['showImages'] = {
 			name: 'steam',
 			domains: ['steampowered.com', 'steamusercontent.com'],
 			detect: function(href, elem) {
-				return (/^cloud(?:-\d)?.(?:steampowered|steamusercontent).com$/i).test(elem.host);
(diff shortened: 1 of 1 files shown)


lib/modules/showImages.js
```diff
@@ -3072,9 +3072,9 @@ modules['showImages'] = {
 		},
 		steampowered: {
 			name: 'steam',
-			domains: ['steampowered.com'],
+			domains: ['steampowered.com', 'steamusercontent.com'],
 			detect: function(href, elem) {
-				return (/^cloud(?:-\d)?.steampowered.com$/i).test(elem.host);
+				return (/^cloud(?:-\d)?.(?:steampowered|steamusercontent).com$/i).test(elem.host);
```
mc10: You should escape the periods so they don't match *every* character, like so: `\.`.

JonBons:
Fixed unescaped periods in the regex.

Issue #844, opened 2014-03-24
Spoiler detector module


jewel-andraia:
http://www.reddit.com/r/Enhancement/comments/2184ii/feature_request_integrated_spoiler_prevention/

I envision a "spoiler" module which supports the common conventions, e.g. `[blah blah](/spoiler)`, `[](#spoiler "blah blah")` etc. Subreddit-specific syntaxes which aren't covered can then be specified by the user (via the settings console) or subreddit mods (via a hidden 'RES Settings' flag in the sidebar). It should probably be specified as a CSS selector in RES as well.

StyleTweaks' dark mode and image viewer can then ask the RES spoiler manager how to identify spoilers and act appropriately.

PR #1978, opened 2015-01-14
Add global spoiler tags module

SirCmpwn:
This adds support for common spoiler tag formats on user pages.

### DIFF
--- a/lib/modules/spoilerTags.js
+++ b/lib/modules/spoilerTags.js
@@ -1,0 +1,28 @@
+modules['spoilerTags'] = {
+	moduleID: 'spoilerTags',
+	moduleName: 'Global Spoiler Tags',
+	category: 'UI',
+	options: { },
+	description: 'Hides spoiler-tagged comments globally across Reddit',
+	isEnabled: function() {
+		return RESConsole.getModulePrefs(this.moduleID);
+	},
+	include: [
+		/^https?:\/\/([a-z]+)\.reddit\.com\/user\/.*/i
+	],
+	exclude: [ ],
+	isMatchURL: function() {
+		return RESUtils.isMatchURL(this.moduleID);
+	},
+	go: function() {
+		if (!this.isEnabled() || !this.isMatchURL()) {
+			return;
+		}
+		var styles = RESTemplates.getSync('spoiler-styles');
+		var tag = document.createElement('style');
+		tag.setAttribute('type', 'text/css');
+		tag.innerHTML = styles.text(); // Security note: not user input
+		document.head.appendChild(tag);
+	},
+	hosts: {}
+};
--- a/lib/modules/spoilerTags.js
+++ b/lib/modules/spoilerTags.js
@@ -3,12 +3,12 @@ modules['spoilerTags'] = {
 	moduleName: 'Global Spoiler Tags',
 	category: 'UI',
 	options: { },
-	description: 'Hides spoiler-tagged comments globally across Reddit',
+	description: 'Hides spoiler-tagged comments on user profiles.',
 	isEnabled: function() {
 		return RESConsole.getModulePrefs(this.moduleID);
 	},
 	include: [
-		/^https?:\/\/([a-z]+)\.reddit\.com\/user\/.*/i
+		"profile"
 	],
 	exclude: [ ],
 	isMatchURL: function() {
--- a/lib/modules/spoilerTags.js
+++ b/lib/modules/spoilerTags.js
@@ -23,6 +23,5 @@ modules['spoilerTags'] = {
 		tag.setAttribute('type', 'text/css');
 		tag.innerHTML = styles.text(); // Security note: not user input
 		document.head.appendChild(tag);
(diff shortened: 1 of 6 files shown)


lib/modules/spoilerTags.js
```diff
@@ -0,0 +1,28 @@
...
+		return RESUtils.isMatchURL(this.moduleID);
+	},
+	go: function() {
+		if (!this.isEnabled() || !this.isMatchURL()) {
+			return;
+		}
+		var styles = RESTemplates.getSync('spoiler-styles');
+		var tag = document.createElement('style');
+		tag.setAttribute('type', 'text/css');
+		tag.innerHTML = styles.text(); // Security note: not user input
```
andytuba: thanks for the security note

lib/modules/spoilerTags.js
```diff
@@ -0,0 +1,28 @@
...
+		document.head.appendChild(tag);
+	},
+	hosts: {}
```
andytuba: copy-pasta?

SirCmpwn: Yeah, copy-pasta.

lib/modules/spoilerTags.js
```diff
@@ -0,0 +1,28 @@
...
+	moduleID: 'spoilerTags',
+	moduleName: 'Global Spoiler Tags',
+	category: 'UI',
+	options: { },
+	description: 'Hides spoiler-tagged comments globally across Reddit',
+	isEnabled: function() {
+		return RESConsole.getModulePrefs(this.moduleID);
+	},
+	include: [
+		/^https?:\/\/([a-z]+)\.reddit\.com\/user\/.*/i
```
andytuba: should be  `"profile"`

lib/modules/spoilerTags.js
```diff
@@ -0,0 +1,28 @@
+modules['spoilerTags'] = {
+	moduleID: 'spoilerTags',
+	moduleName: 'Global Spoiler Tags',
+	category: 'UI',
+	options: { },
+	description: 'Hides spoiler-tagged comments globally across Reddit',
```
andytuba: this should read "on user profiles" until support is added elsewhere

lib/modules/spoilerTags.js
SirCmpwn: No it shouldn't be?

lib/modules/spoilerTags.js
SirCmpwn: I thought you were going to extend this to work everywhere?

lib/modules/spoilerTags.js
erikdesjardins: Yes? [`profileRegex`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/a2acbcd09e7293c0fc548eae8d462184912289c5/lib/core/utils.js#L287) should match all URLs that this line does.

SirCmpwn: Ah, I see what you mean now. Thanks.

lib/modules/spoilerTags.js
andytuba: Yes, and that will include updating the description to reflect what the module can actually do.

SirCmpwn: Okidoki.

SirCmpwn:
Bump. Coming up on release and this would be easy to include.

mc10:
Relevant: #844

erikdesjardins:
A few comments:

Even though the current build doesn't work in Opera 12, you should still add it to `Opera/includes/loader.js` in case we have a future release.
Unless there's a compelling reason to create the element manually, you should use `RESUtils.addCSS()` - and then you can do it in `beforeLoad()` instead of `go()`.

And some things that would be useful but aren't too important: using `addModule()` allows you to omit a lot of the boilerplate code (`isMatchURL`, etc.), and `RESTemplates.load` (async) is preferable to `RESTemplates.getSync`.

andytuba:
i'll implement erik's comments and throw in some nightmode, too

allthefoxes:
How is this invoked?

SirCmpwn:
^_^

andytuba:
Right now it only applies when you're looking at a user profile page (main overview).

allthefoxes:
Right, but, how does it know what to hide?

SirCmpwn:
It just applies CSS to the user pages that recognizes the de-facto standard for spoilers: `[text text text](#spoiler)` (and variants).

andytuba:
eventually i want to add in more subreddit-specific options per #844, also apply it in more places - but figured we can just push this out now and collect feedback (and more standards)

allthefoxes:
:+1:

mc10:
Does the code in this PR fix #844?

SirCmpwn:
Yes.

andytuba:
It only partially implements #844.

SirCmpwn:
:cry:

PR #1982, opened 2015-01-16
vendor: Update guiders to 2.0.0

mc10:
I don't think anything major has changed, but this deserves another look.


andytuba:
I haven't dug in too deep yet, but --

I see a lot of class name changes from `guider...` to `guiders...`.  Does it matter to us?

mc10:
Yes, we need to update our CSS as well. I'll look into that.

mc10:
I've updated the CSS, though I didn't add in the `guider_content h2`, `guider_content td`, `.guider_content td + td`, and `.guider_content td code` blocks. Any idea what they were for?

andytuba:
Those .guider_content blocks were for the RES Tips / keyboard nav table that shows up on some tips.  They probably deserve a visual redesign.

PR #1983, opened 2015-01-16
res.user: Remove unnecessary var declarations

mc10:
As these variables are already declared in other files, these declarations should be unnecessary.


andytuba:
seems okay.  i guess i'll find out when I actually go fix opera12.

mc10:
I don't have a copy of Opera to test on, but this change works fine on Firefox.

andytuba:
well, there are other blocking issues on Opera12, so I'll just sort it out all then if need be.. probably do the vars inside browsersupport or something.

PR #1984, opened 2015-01-17
Subreddit shortcut A-Z/Z-A button doesn't sort

andytuba:
https://www.reddit.com/r/RESissues/comments/2spz5y/bug_trying_to_sort_subreddits_az_doesnt_work/


andytuba:
@mc10 Please test your changes when janitoring. Sometimes fixing something "that should've been that way in the first place" exposes other issues downstream.

andytuba:
also if you feel it is semantically better to solve this the other way around (e.event.type == button), that's cool too

PR #1992, opened 2015-01-22
Fix betteReddit option description

allthefoxes:
Reddit has rolled out its own version of unread counts


andytuba:
Seems reasonable..

PR #1999, opened 2015-01-31
Show expandos for twitter's alternate image resolutions

### LINKED ISSUE #913: Expanding Twitter Images
Twitter images aren't recognized by the expander (probably because of the ending, like http://pbs.twimg.com/media/BheNWJSCQAA8Q18.jpg:large).

erikdesjardins:
Fixes #913.

~~I can't imagine it's very likely that `img.<png/jpg/gif>:<something>` wouldn't be an image, so I think it would be fine to do this instead of adding a special case for Twitter.~~

### DIFF
--- a/chrome/manifest.json
+++ b/chrome/manifest.json
@@ -110,6 +110,7 @@
 				"modules/modhelper.js",
 				"modules/quickMessage.js",
 				"modules/hosts/imgur.js",
+				"modules/hosts/twimg.js",
 				"modules/hosts/xkcd.js",
 				"modules/hosts/dailymotion.js",
 				"modules/hosts/jsfiddle.js",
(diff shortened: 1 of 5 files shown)


honestbleeps:
> I can't imagine it's very likely that img.<png/jpg/gif>:<something> wouldn't be an image, so I think it would be fine to do this instead of adding a special case for Twitter.

Hmm, aren't wikipedia "image" URLs (actually pages) something like this?

honestbleeps:
hmm I guess not quote:

http://en.wikipedia.org/wiki/Destin_Sandlin#mediaviewer/File:6TdMNb1_-_Imgur.jpg

erikdesjardins:
Hmmm, perhaps there are other conflicts then. What do you think? (it wouldn't take that long to put it in a sitemodule)

erikdesjardins:
Just made a simple `twimg` host instead of modifying `defaultImage`

PR #2001, opened 2015-02-01
Images to batch icons

andytuba:
Swapped out a few icon images for batch font icons.  This should supersede some of the "move icons from `url(https://aws...)` to `uri(data:...)`" PR

thanks to @erikdesjardins for suggestions.

Still need to do gear dropdown search button

### DIFF
--- a/lib/core/console.js
+++ b/lib/core/console.js
@@ -885,7 +885,7 @@ var RESConsole = {
 			var thisHandle = document.createElement('div');
 			$(thisHandle)
 			.addClass('res-icon-button res-icon handle')
-				.html('&#xF0B5;')
+				.html('&#xF0AA;')
 				.attr('title', 'drag and drop to move this row')
 
 			thisTD.appendChild(thisHandle);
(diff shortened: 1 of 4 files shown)


andytuba:
@erikdesjardins which other ones did you identify?

erikdesjardins:
There's also the [dashboard edit button](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/core/res.css#L182), which maybe could be replaced by F139 or similar.

Other than the search magnifying glass, I think that's it, unless there's anything in [the spritesheet](https://s3.amazonaws.com/b.thumbs.redditmedia.com/tBwwK20XXxtpgudWx1L7bDXla-iotv-JA0jgA0Y-FVs.png) that could be replaced.

andytuba:
Hmm, that spritesheet.. 

+, < and > could definitely be replaced.  I wonder if + is even used anywhere.

Should we swap out the <=  =>  and loop arrows from the image expando gallery controls? dunno which subreddits have styled those.

Comment navigator up/down buttons maybe?

NER play/pause button could definitely become a batch icon, with a smidgen of work.

andytuba:
/r/dashboard's "refresh" button could also become a batch icon.

PR #2002, opened 2015-02-01
Move showImages.siteModules into individual files

### LINKED ISSUE #1868: Refactor showImages.siteModules into separate files a la BetterZoom
\<andytuba> and omg we need to refactor sitemodules
\<andytuba> into their own files
\<allthefoxes> why?
\<allthefoxes> is it that bad
\<andytuba> mrege conflicts

---

Putting this here so it won't be forgotten, and will get done one day

erikdesjardins:
Fixes #1868 
I believe I've tested all of them and they all work ~~(gfycat throws errors, but it performs the same in the release version)~~.

Are there any other changes that should be made? (i.e. using  `.addSiteModule()` a la [`.addModule()`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/1699c7c65c0a80bb5a3f2e2e6ace566b5246f95c/lib/reddit_enhancement_suite.user.js#L101))
Right now the only changes are to the first and last lines of each siteModule for syntax.

We'll also probably need to update gulp.


honestbleeps:
what's this business about gfycat not working? works fine for me

andytuba:
Whatcha think about putting them in the "hosts" folder instead of "siteModules"? I feel like the concepts (pull media from, push media to) are pretty heavily related, and IIRC makelinks and gulp are already set up for files in hosts/

erikdesjardins:
Yeah it isn't RES, but regardless of browser (even a fresh install of Firefox with no other extensions), I got "error loading expando gfycat.com..." for all expandos.
I think it has something to do with CloudFlare thinking I might be DoSing them (I would get "CloudFlare is checking your browser" on like 1/4 pageloads), but it seems to be working now.

@andytuba That sounds reasonable.

andytuba:
Spot-checked it on Chrome, Firefox, and Safari. turns out gulp does need some fixups so I'll do the usual add-on PR trick

PR #2012, opened 2015-02-04
showImages: stop conserveMemory messing up albums

### LINKED ISSUE #1494: Albums sometimes randomly switch back to previous images
This bug is immensely annoying to track down (mainly because it doesn't show up all the time), but sometimes, in albums (I've only seen it in imgur albums as those are the most popular), if you browse through an album and stay at a image, the album randomly switches to another (previous) image in the gallery. The size of the placeholder element is not updated to account for the change in images until you click-and-drag to resize.

erikdesjardins:
Fixes #1494.

Originally, the image would be replaced if the current `src` was different from the stored source. This obviously causes problems in albums, resulting in the image getting reset after each scroll.

Instead, we can check that the image we're replacing is the `transparentGif` used as a substitute for the original image. As far as I can tell, conserveMemory still works properly after this change.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -317,7 +317,7 @@ modules['showImages'] = {
 		if (ele.imageLink && ele.imageLink.image) {
 			$img = $(ele.imageLink.image);
 			src = $img.data('src');
-			if (src && ($img.attr('src') !== src)) {
+			if (src && ($img.attr('src') === this.transparentGif)) {
 				$img.attr('src',src);
 			}
 		}


honestbleeps:
dude... high-five @changetip

changetip:
Hi @erikdesjardins, @honestbleeps sent you a Bitcoin tip worth 1 high-five (22,020 bits/$5.00), and I'm here to deliver it ➔ **[collect your tip](https://www.changetip.com/collect/423170)**.


**[Learn more about ChangeTip](https://www.changetip.com/tip-online/github)**

honestbleeps:
let's make it another $5 because this bug has been annoyingly haunting our background for too long. @changetip

changetip:
Hi @erikdesjardins, @honestbleeps sent you a Bitcoin tip worth 21,975 bits ($5.00), and I'm here to deliver it ➔ **[collect your tip](https://www.changetip.com/collect/423277)**.


**[Learn more about ChangeTip](https://www.changetip.com/tip-online/github)**

mc10:
:+1: Thanks for fixing this annoying bug.

andytuba:
aww yeah, updated known issues

PR #2022, opened 2015-02-07
Settings console - updated layout and partially templated

andytuba:
screen shot 2015-02-07 at 2 12 46 pm

Categories and modules now an accordion menu in the left pane
Search box moved to top of left pane (just above categories/module)

Settings console markup is all in templates except for the options container/markup and each module's description.


andytuba:
One problem that needs addressing: Chrome renders the category/module list poorly if the overflow scrollbar appears/disappears during accordion animation.

PR #2023, opened 2015-02-08
CSS Cleanup

erikdesjardins:


lib/modules/nightmode.css
```diff
@@ -947,8 +946,7 @@ html.res-nightmode .listing-chooser .create button {
 html.res-nightmode .multi-details .subreddits .remove-sr,
 html.res-nightmode .multi-details .add-sr .add {
 	border-color: #222;
-	background-image: url(https://b.thumbs.redditmedia.com/TItqbtZ6QLVach67LqyHqj-cCnYMempXSwW17SDlTKM.png);
-	background-position: 0 0;
+	background: url(https://b.thumbs.redditmedia.com/TItqbtZ6QLVach67LqyHqj-cCnYMempXSwW17SDlTKM.png) 0 0;
```
andytuba: this one might be worth leaving alone, since you're potentially clobbering other background properties.

andytuba:
cool, thanks, looks good except for that one little note

lib/modules/nightmode.css
erikdesjardins: got it

PR #2031, opened 2015-02-11
fadeSpeed corrections

erikdesjardins:
Fixes this issue [(reddit post).](https://www.reddit.com/r/RESissues/comments/2vh45v/bug_cant_disable_fading_on_any_hover_popups)

`$.fadeTo` measures speed in seconds (instead of the fractional speed we used before), causing `fadeSpeed.value = 1` to fade things out over 1 second instead of instantly (as intended).


andytuba:
whoopsie, I think I broke that.

PR #2033, opened 2015-02-12
Added subreddit hover feature for full links.

### LINKED ISSUE #2021: subredditHover does not work on full links
the hover info for subreddits works on things like /r/subreddit, but not http://www.reddit.com/r/subreddit

Lets play, BUG. OR. FEATURE?

rageandqq:
Noticed that subreddit hovering doesn't work for full links, and that there's already an [issue ticket](https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2021) open for it.  [edit: fixes #2021]

I originally thought it would be a lot trickier, as you would need to be able to accept all formats like:
- http://reddit.com/r/sub
- https://reddit.com/r/sub
- http://www.reddit.com/r/sub
- https://www.reddit.com/r/sub
- www.reddit.com/r/sub  

While also ignoring links over other protocols, like: 
- file://reddit.com/r/sub

However, it turns out RES automatically only links together proper Web URLS, so the issue was much simpler and complicated CSS selectors weren't needed.
I moved the selectors to an array (although I didn't space them out one per line) for better readability and modifiability.

### DIFF
--- a/lib/modules/subredditInfo.js
+++ b/lib/modules/subredditInfo.js
@@ -61,7 +61,8 @@ modules['subredditInfo'] = {
 	srRe: /\/r\/(\w+)(?:\/(new|rising|controversial|top))?\/?$/i,
 	addListeners: function(ele) {
 		ele = ele || document.body;
-		var subredditLinks = document.body.querySelectorAll('.listing-page a.subreddit, .explore-page a.subreddit, .comment .md a[href^="/r/"], .side a[href^="/r/"], .usertext-body a[href^="/r/"], .trending-subreddits a[href^="/r/"]');
+        var querySelectors = ['.listing-page a.subreddit','.explore-page a.subreddit','.comment .md a[href^="/r/"]','.comment .md a[href*="reddit.com/r/"]','.side a[href^="/r/"]','.side a[href*="reddit.com/r/"]','.usertext-body a[href^="/r/"]','.usertext-body a[href*="reddit.com/r/"]','.trending-subreddits a[href^="/r/"]','.trending-subreddits a[href*="reddit.com/r/"]'];
+		var subredditLinks = document.body.querySelectorAll(querySelectors.join());
 		if (subredditLinks) {
 			var len = subredditLinks.length;
 			for (var i = 0; i < len; i++) {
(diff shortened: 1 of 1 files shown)


lib/modules/subredditInfo.js
```diff
@@ -61,7 +61,8 @@ modules['subredditInfo'] = {
 	srRe: /\/r\/(\w+)(?:\/(new|rising|controversial|top))?\/?$/i,
 	addListeners: function(ele) {
 		ele = ele || document.body;
-		var subredditLinks = document.body.querySelectorAll('.listing-page a.subreddit, .explore-page a.subreddit, .comment .md a[href^="/r/"], .side a[href^="/r/"], .usertext-body a[href^="/r/"], .trending-subreddits a[href^="/r/"]');
+        var querySelectors = ['.listing-page a.subreddit','.explore-page a.subreddit','.comment .md a[href^="/r/"]','.comment .md a[href*="reddit.com/r/"]','.side a[href^="/r/"]','.side a[href*="reddit.com/r/"]','.usertext-body a[href^="/r/"]','.usertext-body a[href*="reddit.com/r/"]','.trending-subreddits a[href^="/r/"]','.trending-subreddits a[href*="reddit.com/r/"]'];
```
honestbleeps: I wonder if we should perhaps perftest this... it's getting kind of bananas with this number of querySelectors.

might it be faster to querySelector all links and scan them for a subreddit link? I'm honestly not sure, I'm up late / can't sleep and just postulating, but adding this many queryselectors is probably pretty costly.

honestbleeps: sorry, I should add 2 things:

1) @rageandqq thank you for submitting this PR!

2) I'm not suggesting it wasn't already bananas before you added more to that list. :-)

erikdesjardins: One minor thing: can you use tabs instead of spaces here?

rageandqq:
@honestbleeps No worries, I'll try running some performance tests and look into an alternate method of querying the links.

@erikdesjardins Silly me, forgot to change my editor settings, I'll get on that!

allthefoxes:
I like this guy.

lib/modules/subredditInfo.js
andytuba: i like the idea of a two-filter pass:

    $('a[href*="/r/"]').filter(function() { return this.href.test(RESUtils.subredditRegex); })

andytuba: or we could go the other direction, which is probably bad performance in a different way:

    $(document.body).on('mouseover', 'a[href*="/r/"]', 
        function(e) { if (this.href.test(RESUtils.subredditRegex)) {
          hover blah blah
    });

PR #2035, opened 2015-02-14
Obviate calls to /subreddits/mine/moderator.json

JordanMilne:
This is the second most common API call that RES makes to the API, and it's unnecessary for the average user. Change it so we don't check for subreddits we moderate if we're sure we don't moderate any :sun_with_face:


andytuba:
thanks Jordan!  Looks good. I tested with non-mod and mod user, and after 5min cache-bust. 

I further reduced the number of moderator.json API calls by deferring until the quickMessage dialog is actually opened.

JordanMilne:
Woot, thanks andy!

PR #2037, opened 2015-02-16
Allow modmail keyboard shortcut to be saved properly

### LINKED ISSUE #1954: Can't change modmail keyboard shortcut in RES settings
https://www.reddit.com/r/RESissues/comments/2r4g6m/bug_sourcesnudown_shows_gt_instead_of/

erikdesjardins:
Fixes #1954

This occurred because the event handler to update the keycode for each shortcut used `.getElementById(optionKey)`, which would select the modmail icon (for users that have it) instead of the modmail option.

I'm not sure if it would be better to change our `id` naming scheme instead - it would make more sense but require more changes.

@andytuba I don't appear to have permission to make a PR from the issue; I imagine I might if I was the issue's creator.

### DIFF
--- a/lib/modules/settingsConsole.js
+++ b/lib/modules/settingsConsole.js
@@ -96,8 +96,9 @@ addModule('settingsConsole', function(module, moduleID) {
 					} else {
 						keyArray = [e.keyCode, e.altKey, e.ctrlKey, e.shiftKey, e.metaKey];
 					}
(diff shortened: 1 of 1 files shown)


andytuba:
I would like to change the naming scheme, but it can wait for a larger refactor... or iwe could just do that little bit now.

PR #2057, opened 2015-03-01
Remove uneffective RESTemplates.load calls without callbacks

erikdesjardins:
Maybe they used to do something, I don't know.
But since `RESTemplates.load` does absolutely nothing [if the callback is falsy](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/401b0b61363ae7921599c9939dcb68acf1127671/lib/core/template.js#L5), they don't do anything now.


andytuba:
I think at one point that was a signal to RESTemplates to preload everything.  heck with it.

Let's kill getTemplateSync, too, since that leads to bad behavior.

PR #2060, opened 2015-03-01
quickMessage.getValidSendFrom minor refactoring

erikdesjardins:
This fixes an issue where your username would get added to the "from" dropdown twice - once immediately and once when we received a response from reddit.


andytuba:
whoops, thanks

PR #1352, opened 2014-07-01
Added Pastebin and JSFiddle Support, Code Cleanup

sgtfrankieboy:
Implemented support for Pastebin and JSFiddle.

**Pastebin**
Was planning on using the embed feature of Pastebin but there were problems with the IFrame so I decided to use the raw output instead. The `hashRe` should match the following urls:

    http://pastebin.com/7nAAqYRP
    http://pastebin.com/index/7nAAqYRP
    http://pastebin.com/raw.php?i=7nAAqYRP

**JSFiddle**
JSFiddle uses it's own embed functionality. Someone might need to look at the `hashRe` to clean it a bit up. It should match the following urls:

    http://jsfiddle.net/zalun/NmudS/0/
    http://jsfiddle.net/zalun/NmudS/0
    http://jsfiddle.net/zalun/NmudS/
    http://jsfiddle.net/zalun/NmudS
    http://jsfiddle.net/NmudS/0/
    http://jsfiddle.net/NmudS/0
    http://jsfiddle.net/NmudS/
    http://jsfiddle.net/NmudS

**Site Removals**
Removed the websites `picsarus` and `memefive` since they have been down for quite a while.

**Refactoring**
Some minor refactoring `imageLink -> mediaLink`, `imgDiv -> mediaDiv`

PR #2061, opened 2015-03-01
Pastebin expando support

erikdesjardins:
Replaces would-be pastebin support in #1352, since it appears that their iframes work fine now, even with https.

image
(might be too wide, i dunno)

### DIFF
--- a/lib/modules/hosts/pastebin.js
+++ b/lib/modules/hosts/pastebin.js
@@ -1,0 +1,34 @@
+modules['showImages'].siteModules['pastebin'] = {
+	domains: ['pastebin.com'],
+	detect: function(href, elem) {
+		return href.indexOf('pastebin.com/') !== -1;
+	},
+	handleLink: function(elem) {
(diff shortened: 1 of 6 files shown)


lib/modules/hosts/pastebin.js
```diff
@@ -0,0 +1,34 @@
...
+		} else {
+			def.reject();
+		}
+		return def.promise();
+	},
+	handleInfo: function(elem, info) {
+		var generate = function(options) {
+			var element = document.createElement('iframe');
+			element.src = info;
+			element.style.width = '80em';
```
andytuba: maybe slap in an option for width, default 80em?

allthefoxes:
All hail @erikdesjardins

lib/modules/hosts/pastebin.js
andytuba: oh nvm, you put in a maxwidth.

PR #2062, opened 2015-03-01
pastebin.js: option for expando width

erikdesjardins:
No `|| options.default` since it's not catastrophic if it's left blank, they can leave the width unspecified if they want.

### DIFF
--- a/lib/modules/hosts/pastebin.js
+++ b/lib/modules/hosts/pastebin.js
@@ -1,5 +1,12 @@
 modules['showImages'].siteModules['pastebin'] = {
 	domains: ['pastebin.com'],
+	options: {
+		width: {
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/pastebin.js
```diff
@@ -1,5 +1,12 @@
 modules['showImages'].siteModules['pastebin'] = {
 	domains: ['pastebin.com'],
+	options: {
+		width: {
+			description: 'The maximum width of pastebin expandos. (any valid CSS width, default 80em)',
```
andytuba: whenever the options GUI get redesign, we should throw in a place to show the default value more obviously t han hovering over the title

PR #2064, opened 2015-03-02
Fix account switcher overlay toggle

Shraymonks:
Fixes some edge cases with the account switcher menu:
- No more weirdness when hovering from the menu to the alien
- Menu properly closes after opening the accounts settings page


andytuba:
cool, thanks!

Issue #2058, opened 2015-03-01
UserHover not calculating age correctly


allthefoxes:
Have't confirmed yet, just copying from reddit - http://www.reddit.com/r/RESissues/comments/2xl0jw/bug_incorrect_calculation_of_account_age/

allthefoxes:
Can confirm

Test here: http://www.reddit.com/user/meow--

yangzihe:
I think I solved the bug. Just one question: is RESUtils.niceDateDiff supposed to include the current date in the calculation as well?

PR #2072, opened 2015-03-06
Fixed user age calculator on userhover

yangzihe:
I fixed the bug showing wrong age of certain accounts. Also added a little bit of documentation to explain logic for new code.

</img>

### DIFF
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -1105,9 +1105,9 @@ RESUtils.niceDateDiff = function(origdate, newdate) {
 	var aday = origdate.getUTCDate();
 	var ayear = origdate.getUTCFullYear();
 
-	var tyear = newdate.getUTCFullYear();
 	var tmonth = newdate.getUTCMonth() + 1;
 	var tday = newdate.getUTCDate();
+	var tyear = newdate.getUTCFullYear();
 
 	var y = 1;
 	var mm = 1;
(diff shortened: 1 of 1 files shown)


allthefoxes:
Magic words:

Fixes #2058

lib/core/utils.js
```diff
@@ -1105,9 +1105,9 @@ RESUtils.niceDateDiff = function(origdate, newdate) {
 	var aday = origdate.getUTCDate();
 	var ayear = origdate.getUTCFullYear();
 
-	var tyear = newdate.getUTCFullYear();
 	var tmonth = newdate.getUTCMonth() + 1;
 	var tday = newdate.getUTCDate();
+	var tyear = newdate.getUTCFullYear();
```
andytuba: tiny style note: why did you move this?

andytuba:
Yay, thanks for math fixes!

lib/core/utils.js
yangzihe: Oh because I wanted to match the order of the original date variables. I guess it's a little bit nitpicky, but I saw no reason why the variables wouldn't have the same order in the first place, so I just shifted the line.

honestbleeps: I like the cut of your jib, sir.

andytuba: fair nuff.

Issue #1994, opened 2015-01-24
Login panel not night-mode compatible


andytuba:
![](https://i.imgur.com/WQ9NEaD.png)

https://www.reddit.com/r/RESissues/comments/2ti6gr/bug_redditcomlogin_is_not_night_mode_compatible/

andytuba:
@gavin19 can you do that fancy thang you do?

also lemme know if you're ever getting tired of nightmode stuff, or want me to ping you for other more interesting stuff.

githue:
Can't repro in same version of RES and Chrome 40:

![reddit-login-nm](https://cloud.githubusercontent.com/assets/7245595/6077821/7900ae32-ae3e-11e4-9591-2825a86bce22.png)

andytuba:
i meant specifically the white background on the password fields

PR #2078, opened 2015-03-08
Nightmode UI Fixes

yangzihe:
Fixes https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/1994

The filler text is also visible now in the text fields. Hope this is good; I've included some before and after pictures.

Before

After


andytuba:
snazzy!

PR #2093, opened 2015-03-10
Use batch icon for advanced options warning

erikdesjardins:
I don't see the current one on Chrome (a browser specific issue that isn't FF? whaaat), I guess it doesn't choose fallback fonts as intelligently:
image
Here's how it looks in Firefox, original / this PR:
image
image

Alternatively we could specify the font for the original icon, but I don't believe there's a single font that'll work on all platforms. (judging by [this site](http://www.fileformat.info/info/unicode/char/26a0/fontsupport.htm), whose trustworthiness is debatable, we'd have something like: `font-family: 'Segoe UI Symbol', FreeSans, STIXGeneral-Regular;`)

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -274,9 +274,11 @@ body.res-console-open {
 	display: none;
 }
 .optionContainer.advanced > label:after {
-	content: '\26A0';
+	content: '\F15A';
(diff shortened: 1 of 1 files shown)


andytuba:
thanks for fixing that!

PR #2099, opened 2015-03-11
Remove commented out code

Shraymonks:
Let's get rid of all of this commented code?

### DIFF
--- a/Opera/index.html
+++ b/Opera/index.html
@@ -41,11 +41,9 @@ var XHRCache = {
 	count: 0,
 	check: function(key) {
 		if (key in this.entries) {
-//			console.count('hit');
 			this.entries[key].hits++;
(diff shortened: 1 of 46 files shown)


honestbleeps:
while a lot of this code can definitely be removed, it's not universal.

some of that code is useful to leave in / commented because we uncomment for debugging (e.g. the perftest stuff).

Ideal world: make it only run in a debug mode, or something, of course...

I don't have the time to do so just yet, but one of us could maybe submit a PR against your branch that returns the commented code we need to keep in RES - then we could merge this.

hopefully someone's feeling ambitious, otherwise it'll have to wait until I have time to look myself :-\

mc10:
Going to close this for now; if you'd like to make a more finely-tuned comment removal, I'd be happy to merge that instead.

PR #2101, opened 2015-03-13
Use triangle icon for commentNav buttons

### LINKED ISSUE #2083: Use unicode triangle wedge up/down for comment navigator up/down buttons
http://i.imgur.com/87CBWf9.png 

and make 'em real buttons.  with disabled states, and cursor:pointer/default and opacity:100/30.

erikdesjardins:
Fixes #2083

imageimage

image

### DIFF
--- a/lib/modules/commentNavigator.js
+++ b/lib/modules/commentNavigator.js
@@ -89,7 +89,7 @@ modules['commentNavigator'] = {
 						case 'images':
 							thisEle.setAttribute('title', 'Navigate through comments with images');
 							break;
-						case 'images':
+						case 'videos':
 							thisEle.setAttribute('title', 'Navigate through comments with videos');
 							break;
 						case 'popular':
--- a/lib/modules/commentNavigator.js
+++ b/lib/modules/commentNavigator.js
@@ -33,10 +33,9 @@ modules['commentNavigator'] = {
 			RESUtils.addCSS('.commentarea .menuarea { margin-right: 0; }');
 			RESUtils.addCSS('.menuarea > .spacer { margin-right: 0; }');
 			RESUtils.addCSS('#commentNavButtons { margin: auto; }');
-			RESUtils.addCSS('#commentNavUp { margin: auto; cursor: pointer; background-image: url("https://s3.amazonaws.com/b.thumbs.redditmedia.com/tBwwK20XXxtpgudWx1L7bDXla-iotv-JA0jgA0Y-FVs.png"); width: 32px; height: 20px; background-position: 0 -224px; }');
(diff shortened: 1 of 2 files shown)


lib/modules/commentNavigator.js
```diff
@@ -89,7 +88,7 @@ modules['commentNavigator'] = {
 						case 'images':
 							thisEle.setAttribute('title', 'Navigate through comments with images');
 							break;
-						case 'images':
+						case 'videos':
```
andytuba: nice catch

lib/modules/commentNavigator.js
```diff
@@ -152,7 +151,7 @@ modules['commentNavigator'] = {
 					<div id="commentNavCloseButton" class="RESCloseButton">&times;</div> \
 					<div class="RESDialogContents"> \
 						<div id="commentNavButtons"> \
-							<div id="commentNavUp"></div> <div id="commentNavPostCount"></div> <div id="commentNavDown"></div> \
+							<button id="commentNavUp" type="button" disabled>&#x25B2;</button> <div id="commentNavPostCount"></div> <button id="commentNavDown" type="button" disabled>&#x25BC;</button> \
```
andytuba: :+1: `<button>` so semantic

lib/modules/commentNavigator.js
erikdesjardins: I'll have to give that credit to my IDE (IntelliJ)

honestbleeps: @erikdesjardins I have to ask - I tried using IntelliJ with RES and I can't because it seems to not properly obey / update symlinked files when I save the main files (e.g. lib/modules/foo.js) - so I don't see the updates w hen I refresh the extension.  How are you addressing this problem, or is it unique to me and my setup / OS?  I reverted to Sublime for this reason instead of starting to take advantage of IntelliJ!

Are you on OSX or Windows?

erikdesjardins: I'm on Windows, but I just tried building with makelinks and it also doesn't work (but I get "cannot connect to extension", rather than it just not updating).

You shouldn't have issues with symlinks if you use gulp or grunt to build though, right? (both work for me)

andytuba: gulp copies the files to another folder, instead of symlinking.  so yes, you won't have issues with the symlinks =p

honestbleeps: the symlinks work such that the extension loads just fine, it's just that edits with IntelliJ don't seem to propagate to the linked files, whereas edits with SublimeText do.

using gulp and copying files over each time sucks... requires a "watch" command which is lame, but I guess if I worked something into my workflow to automate starting that on save, maybe it wouldn't be such a PITA.

erikdesjardins: So I just tried `makelinks.sh` again, a little more carefully (closing Chrome first :P), and was able to reproduce it.

Then I managed to find [this issue report](https://youtrack.jetbrains.com/issue/IDEA-78480), and followed its suggestion to turn off Settings > Appearance & Behavior > System Settings > Use "safe write" - and voila, it worked!

PR #2102, opened 2015-03-14
Replace gear icon with Batch gear icon

erikdesjardins:
Although the current gear is somewhat iconic, so maybe it wouldn't be a good idea to replace it.

imageimage
imageimage

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -106,9 +106,8 @@ body.res-console-open {
 }
 .gearIcon::before, .gearIcon::after {
 	font-family: Batch;
-	line-height: 16px;
-	font-size: 16px;
-	user-select: none;
+	line-height: 15px;
+	font-size: 15px;
 }
 .gearIcon::after {
 	content: '\F04D';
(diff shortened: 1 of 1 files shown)


andytuba:
flat definitely is the trend the days.

gavin19:
Looks better! Maybe step it down to 15px so it doesn't make the user bar any taller.

Minor tweaks - `:` is sufficient for those pseudo-elements. We still need the webkit-prefixed variant for `user-select` but I can't see that it's needed in this instance?

erikdesjardins:
@gavin19 Yeah, those elements aren't selectable in the first place, I have no idea what I was thinking.

I do realize that we can use `:`, since that's how it was before CSS3, but I feel `::` makes more sense semantically, since it's used for all other pseudo-elements, while `:` is for pseudo-classes. Although it's not like someone won't know what `:after` is, so I can change it if you want.

andytuba:
sorry about all the crazy assigns. github mobile client is a bit twitchy

andytuba:
i haven't put much thought into it, but ::pseudo-element / :pseudo-class() does feel habitual..

PR #2108, opened 2015-03-15
don't hide the gear icon if subreddit style is disabled

erikdesjardins:


andytuba:
Could you add a `modules['styleTweaks'].isSubredditStyleEnabled()` method and call that instead of examining the checkbox? That checkbox really shouldn't be a public property.. yay legacy cruft.

andytuba:
thanks!

PR #2117, opened 2015-03-16
. command line for rpc-executing RES module methods

andytuba:
For example,

    . troubleshooter.notification
    . troubleshooter.notification()

both invoke `modules['troubleshooter'].notification('rpc')` in the RES extension context

    . troubleshooter.notificiation(1, 2, 3)

invokes modules['troubleshooter'].notification(1, 2, 3, 'rpc') in the RES extension context

--

This is very useful for debugging and troubleshooting -- but is it safe for distribution?


andytuba:
@agentlame reports that Mozilla approved toolbox's console module.. maybe if we this were only activated by a hidden option, then it would be safe-ish.

PR #2120, opened 2015-03-17
Macro default values, settings console fix

andytuba:
Move boilerplate macros into macro default values so that users can customize them better and not have to put up with defaults they don't want.

Not bothering to add migration to add default macros because anyone who cares can probably do it themselves or ask for help.

Removed look of disapproval. It's stale. 

---

I fixed settings console showing "undefined" in table options when a value is missing (like when a new field is added)

---

P.S. I left the LOD modal in in case somebody wants to add an autodetect in for "user has pasted a hundred [R..E..S]()".

### DIFF
--- a/lib/modules/commentTools.js
+++ b/lib/modules/commentTools.js
@@ -756,10 +756,6 @@ modules['commentTools'] = {
 				this.macroSelection(box, '[Reddit Enhancement Suite](http://redditenhancementsuite.com/) ');
 			}));
 
-			this.addButtonToMacroGroup('', this.makeEditButton('&#3232;_&#3232;', 'Look of disaproval', null, 'btn-macro btn-lod', function(button, box) {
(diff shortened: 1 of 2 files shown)


lib/modules/commentTools.js
```diff
@@ -111,7 +111,11 @@ modules['commentTools'] = {
 				name: 'key',
 				type: 'keycode'
 			}],
-			value: [],
+			value: [
+				['reddiquette', '[reddiquette](/wiki/reddiquette) '],
+				['Promote RES', '[Reddit Enhancement Suite(http://redditenhancementsuite.com "also /r/Enhancement") '],
```
mc10: You accidentally a `]` after "Reddit Enhancement Suite".

andytuba: oopsie.

allthefoxes:
Add /rules?

PR #2123, opened 2015-03-19
Simplify the copying of browser files in gulp

erikdesjardins:
This produces exactly the same output as before (unless there are unexpected files in `lib/`, in which case it will perform identically to Grunt and makelinks.sh).


andytuba:
didn't like this implementation?

PR #2130, opened 2015-03-22
Stop changing expandos' link URLs

erikdesjardins:
This is no longer necessary, since we use [addUrlToHistory()](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/8719cc456bce7ef6f338ad7c22f172c5873d3833/lib/modules/showImages.js#L1636).

Also, for single-image expandos, make the image link to its source URL (this basically preserves existing behavior, since it used to be set to the title URL, which used to be changed to the source URL).

### DIFF
--- a/lib/modules/hosts/ctrlv.js
+++ b/lib/modules/hosts/ctrlv.js
@@ -24,7 +24,6 @@ modules['showImages'].siteModules['ctrlv'] = {
 
 		elem.type = info.type;
 		elem.src = info.src;
-		elem.href = info.href;
 
 		if (RESUtils.pageType() === 'linklist') {
 			$(elem).closest('.thing').find('.thumbnail').attr('href', elem.href);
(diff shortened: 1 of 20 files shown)


andytuba:
@honestbleeps Remind me why we haven't done this all along. IIRC it was to mark things visited immediately and in history. can we rely on addUrlToHistory?

erikdesjardins:
If there's an issue it's with Safari, since URLs show up as visited immediately for me in FF and Chrome, for all of my [test cases](https://cloud.githubusercontent.com/assets/7673145/6770824/5c63ae0a-d0a3-11e4-8608-ced9774be17a.png).

honestbleeps:
The problem was that "half" the browser we supported (Safari and old Opera) needed the iframe hack.

I think we can stop doing this and just say "too bad" for Safari, because the iframe hack for tracking history doesn't even seem to work anymore anyway so this isn't helping anyone.

andytuba:
I'm sold.

PR #2131, opened 2015-03-22
Instagram expando support

### LINKED ISSUE #2126: instagram expandos
http://www.reddit.com/r/resissues/comments/2zvo8d/_/

erikdesjardins:
Fixes #2126 

~~They do have an oembed API, but it doesn't seem to support CORS.
This means that instagram videos won't work, but it's better than having to add another permission.~~

Test links:
https://www.reddit.com/domain/instagr.am/
https://www.reddit.com/domain/instagram.com/

### DIFF
--- a/lib/modules/hosts/instagram.js
+++ b/lib/modules/hosts/instagram.js
@@ -1,0 +1,14 @@
+export default {
+	moduleID: 'instagram',
+	name: 'Instagram',
+	domains: ['instagram.com', 'instagr.am'],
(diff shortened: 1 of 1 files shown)


honestbleeps:
i'd be fine with optional permissions like we do for twitter...

andytuba:
We should fix noembed and route the request through there.

erikdesjardins:
Turns out they had an iframe embed all along :sweat_smile:

PR #2133, opened 2015-03-23
Move twitter support into its own siteModule

erikdesjardins:
This needs testing ([links](https://www.reddit.com/domain/twitter.com)) in Safari, but I suspect it'll work.

### DIFF
--- a/lib/modules/styleTweaks.js
+++ b/lib/modules/styleTweaks.js
@@ -269,20 +269,6 @@ modules['styleTweaks'] = {
 					}');
 				}
 			}
-			if (this.options.showExpandos.value) {
-				var twitterLinks = document.body.querySelectorAll('.entry > p.title > a.title');
-				var isTwitterLink = /twitter.com\/(?:#!\/)?([\w]+)\/(status|statuses)\/([\d]+)/i;
-				for (var i = 0, len = twitterLinks.length; i < len; i++) {
-					var thisHref = twitterLinks[i].getAttribute('href');
-					thisHref = thisHref.replace('/#!', '');
-					if (isTwitterLink.test(thisHref)) {
-						var thisExpandoButton = document.createElement('div');
-						thisExpandoButton.setAttribute('class', 'expando-button collapsed collapsedExpando selftext twitter');
-						thisExpandoButton.addEventListener('click', modules['styleTweaks'].toggleTweetExpando, false);
-						RESUtils.insertAfter(twitterLinks[i].parentNode, thisExpandoButton);
-					}
-				}
-			}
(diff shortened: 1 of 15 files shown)


lib/modules/styleTweaks.js
```diff
@@ -406,81 +392,6 @@ modules['styleTweaks'] = {
...
-					modules['styleTweaks'].tweetExpando = thisExpando;
-					opera.extension.postMessage(JSON.stringify(thisJSON));
-				} else if (BrowserDetect.isFirefox()) {
-					// we've got a jetpack extension, hit up the background page...
-					// we have to omit the script tag and all of the nice formatting it brings us in Firefox
-					// because AMO does not permit externally hosted script tags being pulled in from
-					// oEmbed like this...
-					jsonURL += '&omit_script=true';
-					modules['styleTweaks'].tweetExpando = thisExpando;
-					self.postMessage(thisJSON);
```
erikdesjardins: > AMO does not permit externally hosted script tags being pulled in from oEmbed like this
The thing is, we're adding `&omit_script=true` to the original `jsonURL`, which doesn't affect the URL used by our request (in thisJSON, which has already been created [here](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2133/files#diff-6360b0be2d4a48c79cdcd07e6d8c9820L423)).

So we need to decide if we want to correctly omit the script (current behavior; loses twitter formatting in FF) or keep the same behavior and load it anyways.

honestbleeps: we were losing formatting before in FF, because of this very issue.

we can't include the script in FF, they'll reject it as you point out.

erikdesjardins: Well, they must not have been paying much attention, since they didn't reject [4.5.4](https://i.imgur.com/var8jVQ.png), or [4.5.3](https://i.imgur.com/GXVKzYJ.png), or [4.5.0.2](https://cloud.githubusercontent.com/assets/7673145/6791007/27204c5a-d182-11e4-8595-d6825da9bff2.png) - all of which include the script (we still shouldn't risk it for v4.6 - but it is amusing).

After some investigation: Twitter expandos were downgraded for about a year, which ended about a year ago: the [original commit](https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/d865829ccba7769703c2dfd74bc48528065b0b33) to omit the script was after release v4.2.0.2, and it was accidentally nullified by [this commit](https://github.com/mc10/Reddit-Enhancement-Suite/commit/c993360b64f5bef1b507c54dd971bf99d3efd7f5#diff-0767dec274256dd1746817f7f75c9890L15095) after release v4.3.1.2.

PR #2134, opened 2015-03-24
gear icon font size 15px -> 14px

erikdesjardins:
For better alignment:
(original / 15px / 14px - Chrome without notifications, FF with) - *now with valign:middle it's 2px lower than this image*
image

Also removes the 3px extra padding-bottom from the userbar - here's how it looks now:
image

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -108,6 +108,7 @@ body.res-console-open {
 	font-family: Batch;
 	line-height: 15px;
 	font-size: 14px;
+	vertical-align: middle;
 }
 .gearIcon::after {
 	content: '\F04D';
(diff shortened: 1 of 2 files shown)


gavin19:
I think it still needs that middle valign so it fits with the alignment of the other RES icons in the user bar.

    font-size: 14px;
    line-height: 1;
    vertical-align: middle;

?

Somewhat related, but I think we should kill the extra bottom padding on #header-bottom-right to even it out. Currently 4px 4px 7px, when 4px would suffice.

erikdesjardins:
I like both of those suggestions; FF will be off by 1px if line-height isn't larger than font-size, though.

PR #2136, opened 2015-03-25
gulp overhaul

### LINKED ISSUE #2141: Update readme with new gulp zip syntax
@allthefoxes or @erikdesjardins

erikdesjardins:
Fixes #2141 

This produces exactly the same output as before (unless there are unexpected files in lib/, in which case it will perform identically to Grunt and makelinks.sh).

Gulp now works as follows:

There are 6 tasks: `build, zip, watch, add-module, add-host, clean`.
Browsers (`chrome, firefox, safari, opera, oblink`) can be specified with `-b browser1 -b browser2`. If none are specified, gulp will use all of them.
Files for add-module and add-host are still specified with `--file filename.js`.
The zip output directory now defaults to `<rootBuildDir>/zip`, and can be specified with `--zipdir /path/to/dir`.

If you run multiple tasks the specified browser will apply to all of them, i.e. `gulp build watch -b chrome` will build only chrome and watch only chrome.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -211,5 +211,5 @@ gulp.task('opera-zip', function() {
 
 // Other
 gulp.task('clean', function(cb) {
-	del(['dist/**/*'], cb);
+	del(['dist/*'], cb);
 });
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -98,16 +98,10 @@ gulp.task('zip', function(cb) {
 });
 
 gulp.task('watch', function() {
-	var sources = selectedBrowsers.reduce(function(previous, browser) { // combine the sources of each browser
-		return previous.concat(
-			config[browser].buildFiles.reduce(function(pre, cur) { // combine the sources within each browser
-				return pre.concat(
-					cur.src.filter(function(src) { // filter out repeated directories (i.e. commonFiles, chrome/oblink common files)
-						return previous.indexOf(src) == -1;
-					})
-				);
-			}, [])
-		);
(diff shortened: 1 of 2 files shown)


honestbleeps:
haven't tested this yet, but the spaghetti that was the gulp script before was a big reason I was hesitant to even use it, so :+1: :)

allthefoxes:
Can you please submit this as a PR to /allthefoxes/RESGulp please? Thanks

allthefoxes:
Tested. Working fine. :+1:

gulpfile.js
```diff
@@ -427,108 +181,38 @@ gulp.task('add-host-opera', function() {
...
-
 // Zip tasks
 gulp.task('chrome-zip', function() {
-	return gulp.src(path.join(chromeBuildDir, '**/*'))
+	return gulp.src(path.join(buildDir.chrome, '**/*'))
 		.pipe(zip('chrome.zip'))
 		.pipe(gulp.dest(zipDir));
 });
 
 gulp.task('safari-zip', function() {
```
mc10: Is there a way to combine zip tasks into a single one that takes in a browser parameter?

allthefoxes: like `gulp zip (all|chrome|firefox|etc)`?

mc10: Yeah something like that!

mc10:
Is this good to go? If so, I'll merge this in.

allthefoxes:
:+1: I've already been using it, so its fine with me

PR #2140, opened 2015-03-28
mediaBrowseMode: remember state

### LINKED ISSUE #2119: mediabrowsemode behavior is borked
seems to lose track of "state" when navigating over posts that don't have an expando now, regression

erikdesjardins:
Fixes #2119


andytuba:
finally the world feels normal again

PR #2144, opened 2015-03-29
manual sort module categories in settings console, split about/core

andytuba:
Keep "About RES" at the top with a few high-interest core modules, move "core components of RES" into its own category at the bottom.

screen shot 2015-03-29 at 1 23 32 pm
screen shot 2015-03-29 at 1 23 37 pm


andytuba:
i am open to shuffling around modules to other categories, too.

PR #2146, opened 2015-03-29
highlight nsfw buttons

andytuba:
and an option to turn it off, of course

closes #1804 

screen shot 2015-03-29 at 3 52 13 pm

screen shot 2015-03-29 at 3 52 05 pm

### DIFF
--- a/lib/core/init.js
+++ b/lib/core/init.js
@@ -30,9 +30,7 @@ function RESdoBeforeLoad() {
 				}
 			}
 		}
-		if (RESUtils.htmlClasses.length > 0) {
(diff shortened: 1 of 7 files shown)


agentlame:
Not to knock anyone's work, but I'm not sure those stand out enough to have the desired effect.  Still, glad it made it in!

allthefoxes:
@agentlame I agree. I would not notice that on my screen.

Is there any issues with making the stripes red or something?

mc10:
Maybe make red versions of the expando buttons? The stripes are pretty subtle.

andytuba:
They're at ~15% opacity, we could easily bump that up to make it more obvious.

PR #2148, opened 2015-03-30
Fade in the image loading spinner

erikdesjardins:
So it doesn't flicker when you move to the next gallery image.

Alternatively, it could pop into existence after a little while, or maybe something else. (any suggestions?)


andytuba:
I felt like it was unresponsive, so I made it show very faded almost immediately, then faded it in basically how you had it.

Issue #2059, opened 2015-03-01
Orangered might incorrectly update incognito tabs logged in on different account


andytuba:
two potential solutions:

1. setUnreadCount(count, **username**, source)
1. modules should subscribe to multicast message instead of using rpc. this will allow future functionality like accountSwitcher picking up unreads from background process checking multiple accounts

PR #2157, opened 2015-04-05
Prevent multicast between incognito and normal tabs

andytuba:
fixes orangered misfiring read/unread at the wrong account


erikdesjardins:
Fixes #2059, then?

andytuba:
yes, thanks much for looking up the number.

PR #2158, opened 2015-04-06
Announcement badge pizzazz

### LINKED ISSUE #2155: Make the announcement (!) badge pop
So many people think the red ! is a normal part of the RES gear. It deserves a little pizzazz to set it apart.

1. Fade it in a second or two after page load
1. If RES knows the user has seen the exclamation mark before (maybe for at least a day or two), then make it dance after fading it.

erikdesjardins:
Fixes #2155.
Whenever the "!" is added on page load, it will flash a few times (pretty non-intrusively):
test
However, if the user hasn't cleared it for 3 days, it will get a little more intrusive (and it won't stop):
img

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -127,6 +127,18 @@ body.res-console-open {
 	-webkit-animation: fadeIn 0.5s ease-out 7 alternate;
 	animation: fadeIn 0.5s ease-out 7 alternate;
 }
+.gearIcon.newNotification.important::before {
+	-webkit-animation: gearIconPop 1s ease-out infinite alternate;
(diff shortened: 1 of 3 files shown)


andytuba:
Nifty! :fireworks:Does touching the `fadeIn` transition affect the picture-loading throbber?

erikdesjardins:
It should be unaffected (and it appeared to be in testing).
The only difference is that it will fade from `0` to to the element's opacity instead of always fading to `opacity: 1` - but since the throbber is opaque by default, it acts the same.

andytuba:
groovy, thanks for double-checking.

PR #2164, opened 2015-04-06
Made blockquotes more readable in night mode.

Lovatoty:
Changed the color of the text for quotes in night mode, to it is easier to read.


erikdesjardins:
This has already been fixed (last release vs. master branch):
imageimage
Pure white makes it lighter than normal text, which is probably too bright.

PR #2168, opened 2015-04-07
Fixed #2139 -- new comment count now shows in Japanese language interface

### LINKED ISSUE #2139: new comment count doesn't show in non English locale
https://www.reddit.com/r/RESissues/comments/30c040/bug_invisible_new_comment_count_in_japanese/

svjharris:
Fixes #2139. There was a split of the number of the comments and the word "comments" predicated on a space. Japanese doesn't use a space after numbers from what I can tell. Changed it to strip all non-numeric characters.

I noticed it always uses the word "new" no matter the interface language. Does RES have multiple interface languages? Is it an issue that "new" is always in English?


andytuba:
thanks! hurray small steps towards i18n

PR #2177, opened 2015-04-09
settingsConsole: Replace RuntimeError with Exception

mc10:
`RuntimeError` does not exist as an error for JavaScript.


erikdesjardins:
Another one here: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/d0a0da5a5eefcc61d7948f2e5dacd71b4096bda7/lib/modules/filteReddit.js#L494

PR #2179, opened 2015-04-09
#1875 -- Command line launches from menu and from gear icon click

svjharris:
Implements #1875 in both ways mentioned.


andytuba:
Needs options (default off)  and code clean up. I can take care of it.

Issue #963, opened 2014-04-19
Educate user about reddit preference "don't use subreddit style"


jewel-andraia:
I see the occasional question like [how do I turn off subreddit style for all subreddits?](http://www.reddit.com/r/Enhancement/comments/23ff9v/feature_request_please_summarize_your_feature/)  RES could educate users that there exists a reddit preference > display options > [☑ allow subreddits to show me custom styles](https://ssl.reddit.com/prefs/#show_stylesheets).

I propose that, every 4th or 5th subreddit that a user disables style on, RES shows a notification suggesting they disable all styles.  (Note: every nth _subreddit_, not every nth _user turned off subreddit style_.  RES already knows which/how many subreddits you've disabled RES on.  Also, if the user is switching the toggle often, this avoids excessive notifications... this should probably also not fire more than once every day.)

Issue #945, opened 2014-04-15
Notify user about reddit global "use subreddit styles" preference if it's off and they enable srstyle 


jewel-andraia:
http://www.reddit.com/r/RESissues/comments/232jzx/bug_subreddit_styles_are_still_disabled_even/cgt6qip?context=3

PR #2181, opened 2015-04-12
option to always disable subreddit style

andytuba:
https://www.reddit.com/r/Enhancement/comments/329tyy/feature_request_disable_custom_styles_by_default 

On the one hand, this is an antisocial feature. On the other hand, it's still traffic, and it's another step towards allowing RES to manage subreddit stylesheets better.


lib/modules/styleTweaks.js
```diff
@@ -111,6 +111,12 @@ modules['styleTweaks'] = {
 			description: 'Add an icon in the omnibar (where the page address is written) to disable/enable current subreddit style.',
 			noconfig: true // only show for chrome
 		},
+		disableSubredditStyle: {
+			type: 'boolean',
+			value: false,
+			description: 'Always remove subreddit style.'
```
allthefoxes: I personally don't think this is enough. People might be really stupid.

Use a hover similar to np warnings when enabling?

andytuba: probably yeah. it would pair well with #945. also #963.  

could do something like how userbar collapser notifies you on next page load "uh are you sure you wanted to do that?"  but for this case, since it's a setting, it'll need a little more architecture.. which I'd rather defer until I hear what kinds of questions/confusions people get into.

erikdesjardins:
This also needs #4342 to not be fragile / difficult to implement w.r.t. timing

Jeremy-Laier:
Any eta on this? Passing checks

PR #2185, opened 2015-04-15
YouTube expandos: only ignore post if it already has an expando

### LINKED ISSUE #2184: add youtu.be to youtube expando handler
https://www.reddit.com/r/Enhancement/comments/32mgad/feature_request_an_expander_for_youtube_links/

erikdesjardins:
–instead of ignoring all posts.

Fixes #2184 

Also, we already know that the domain matches, so we don't need to check it again.

### DIFF
--- a/lib/modules/hosts/youtube.js
+++ b/lib/modules/hosts/youtube.js
@@ -2,11 +2,8 @@ modules['showImages'].siteModules['youtube'] = {
 	name: 'youtube',
 	domains: ['youtube.com', 'youtu.be'],
 	detect: function(href, elem) {
-		// Only find comments, not the titles.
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/youtube.js
```diff
@@ -25,7 +22,7 @@ modules['showImages'].siteModules['youtube'] = {
 
 			if (timecodeResult !== null) {
 				var time_blocks = {'h':3600, 'm':60, 's':1},
-					timeRE = /[0-9]+[h|m|s]/ig;
+					timeRE = /[0-9]+[hms]/ig;
```
andytuba: her majesty's regex

PR #2186, opened 2015-04-15
Use commas by default to match reddit's style

fj:
Reddit now shows commas in your link karma total by default, so for consistency we should show commas when displaying comment karma. Screenshot without RES active:

screenshot of Reddit showing commas in karma counts

### DIFF
--- a/lib/modules/showKarma.js
+++ b/lib/modules/showKarma.js
@@ -11,7 +11,7 @@ modules['showKarma'] = {
 		},
 		useCommas: {
 			type: 'boolean',
-			value: false,
+			value: true,
 			description: 'Use commas for large karma numbers'
 		}
 	},


allthefoxes:
:+1:

andytuba:
This will need a migration as well.  I'll get to it tonight.

andytuba:
also thanks for superfast turnaround!

xiongchiamiov:
Sorry, I forgot this was an RES thing or I would've warned you before pushing it out.

Question about the implementation: does it pick up on locale settings?  If not, things will still be a bit weird for people with languages set to places that use period instead of comma for the thousands separator.

fj:
> Question about the implementation: does it pick up on locale settings? If not, things will still be a bit weird for people with languages set to places that use period instead of comma for the thousands separator.

This only changes the default representation, if you look at the commit. It looks like RES uses `commaDelimitedString` to add commas, which chops up the input string into groups of 3 digits and adds commas between them. See: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/255a399e1f2b9421a2f0322ce4cbe9e661bbe1b6/lib/core/utils.js#L1375-L1385

IMO, that's not ideal and it should just use `number.toLocaleString()`:

```
var n = 1234;
n.toLocaleString();
// "1,234"
```

If you like, I can modify `commaDelimitedString` to fix this in a separate PR, though it seems to be used in a lot of places.

andytuba:
It sounds reasonable to change RESUtils.createElement.commaDelimitedString to just use Number.prototype.toLocaleString.  I'm okay with doing it in this same PR.

andytuba:
I took care of the comma number change, thanks for the suggestion. Will merge sometime in the next few days.

allthefoxes:
Only issue here is that someone turning off useCommas would expect it to remove the link karma comma - Should a little disclaimer be added to the description?

andytuba:
maybe I should rename commas to commentCommas and add a linkCommas too

andytuba:
merged to master

marcodeppe:
So the pull request says "merged to master" on 4/16. Why is it still an issue? http://imgur.com/pt9lq6X
I'm using RES 4.5.4 on Chrome.

honestbleeps:
because we haven't pushed out a release yet. "master" is the latest bleeding edge code.  it will be fixed in the next release which doesn't have a specific scheduled date yet because we're all quite busy w/jobs etc.

mc10:
master != release; if you checked out the [releases page] you'd see that we haven't had another release yet since v4.5.4.

marcodeppe:
Apologies for the spam. Last release (v4.5.4) is from Dec 29, 2014.

I pulled master locally and the bug is fixed :+1:

honestbleeps:
please just be careful running master. you'll get out of date unless you keep pulling and updating, and you should expect bugs -- LOTS of bugs ... so please don't post to /r/RESIssues for support unless you're running on a production build, thanks!

Issue #239, opened 2013-01-03
Back to top?


colby-swandale:
i may be blind but does RES not have a back to top button? sometimes when i get onto the 15th page for example i want to go back to the very top. I am happy to contribute some code if anyone else agrees

PR #2192, opened 2015-04-19
back to top button, some floater refactoring, some commentNavigator bugfixes

andytuba:
screen shot 2015-04-18 at 9 55 44 pm
screen shot 2015-04-18 at 9 58 43 pm

could use a little vertical alignment


mc10:
Fixes #239.

PR #2194, opened 2015-04-21
More expando improvements

erikdesjardins:
Mainly this PR removes the caching (anti)pattern that appears in a lot of siteModules, where responses are cached in an object (i.e. only for one pageload) - so it only makes a difference when a link appears on the same page twice, which is uncommon (especially since RES ignores repeated links by default).
Technically it doesn't *hurt* anything, but I know that many people have blindly copied it when creating new siteModules (I probably have, at some point), so I think it's best to get rid of it.

And if we absolutely need to avoid sending the same request twice, we can use `aggressiveCache: true`, which does the exact same thing.

Also includes support for `tumblr.com/image/` URLs, a few other kinds of tumblr text posts - specifically `quote, link, chat, answer` - and better handling of tumblr image captions (the album caption is used if an image doesn't have a caption).

### DIFF
--- a/lib/modules/hosts/deviantart.js
+++ b/lib/modules/hosts/deviantart.js
@@ -1,7 +1,6 @@
 modules['showImages'].siteModules['deviantart'] = {
 	name: 'deviantART',
 	domains: ['deviantart.com', 'fav.me'],
-	calls: {},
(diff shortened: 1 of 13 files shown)


andytuba:
yay decrufting!  yay better tumblr support!

PR #2202, opened 2015-04-26
Don't hide the save button for disabled modules

### LINKED ISSUE #2197: settings console hides "save options" button when showing disabled modules
solution: just never hide the button

https://www.allthefoxes.me/res/bug.webm thanks for reporting @allthefoxes

erikdesjardins:
Fixes #2197


andytuba:
Thanks!

PR #2207, opened 2015-04-27
Optionally show hidden sort options on all pages (random, Q&A, etc.)

erikdesjardins:
Fixes an issue [from reddit.](https://www.reddit.com/r/Enhancement/comments/33ocbr/feature_request_add_random_to_the_comments)

### DIFF
--- a/lib/modules/betteReddit.js
+++ b/lib/modules/betteReddit.js
@@ -103,6 +103,12 @@ addModule('betteReddit', (module, moduleID) => {
 			type: 'boolean',
 			value: false,
 			description: 'Modify reddit\'s comment/link buttons ("perma-link source save...") such that they don\'t show up in searches. Disabled by default due to a slight performance impact.'
+		},
+		showHiddenSortOptions: {
+			type: 'boolean',
+			value: false,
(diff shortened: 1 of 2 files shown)


andytuba:
heck, show em all?

erikdesjardins:
Sure, sounds good.

lib/modules/betteReddit.js
```diff
@@ -121,6 +121,12 @@ addModule('betteReddit', function(module, moduleID) {
 			type: 'boolean',
 			value: false,
 			description: 'Modify reddit\'s comment/link buttons ("perma-link source save...") such that they don\'t show up in searches. Disabled by default due to a slight performance impact.'
+		},
+		showHiddenSortOptions: {
+			type: 'boolean',
+			value: true,
```
erikdesjardins: Not sure if this should be enabled by default. I'm imagining a bunch of reports of "q&a sort is showing up on non-q&a pages".

andytuba: Yeah, false by default is a good idea.  I could envision other special-case sorts appearing in the future, too.

andytuba: Yeah, false by default is a good idea.  I could envision other special-case sorts appearing in the future, too.

PR #2112, opened 2015-04-28
Option to sentence-case or Title Cast post titles

andytuba:
(or do nothing by default)

### DIFF
--- a/lib/modules/styleTweaks.js
+++ b/lib/modules/styleTweaks.js
@@ -139,6 +139,24 @@ modules['styleTweaks'] = {
 			value: false,
 			description: 'Makes the left sidebar (with multireddits) float as you scroll down so you can always see it.',
 			advanced: true
+		},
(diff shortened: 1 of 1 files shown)


lib/modules/styleTweaks.js
```diff
@@ -753,5 +772,33 @@ modules['styleTweaks'] = {
...
+			case 'none': // fall through
+			default:
+				// do nothing
+		}
+
+		if (this.options.postTitleCapitalization.value === 'sentence') {
+			RESUtils.watchForElement('siteTable', modules['showImages'].findAllImages);
+		}
+
+		function segregateFirstLetter(ele) {
```
erikdesjardins: Is this left over from something? It never gets called.

andytuba: whoops.  yeah, this was for Sentence Case but then I figured out inline-block worked.

PR #2221, opened 2015-05-09
Add user-select:none to RES widgets in comments

ArrestedDevelopment:
So when you're copy-pasting stuff from a comment, the navigation helpers, user tags, embeds, etc. don't get copied with it.


andytuba:
Huh, I thought keyNavAnnotations at least had something addressing that issue. Maybe it regressed.

Thanks!

PR #2228, opened 2015-05-12
Deferred-based RESInit

andytuba:
intended to simplifying modules' dependencies on loading document, storage, options, other modules, and its own setup functions (loadDynamicOptions, beforeLoad, go, afterLoad)

seems to work pretty well.


mc10:
Yesyesyes

allthefoxes:
eli5

honestbleeps:
The ELI5:

I've wanted for a long time to redo the way RES handles initialization because too much of it was built at totally different times and often in a rush to release. It's a complete cluster!@#!@ of bad things like setTimeout checking, etc.

When we have the async storage refactor (something I've wanted to do for ages) done, where RES no longer "copies" all the localstorage from background to foreground, we'll be able to complete this sort of behavior.

I haven't looked at the diff yet (sorry, no time), but basically the RES init flow should look something more like:

### Before page load / immediately

get all (or at least "important to have early") module settings so we know important things like "is night mode supposed to be enabled"?

add CSS for things like night mode that need to not have a flash of unstyled content

add CSS classes to HTML tag and/or HEAD as early as is humanly possible to avoid FOUC (some of the CSS that RES adds depends on these classes existing, adding them to the body 'activate's those sections of CSS)

### at page load / later

run / apply all modules (these may still need to be run synchronously, not all spawned individually via an async process, due to dependencies, so we probably need to get "all module options" via async storage, then run all modules via the same sort of for loop we do now)

add other optional/conditional CSS that's not as necessary to load before page load

as needed, get any/all module related data


There's more to it than that, but that's the ELI5.

allthefoxes:
Cool, thanks!

lib/core/init.js
```diff
@@ -1,57 +1,229 @@
...
+	RESUtils.init._watch();
+	RESUtils.init._loadModules();
+
+	RESUtils.init.await.documentElementReady().done(RESUtils.addBodyClasses);
+	RESUtils.init.await.documentBodyReady().done(RESUtils.addBodyClasses);
+	RESUtils.init.await.beforeLoad().done(RESUtils.addBodyClasses);
+	RESUtils.init.await.go().done(RESUtils.addBodyClasses);
+	RESUtils.init.await.documentBodyReady().done(RESUtils.watchMouseMove);
+	RESUtils.init.await.documentBodyReady().done(RESUtils.init._homePage);
+	RESUtils.init.await.documentBodyReady().done(RESUtils.init._reportVersion);
```
mc10: I'm pretty sure you can chain `.done` callbacks together, assuming `documentBodyReady` returns a `$.Deferred` object.

lib/core/init.js
```diff
@@ -1,57 +1,229 @@
...
+		RESUtils.init.await.documentBodyReady(),
+		RESUtils.init.await.go()
+	]).done(RESUtils.initObservers);
+};
+
+RESUtils.init.await = function(type, id) {
+	var def = RESUtils.init.await._deferred(type, id);
+	return def.promise();
+};
+[ 'loadDynamicOptions', 'options', 'beforeLoad', 'documentElementReady', 'documentHeadReady', 'documentBodyReady', 'go', 'afterLoad' ].forEach(function(type) {
```
mc10: You can probably shorten `documentElementReady`, `documentHeadReady`, and `documentBodyReady` to `documentReady`, `headReady`, and `bodyReady` without confusion.

lib/core/init.js
```diff
@@ -1,57 +1,229 @@
...
+				}
+			}
+		}
+	}
+	RESUtils.init.await._deferreds = {}; // allow garbage collection
+}
+
+RESUtils.init.complete = function() {
 	// $.browser shim since jQuery removed it
 	$.browser = {
```
mc10: It doesn't seem like any of our code uses `$.browser` so we should be able to drop this complete code without issues.

lib/core/init.js
```diff
@@ -1,57 +1,229 @@
...
-	// body should be present now, add those classes.
-	RESUtils.addBodyClasses();
 
-	// if RESStorage isn't fully loaded, and _beforeLoadComplete isn't true,
-	// then wait. It means we haven't read all of the modules' options yet.
-	if (!RESStorage.isReady || !_beforeLoadComplete) {
-		setTimeout(RESInit, 10);
+RESUtils.init._loadModules = function () {
+	if (!RESUtils.regexes.all.test(location.href)) {
+		// added this if statement because some people's Greasemonkey "include" lines are getting borked or ignored, so they're calling RES on non-reddit pages.
```
mc10: I thought we dropped support for Greasemonkey?

lib/core/init.js
andytuba: oh yeah, I think we upgraded/replaced whatever third-party lib was using it. Thanks!

lib/core/init.js
andytuba: good idear

lib/core/init.js
andytuba: Yeah, I should trim that copypasta. It's not a bad idea anyway, but for other reasons like "andytuba poorly refactored the first round of environment checks."

andytuba: There's also a "migrate settings from Greasemonkey" bit still in browsersupport-firefox.js but dunno if it actually works..

lib/modules/nightMode.js
```diff
@@ -66,6 +66,20 @@ addModule('nightMode', function(module, moduleID) {
 			module.disableNightMode();
 		}
 	},
+	loadDynamicOptions: function() {
+		RESUtils.deferred.all([
+			RESUtils.init.await.documentHeadReady(),
```
mc10: This should probably be changed.

lib/core/utils.js
```diff
@@ -683,11 +663,11 @@ RESUtils.setSelectValue = function(obj, value) {
 RESUtils.addStyle = function(css) {
 	var style = document.createElement('style');
 	style.textContent = css;
-	var head = document.getElementsByTagName('head')[0];
-	if (head) {
-		head.appendChild(style);
-		return style;
-	}
+	RESUtils.init.await.documentHeadReady().done(function() {
```
mc10: Ditto.

andytuba: oops, thanks

erikdesjardins:
stylesheet.js looks like a [prime candidate](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/066cab242f01867f142ad573c0049b3e08767546/lib/modules/stylesheet.js#L235-L239) for `headReady`

lib/core/options.js
```diff
@@ -268,6 +268,23 @@ $.extend(RESUtils.options, {
...
 		RESStorage.setItem('RESoptions.' + moduleID, JSON.stringify(minify(modules[moduleID].options)));
 	},
+	loadOptions: function(moduleID) {
+		var needsStorage = false;
+		Object.getOwnPropertyNames(modules[moduleID].options || {}).some(function(optionKey) {
+			var type = modules[moduleID].options[optionKey].type;
+			if (type !== 'button') {
+				needsStorage = true;
+				return true;
+			}
```
erikdesjardins: This can be simplified a bit:

	var needsStorage = Object.getOwnPropertyNames(modules[moduleID].options || {}).some(function(optionKey) {
		var type = modules[moduleID].options[optionKey].type;
		return type !== 'button';
	});

andytuba:
@mc10 @erikdesjardins @honestbleeps @allthefoxes I think it's ready to get merged and then we can see what I forgot over the next few days. Opinions?

erikdesjardins:
Sounds good to me.

andytuba:
Alright, boyos, I got a thumbs-up.  and look I EVEN GOT OPERA12 WORKING  those lucky gits

PR #2235, opened 2015-05-15
Add fastdom.js to RES

empyrical:
Library home:

https://github.com/wilsonpage/fastdom

This is a library that's designed to help queue up DOM read and write operations using [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame). As far as I know, every browser RES targets has support for that function; I'm not sure about pre-chromium Opera though

I thought I'd try adding this to help make an initial page load be a little faster. I notice a nice improvement with the changes I've done so far on the old laptop I'm using right now, but I figured it'd be a good idea if I got thoughts on this idea before putting more effort into this/fixing warts/etc.

Tested and working on Firefox and Chrome, haven't tried Opera yet and don't currently have a way to test on Safari.

### DIFF
--- a/lib/modules/commentTools.js
+++ b/lib/modules/commentTools.js
@@ -298,114 +298,132 @@ modules['commentTools'] = {
 				this.fetchTitle();
 			}
 
+			fastdom.write(function() {
+				$body.on('click', 'div.markdownEditor-wrapper a:not(.userTagLink)', function(e) {
+					e.preventDefault();
 
-			$body.on('click', 'div.markdownEditor-wrapper a:not(.userTagLink)', function(e) {
-				e.preventDefault();
+					var index = $(this).data('macro-index');
+					var box = modules['commentTools'].findTextareaForElement(this)[0];
+					// var box = $(this).closest('.usertext-edit, .RESDialogContents, .wiki-page-content').find('textarea[name=text], textarea[name=description], textarea[name=public_description]')[0];
+					if (box === null) {
+						console.error('Failed to locate textarea.');
+						return;
+					}
+					var handler = modules['commentTools'].macroCallbackTable[index];
+					if (handler === null) {
+						console.error('Failed to locate find callback.');
+						return;
+					}
+					handler.call(modules['commentTools'], this, box);
 
-				var index = $(this).data('macro-index');
-				var box = modules['commentTools'].findTextareaForElement(this)[0];
-				// var box = $(this).closest('.usertext-edit, .RESDialogContents, .wiki-page-content').find('textarea[name=text], textarea[name=description], textarea[name=public_description]')[0];
-				if (box === null) {
-					console.error('Failed to locate textarea.');
-					return;
-				}
-				var handler = modules['commentTools'].macroCallbackTable[index];
-				if (handler === null) {
-					console.error('Failed to locate find callback.');
-					return;
-				}
-				handler.call(modules['commentTools'], this, box);
-
-				box.focus();
-				//Fire an input event to refresh the preview
-				var inputEvent = document.createEvent('HTMLEvents');
-				inputEvent.initEvent('input', true, true);
-				box.dispatchEvent(inputEvent);
-			}).on('click', '.RESMacroDropdownTitle', function(e) {
-				var thisCat = e.target;
-				if (thisCat.classList.contains('openMacro')) {
-					thisCat.classList.remove('openMacro');
-					modules['styleTweaks'].setSRStyleToggleVisibility(true, 'commentTools-macroDropdown');
-				} else {
-					$('.RESMacroWrappingSpan span').removeClass('openMacro');
-					thisCat.classList.add('openMacro');
-					modules['styleTweaks'].setSRStyleToggleVisibility(false, 'commentTools-macroDropdown');
-				}
-				// position the drop down so it's flush with the right of the category button.
-				$(this).next().css({
-					top: thisCat.offsetTop + thisCat.offsetHeight + 'px',
-					left: (thisCat.offsetLeft + thisCat.offsetWidth) - thisCat.nextSibling.offsetWidth + 'px'
+					box.focus();
+					//Fire an input event to refresh the preview
+					var inputEvent = document.createEvent('HTMLEvents');
+					inputEvent.initEvent('input', true, true);
+					box.dispatchEvent(inputEvent);
 				});
 			});
 
+			fastdom.write(function() {
+				$body.on('click', '.RESMacroDropdownTitle', function(e) {
+					var thisCat = e.target;
+					if (thisCat.classList.contains('openMacro')) {
+						thisCat.classList.remove('openMacro');
+						modules['styleTweaks'].setSRStyleToggleVisibility(true, 'commentTools-macroDropdown');
+					} else {
+						$('.RESMacroWrappingSpan span').removeClass('openMacro');
+						thisCat.classList.add('openMacro');
+						modules['styleTweaks'].setSRStyleToggleVisibility(false, 'commentTools-macroDropdown');
+					}
+					// position the drop down so it's flush with the right of the category button.
+					$(this).next().css({
+						top: thisCat.offsetTop + thisCat.offsetHeight + 'px',
+						left: (thisCat.offsetLeft + thisCat.offsetWidth) - thisCat.nextSibling.offsetWidth + 'px'
+					});
+				});
+			});
+
+
 			if (this.options.showInputLength.value) {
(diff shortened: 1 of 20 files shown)


honestbleeps:
thanks for submitting this, it's definitely the sort of tweak/performance improvement experimentation we welcome! I don't have time to look at this just now but I'm sure one or more of us will give it a whirl soon. Appreciate the contribution!

high-five @changetip

changetip:
Hi @ArrestedDevelopment, @honestbleeps sent you a Bitcoin tip worth 1 high-five (21,256 bits/$5.00), and I'm here to deliver it ➔ **[collect your tip](https://www.changetip.com/collect/842280)**.


**[Learn more about ChangeTip](https://www.changetip.com/tip-online/github)**

ArrestedDevelopment:
Awesome, and thank you for the tip! :D

erikdesjardins:
>not sure about pre-chromium Opera though

[It apparently doesn't](http://caniuse.com/#feat=requestanimationframe), but fastdom will just [fall back to](https://github.com/wilsonpage/fastdom/blob/a20293848431773f16cdc64b249273573864e06a/index.js#L20) `setTimeout` in that case, so it'll still function.

honestbleeps:
I would just add that our commitment to Opera 12 maintenance is inversely proportional to the time since the last Opera 12.* release.

It's just too much of a hassle to maintain long term, way too many quirks and it holds us back from improvements elsewhere.

lib/modules/commentTools.js
```diff
@@ -298,114 +298,132 @@ modules['commentTools'] = {
 				this.fetchTitle();
 			}
 
+			fastdom.write(function() {
+				$body.on('click', 'div.markdownEditor-wrapper a:not(.userTagLink)', function(e) {
```
andytuba: @ArrestedDevelopment, is there an advantage to attaching event handlers during the write cycle?  I'm wondering if I should emulate this or if you just got on a roll :p

ArrestedDevelopment: I don't know exactly why I put one there, haha. I think it might have been the fact it's got an element selector there, but it should be in the read cycle instead

ArrestedDevelopment: @andytuba should be less nonsensical now!

empyrical:
Now here's an even crazier idea I came up with, but it'll require the use of [Babel](https://babeljs.io) on any files that use it. I think it'd be more convenient to work with, though.

The spec for the ES2016/"ES7" hasn't finalized yet, but Babel/Regenerator works with it pretty well in the current state. Basically, it's heavy syntactic sugar for Promises.

Let's say I make these extremely basic Promise wrappers for Fastdom:

```js
function fwrite() {
  return new Promise( resolve => fastdom.write(resolve) );
}

function fread() {
  return new Promise( resolve => fastdom.read(resolve) );
}
```

I can then do this:

```js
async function test() {
  await fread();
  var testDiv = document.getElementById('testDiv');
  
  for ( let i = 0; i < 1000; i++ ) {
    await fwrite();

    let newElement = document.createElement('div');
    newElement.textContent = 'hello, world!';
    
    testDiv.appendChild(newElement);
  }
}
```

The `await` expression completely pauses the function until the promise resolves, which lets you code this asynchronous function as if it's synchronous and avoiding the callback hell that normal fastdom needs.

I highly recommend waiting until ES7 finalizes, but I think this could be really useful in the future!

[You can try this live demo I hacked together](http://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=%2F%2F%20babel%20fiddle%20has%20a%20getter%20on%20these%20that%20throw%20an%20error%2C%0D%0A%2F%2F%20and%20fastdom%20tries%20to%20access%20them%3B%0D%0Adelete%20window.require%3B%0D%0Adelete%20window.exports%3B%0D%0Adelete%20window.module%3B%0D%0A%0D%0Afunction%20fwrite()%20%7B%0D%0A%20%20return%20new%20Promise(%20resolve%20%3D%3E%20fastdom.write(resolve)%20)%3B%0D%0A%7D%0D%0A%0D%0Afunction%20fread()%20%7B%0D%0A%20%20return%20new%20Promise(%20resolve%20%3D%3E%20fastdom.read(resolve)%20)%3B%0D%0A%7D%0D%0A%0D%0Aasync%20function%20test()%20%7B%0D%0A%20%20await%20fread()%3B%0D%0A%20%20var%20testDiv%20%3D%20document.getElementById('testDiv')%3B%0D%0A%20%20%0D%0A%20%20for%20(%20let%20i%20%3D%200%3B%20i%20%3C%201000%3B%20i%2B%2B%20)%20%7B%0D%0A%20%20%20%20await%20fwrite()%3B%0D%0A%0D%0A%20%20%20%20let%20newElement%20%3D%20document.createElement('div')%3B%0D%0A%20%20%20%20newElement.textContent%20%3D%20'hello%2C%20world!'%3B%0D%0A%20%20%20%20%0D%0A%20%20%20%20testDiv.appendChild(newElement)%3B%0D%0A%20%20%7D%0D%0A%7D%0D%0A%0D%0Aasync%20function%20setup()%20%7B%0D%0A%20%20var%20clicked%20%3D%20false%3B%0D%0A%20%20%0D%0A%20%20await%20fread()%3B%0D%0A%20%20var%20babelOutput%20%3D%20%24('.babel-repl-output')%3B%0D%0A%20%20%0D%0A%20%20await%20fwrite()%3B%0D%0A%20%20babelOutput.empty()%3B%0D%0A%20%20%0D%0A%20%20await%20fwrite()%3B%0D%0A%20%20%24('%3Cbutton%3EClick%20to%20test!%3C%2Fbutton%3E').click(function()%20%7B%0D%0A%20%20%20%20if%20(clicked)%20return%3B%0D%0A%20%20%20%20%0D%0A%20%20%20%20test()%3B%0D%0A%20%20%20%20clicked%20%3D%20true%3B%0D%0A%20%20%7D).appendTo(babelOutput)%3B%0D%0A%20%20%0D%0A%20%20await%20fwrite()%3B%0D%0A%20%20%24('%3Cdiv%20id%3D%22testDiv%22%20%2F%3E').appendTo(babelOutput)%3B%0D%0A%7D%0D%0A%0D%0Avar%20fastdomTag%20%3D%20document.createElement('script')%3B%0D%0AfastdomTag.src%20%3D%20'https%3A%2F%2Frawgit.com%2Fwilsonpage%2Ffastdom%2Fmaster%2Findex.js'%3B%0D%0A%0D%0AfastdomTag.onload%20%3D%20()%20%3D%3E%20setup()%3B%0D%0A%0D%0Adocument.head.appendChild(fastdomTag)%3B) in the Babel REPL! My example is a bit excessive as it has some unnecessary fastdom pauses, I just have those there to show it off.

erikdesjardins:
Babel, you say? [I've got just the thing.](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2301)
Although right now that PR has only the default transformers enabled (async functions are not).

Maybe they should be, though - the possibilities are impressive:

In module init functions:
```javascript
module.beforeLoad = async function() {
	await headReady();
	await fwrite();
	// add styles

	await bodyReady();

	// not necessary but this allows beforeLoad to resolve when everything is done
	await Promise.all(addBodyClasses(), addEventListeners());
};
async function addBodyClasses() {
	await fwrite();
	// stuff
}
async function addEventListeners() {
	await fread();
	// stuff
}
```
And in site modules (with some changes to `RESUtils.runtime.ajax`, and if `handleLink` only returns `info`):
```javascript
handleLink: async function(elem) {
	var groups = (/^https?:\/\/gist\.github\.com\/(?:[\w-]+\/)?([a-z0-9]{20,}|\d+)/i).exec(elem.href);
	if (!groups) {
		throw new Error('gist.github.com url is not valid: ' + elem.href);
	}

	var response = await RESUtils.runtime.ajax({
		method: 'GET',
		url: 'https://api.github.com/gists/' + groups[1],
		aggressiveCache: true
	});

	return JSON.parse(response.responseText);
},
// from a totally different module, just to demonstrate
handleInfo: async function(elem, info) {
	var template = await RESTemplates.load('TwitchUI');

	elem.type = 'GENERIC_EXPANDO';
	elem.expandoClass = ' video';
	elem.expandoOptions = {
		generate: () => template.html(info)[0],
		media: info
	};

	return elem;
}
```

empyrical:
Nice! I'm doing basically just that in a reddit browser extension I'm working on - async/await everywhere. A bit risky because the spec isn't finalized, but I think the convenience this syntax brings is worth the risk :) It's especially nice for browser extension development if you make promise wrappers for the browser's message passing API

erikdesjardins:
moved to #4151

PR #2239, opened 2015-05-19
support i.stack.imgur.com and .yahoo.imgur images, hit API for .jpg's

andytuba:
too many mislabelled .jpgs are out in the wild thanks to RES


vermiculus:
This seems to be broken in at least 4.5.4. [19] and [20] aren't showing inline on [this thread](https://redd.it/2zk1ae).

Issue #1481, opened 2014-07-31
code validation


andytuba:
We should set up some automated tests to address recurring issues with code standards.

## Concerns
* Mozilla concerns
    * prefer `$(...).safeHtml(...)` over `$(...).html(...)` over `....innerHTML`
* code style ( @mc10)
    * var at top of function

## Implementation ideas

What framework / mechanisms should we use to implement this?  

ideas:

* git post-commit hook
* grunt/gulp task on a cronjob or post-merge github hook

Issue #1112, opened 2014-05-20
Whitespace, menial syntax (semicolons) cleanup build task


andytuba:
ref #862

mc10:
The ESLint checking should be sufficient.

PR #2244, opened 2015-05-20
Add Travis task to run JSHint and JSCS over all files

mc10:
We can fine tune JSHint and JSCS so that they don't complain about problems that aren't, well, problems, but it's nice to get started with Travis. Works towards #1112 and #1481.

@honestbleeps I believe you have to flip Travis on for this project on their website: https://travis-ci.org/

EDIT: Added JSCS support.


andytuba:
:+1:

mc10:
We should honestly just decide to drop one or the other. I'm fine with either way but I guess that should probably be a separate issue.

andytuba:
Let's do grunt for now since you already put in the legwork and we can maybe migrate over eventually.

andytuba:
@honestbleeps Travis integration on this repo?  https://travis-ci.org/  looks like a pretty streamlined process if you login via github (top right)

PR #2248, opened 2015-05-21
changed futurism.co to futurism.com

dansutera:
to reflect the new domain, thanks!  please let me know if I did anything wrong...

### DIFF
--- a/lib/modules/hosts/futurism.js
+++ b/lib/modules/hosts/futurism.js
@@ -1,8 +1,8 @@
 modules['showImages'].siteModules['futurism'] = {
 	name: 'futurism',
-	domains: ['futurism.co'],
+	domains: ['futurism.com'],
 	detect: function(href, elem) {
-		if (href.indexOf('futurism.co') !== -1) {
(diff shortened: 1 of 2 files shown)


mc10:
Do old `futurism.co` links still redirect properly? If so, I think we should also keep support for links using the old domain.

dansutera:
Hi Kevin,

Thanks for looking at this!  We tested out the new changes and it works
both with futurism.com and with the old futurism.co links.  This is because
the old links are .jpg’s.  So yes, it supports both.

Cheers,

Dan

mc10:
My question is just in the `detect` function, shouldn't you reference the old links as well, using `href.indexOf('futurism.co') !== -1 || href.indexOf('futurism.com') !== -1`?

andytuba:
What @mc10 said -- leave the detect as-is, and make the hosts `[ 'futurism.com', 'futurism.co' ],

dsutera:
Hi guys, I just made the recommended changes and committed.  Tested and all looks good for both old and new.  Thanks for the help!

mc10:
Looks good to me!

Issue #2242, opened 2015-05-20
NER support for new search layout


andytuba:
https://www.reddit.com/r/beta/comments/36k0u3/beta_update_519_new_search_page/

This will probably sit in beta for a little bit, but we should get on it.

andytuba:
Note that subreddit search has been mixed into the main search. If you do not restrict searching by a particular subreddit or post listing, then search results show two separate lists: subreddits and posts.  I envision only applying NER to the posts list on the mixed search (but make sure to apply NER on /subreddits/search since it's a single listing).

andytuba:
1c51413

PR #2268, opened 2015-06-07
Never Ending Reddit quits without a #siteTable

andytuba:
At least don't break the search page before properly taking care of #2242

### DIFF
--- a/lib/modules/neverEndingReddit.js
+++ b/lib/modules/neverEndingReddit.js
@@ -70,30 +70,13 @@ modules['neverEndingReddit'] = {
 	},
 	beforeLoad: function() {
 		if ((this.isEnabled()) && (this.isMatchURL())) {
-			RESTemplates.load('neverEndingReddit-CSS', function(template) {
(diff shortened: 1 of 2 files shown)


lib/modules/neverEndingReddit.js
```diff
@@ -70,9 +70,31 @@ modules['neverEndingReddit'] = {
...
+
+			// store access to the siteTable div since that's where we'll append new data...
+			var stMultiCheck = document.querySelectorAll('#siteTable');
+			this.siteTable = stMultiCheck[0];
+			// stupid sponsored links create a second div with ID of sitetable (bad reddit! you should never have 2 IDs with the same name! naughty, naughty reddit!)
+			if (stMultiCheck.length === 2) {
+				// console.log('skipped first sitetable, stupid reddit.');
+				this.siteTable = stMultiCheck[1];
+			}
+			if (!this.siteTable) {
```
andytuba: This can be removed as soon as RES can handle the new post listing layout on the search page.

PR #2280, opened 2015-06-12
[HOLD] remove toolbarfix

andytuba:
It's been swell, except for all the bug reports where people thought RES owned the toolbar.

http://github.com/reddit/reddit/commit/38928f110d1a

await toolbar removal on June 26


mc10:
You can probably merge this in... Toolbar should be gone.

Issue #2279, opened 2015-06-12
Automatic night mode broken? 


andytuba:
There have been several reports of this. 

http://www.reddit.com/r/Enhancement/comments/39fccw/how_do_i_customise_night_mode/

mc10:
Huh, has anything changed recently? It's odd that it'd break so randomly, but I'll take a look.

andytuba:
I've seen the occasional report over the past few months.

andytuba:
https://www.reddit.com/r/RESissues/comments/2gaact/bug_automatic_night_mode_has_never_worked_for_me/?ref=search_posts

andytuba:
potential: https://www.reddit.com/r/RESissues/comments/347cgt/proper_way_of_toggling_night_mode_when_automatic/?

PR #2282, opened 2015-06-12
Improve night mode

mc10:
Hopefully addresses problems in #2279 at some point, but until I figure out the problem I'll keep fixing some minor issues.

### DIFF
--- a/lib/modules/nightMode.js
+++ b/lib/modules/nightMode.js
@@ -60,9 +60,9 @@ addModule('nightMode', function(module, moduleID) {
 			description: 'Allow the subreddits listed to display subreddit styles during night mode if useSubredditStyles is disabled.'
 		}
 	},
-	onToggle: function(onOrOff) {
+	onToggle: function(toggle) {
 		// If night mode is turned off, disable night mode completely
-		if (!onOrOff) {
+		if (!toggle) {
 			module.disableNightMode();
 		}
 	},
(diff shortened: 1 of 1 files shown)


andytuba:
Thanks for clean-up!  While you're in there, there's a report of some weirdness with nightmode setting the localStorage key on not-www.reddit.com (and not getting cleared out when turning off NM?)

mc10:
Hmm maybe that's the problem... perhaps localStorage is getting trashed across subdomains? I'm assuming we still use localStorage for our own RESStorage.
* http://stackoverflow.com/questions/4026479/use-localstorage-across-subdomains
* http://stackoverflow.com/questions/4177803/can-html5-databases-and-localstorage-be-shared-across-subdomains

andytuba:
As I understand it, RES storage comes from a local storage

andytuba:
Nevermind that comment, will sort it out tomorrow from desktop

andytuba:
localStorage is siloed by subdomain. Since reddit.com redirects to www.reddit.com (for their SSL cert, I believe), we can't take advantage of localStorage in reddit.com (no subdomain).  




Yeah, Chrome RES' RESStorage is populated from the extension's localStorage in chrome-extension://kbmf...: 
screen shot 2015-06-14 at 12 35 56 pm

andytuba:
I think the "nightmode stuck on subdomains" could be resolved by turning off nightmode and removing the tab's localStorage as soon as RESStorage loads to affirm "no night mode isn't active right now after all".. probably a little flicker of nightmode styling which might confuse the user or other modules which pull values from CSSS ... but hopefully modules.nightMode can win the race

mc10:
I can't figure out the ANM issues, but I'll merge this in anyways as a minor improvement.

PR #2287, opened 2015-06-14
Migrate options from userTagger to userInfo

### LINKED ISSUE #2252: moveOption() migrators don't work with removed options
The issue appears to lie [here](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/a9bbe83ece42ad6d3f1f3c4209405cb328e02434/lib/core/options.js#L280-L291), in `getOptions()` - it only saves the values of options that are in the code, so the values of any options in storage but not in the code will be unaccessible.

Although it's been this way for a long time (since at least [v4.5.0.2](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/v4.5.0.2/lib/utils.js#L113-L125)), so either something else is contributing to the issue or `moveOption()` migrators haven't worked for a few releases (which seems unlikely).

image

### LINKED ISSUE #2262: migrate userInfo options from userTagger
looks like i forgot to add them.

erikdesjardins:
Fixes #2262, edit: fixes #2252

Also fix an issue preventing number primitives from being migrated, and ensure that we always set default values.


andytuba:
Thanks! Does  #2252 affect this?

erikdesjardins:
Yes, so these options won't successfully migrate in master, but they do work if I add a hacky fix for #2252.

andytuba:
Do you mean hack where "migrate calls loadOptions with a parameter to 'load all the stored options'"? I think that could still work, especially if the migrator framework deleted the options not in code.  (Maybe the "load stored options" routine could add a "obsolete" flag to options not in code.)

erikdesjardins:
This should now fix #2252, but if others could test, that would be good. (Just migrate options from the 4.5.4 release build)

PR #2293, opened 2015-06-16
L=C doesn't get set properly on search pages

erikdesjardins:
Fixes an issue [from reddit.](https://www.reddit.com/r/RESissues/comments/39zqd3/bug_single_click_opener_treats_self_posts/)

This'll be fine as long as nobody manages to guess their post's ID (good luck).


andytuba:
Hmm, I wonder if we'll get this released before reddit pushes the [new search layout](http://i.imgur.com/UBKkUJ5.png).. seems like a good long-term solution, though.

PR #2297, opened 2015-06-16
Fix various JSHint errors

mc10:
Most of these are simple ones like adding semicolons and converting double quotes to single quotes.


andytuba:
thanks, sorry I've been ignoring JSHint on my commits.

PR #2299, opened 2015-06-17
JSHint cleanup

mc10:
Various QOL cleanups across the board.

EDIT: This is now good to go, but this definitely needs another pair of eyes (at the very least).

### DIFF
--- a/lib/core/storage.js
+++ b/lib/core/storage.js
@@ -1,5 +1,6 @@
-var RESStorage = RESStorage || {};
+/* exported RESStorage, RESLoadResourceAsText */
 
+var RESStorage = RESStorage || {};
 
 RESStorage.setup = function() {
 	if (!RESStorage.setup._complete) {
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -775,7 +775,7 @@ RESUtils.firstValid = function() {
 	for (var i = 0, len = arguments.length; i < len; i++) {
 		var argument = arguments[i];
 
-		if (argument === void 0) continue;
+		if (argument === undefined) continue;
 		if (argument === null) continue;
 		if (typeof argument === 'number' && isNaN(argument)) continue;
 
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -776,7 +776,7 @@ RESUtils.getHeaderOffset = function() {
 				case 'header':
 					header = document.getElementById('header');
 					break;
-				case 'none':
+				// case 'none':
 				default:
 					break;
 			}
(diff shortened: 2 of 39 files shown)


lib/core/utils.js
```diff
@@ -3,6 +3,29 @@ $.fn.safeHtml = function(string) {
 	else return $(this).html(RESUtils.sanitizeHTML(string));
 };
 
+if (typeof Array.prototype.find !== 'function') {
+	Array.prototype.find = function(predicate) {
```
andytuba: :+1:

lib/modules/selectedEntry.js
```diff
@@ -168,27 +166,26 @@ addModule('selectedEntry', function(module, moduleID) {
 		}
 	}
 
+	function scrollTo(selected, last, options) {
+		if (!selected) return;
+		options = $.extend(true, {}, {
+				makeVisible: selected.entry,
```
erikdesjardins: Very minor but the comma on this line and the empty object can be removed.

erikdesjardins:
I didn't look extra hard, but everything else seems to look good.

lib/modules/hosts/imgur.js
```diff
@@ -38,20 +38,16 @@ modules['showImages'].siteModules['imgur'] = {
...
 						def.resolve(json.data);
 						return;
 					}
 				} catch (error) {
-					// debugger;
 					def.reject(error);
 					return;
 				}
-				// debugger;
 				def.reject(response);
```
erikdesjardins: This function could do with even more cleanup: this line is unreachable, and the return statements are somewhat redundant.

mc10:
Done, thanks for the suggestions.

lib/core/storage.js
```diff
@@ -1,5 +1,6 @@
-var RESStorage = RESStorage || {};
+/* exported RESStorage, RESLoadResourceAsText */
```
andytuba: Some day we'll corral all of these into `RES.thing...`

lib/core/utils.js
```diff
@@ -653,15 +699,19 @@ RESUtils.getHeaderOffset = function() {
...
-		if ((obj[i]) && (obj[i].value === value)) {
-			obj[i].selected = true;
-		}
+	// Only call Array.prototype.slice.call if obj is array-like
+	// e.g. getElementsByTagName, but *not* getElementById
+	if (obj.length !== undefined) {
+		obj = Array.prototype.slice.call(obj);
+	} else {
+		obj = [obj];
 	}
```
andytuba: We should probably replace this with `$(obj).val(value)`

andytuba:
note to self: pick up again at accountSwitcher

lib/core/utils.js
mc10: That's a good point, I'll do that.

mc10:
Rebased to the latest changes on `master`.

andytuba:
phew, what a day.  Hope all those `!== null`'s work out as expected.

PR #2301, opened 2015-06-18
there can be only one (build script)

erikdesjardins:
Compiles `.js` with Babel and `.scss` with Sass + Autoprefixer.
Build times are very quick: ~10s for a full build and <500ms for one file.

Removes Grunt and `makelinks.sh`.
Also removes OperaBlink from the build because it was exactly the same output as Chrome.

ESLint replaces JSHint and JSCS since they do not support ES7. Currently, all rules ESLint supports are listed in `.eslintrc`, and failing rules are commented out. They'll be uncommented as code is brought up-to-spec in future PRs.


andytuba:
This is cool!  Let's target this for just after 4.6.1 (i.e. new major release, hotfix release, then merge).

erikdesjardins:
Sounds good!
Expect a PR for Sass, too, after this one is merged.

erikdesjardins:
@mc10 This will replace JSCS and JSHint with ESLint, since it's the only linter that supports ES7.

It supports every rule they do (and more), aside from `requireDollarBeforejQueryAssignment`, so it shouldn't be too disruptive to #2303. Currently all of the failing rules in `.eslintrc` are commented out.

Ajedi32:
Wait, does this mean I can use ES6/7 features in my code and it'll just work? It would be sweet if I could refactor #2531 to use `async/await`.

erikdesjardins:
@Ajedi32 Yep, but you'll probably want to wait for the refactoring to be done - I'm currently converting all of our APIs to promises, and some other stuff - like this:
image
I'll ping you when I open the PR.

Ajedi32:
@erikdesjardins Ah, that's cool. A batch API would definitely simplify things a bit in my PR. Let me know when the PR is ready so I can take a look. :+1:

PR #2303, opened 2015-06-20
[WIP] jscs: Use Node style guide as a preset

mc10:
This will take a long while to do (especially as it depends on #2299), so this won't be ready for quite some time.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -4,6 +4,5 @@
 	"requireCurlyBraces": ["if", "else", "for", "while", "do", "try", "catch"],
 	"requireDollarBeforejQueryAssignment": true,
 	"requireTrailingComma": null,
-	"validateIndentation": "\t",
-	"validateQuoteMarks": "'"
+	"validateIndentation": "\t"
 }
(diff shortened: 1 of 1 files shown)


andytuba:
:+1: for tooling

erikdesjardins:
Another :+1:, it's gonna be awesome when we have all PRs linted with this.

What do you think about turning off `requireCapitalizedComments` (IMO it makes short comments look worse and it's too nitpicky) and maybe not requiring braces on case/default statements? (they're verbose enough with `break;`)

That would be: `"requireCurlyBraces": ["if", "else", "for", "while", "do", "try", "catch"]`.

andytuba:
:+1: on erik's suggestion

mc10:
Personally I'm more of a fan of all-caps comments but it'd be more practical not to require them.

erikdesjardins:
Also I'd be fine with [requireDollarBeforejQueryAssignment](http://jscs.info/rule/requireDollarBeforejQueryAssignment.html), if you want to enforce that.

I don't think it will catch stuff like this, though:

    var $foo = $('#foo');
    var notPrefixed = $foo.find('.bar');

mc10:
Thanks! It's better than nothing, I'd say.

PR #2313, opened 2015-06-28
SelectedEntry, KeyboardNav support for new search layout

### LINKED ISSUE #2243: KeyboardNav support for new search layout
https://www.reddit.com/r/beta/comments/36k0u3/beta_update_519_new_search_page/

This will probably sit in beta for a little bit, but we should get on it.

andytuba:
Fixes #2243

Anything that includes 'linklist' now needs to explicitly include 'search'

Includes `RESUtils.thing(elementInsideThing)` which provides `.thing`, `.entry`, and getter methods for various elements in a thing

Might also fix a problem with /r/{{subreddit}}/wiki/submit whichI think @hermithome mentioned


andytuba:
not exhaustively tested

PR #2316, opened 2015-06-30
Fix navTop FOUC

Shraymonks:
This change removes the FOUC of the right header by moving the `navTop` CSS initialization to `beforeLoad`.


lib/modules/styleTweaks.js
```diff
@@ -412,6 +408,7 @@ modules['styleTweaks'] = {
 		}
 	},
 	navTop: function() {
+		RESUtils.bodyClasses.add('res-navTop');
 		RESUtils.addCSS('#header-bottom-right { top: 19px; border-radius: 0 0 0 3px; bottom: auto;  }');
 		RESUtils.addCSS('.beta-notice { top: 48px; }');
 		$('#header-bottom-right').addClass('res-navTop');
```
andytuba: Could you wrap this in an `RESUtils.init.await.bodyReady(function() { ... }` ?

andytuba:
Good idea, thanks!

PR #2319, opened 2015-07-02
showViewImagesTab=false should prevent tab from rendering

andytuba:
https://www.reddit.com/r/RESissues/comments/3btl2l/bug_showviewimagestab_should_prevent_the_view/


andytuba:
That is a heinous hack. Key nav should not use a fake click to trigger RES functionality. I'll just tidy that up.. But I thought when I tested it that shift x still worked okay

andytuba:
3e0e7b079752b6877ef21c924ba9286d80b153a7

PR #2318, opened 2015-07-02
Optionally add ?feature=legacy_search to search forms

andytuba:
https://www.reddit.com/r/Enhancement/comments/3bswno/can_res_add_featurelegacy_search_into_search/


erikdesjardins:
Do you remember if you wrote this before the reddit pref? If so, I think we should probably revert it.

andytuba:
I think I wrote it just before the reddit preference was added, yeah.

However, this is still useful for people browsing anonymously.

erikdesjardins:
True. I didn't consider that.

PR #2333, opened 2015-07-06
utils: Use event.target instead of eventTarget

mc10:
`eventTarget` doesn't exist, and `event.target` seems like the best replacement.


andytuba:
thanks for catching that.. I musta just autocompleted without thinking.

PR #2334, opened 2015-07-06
Remove duplicate buttons on new search page

githue:
Prevents this from happening: 
reddit-search-dupe-btns

[Check for yourself.](https://www.reddit.com/r/funny/search?q=search&sort=relevance&t=all)

It's odd that `.side` wasn't specified in the first place, is there a reason for that?

### DIFF
--- a/lib/modules/dashboard.js
+++ b/lib/modules/dashboard.js
@@ -867,7 +867,7 @@ modules['dashboard'] = {
 		};
 	},
 	addDashboardShortcuts: function() {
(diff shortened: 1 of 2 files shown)


andytuba:
Does /subreddits/search still work the same? That may have been why not selecting for `.side` before.

githue:
No issue on /reddits and /subreddits because subredditManager deals with that page individually.

mc10:
Everything looks good to me, thanks!

PR #2338, opened 2015-07-07
Minor lint fix

erikdesjardins:
Even with three of us looking it over, we still missed something -.-
(hopefully only one thing...)


mc10:
Whoops thanks for that.

PR #2346, opened 2015-07-08
utils: Simplify pageType calls

mc10:
Removed the performance simplification, so all of these are just style changes.

### DIFF
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -290,7 +290,7 @@ RESUtils.pageType = function() {
 		pageType = 'search';
 	} else if (RESUtils.regexes.profile.test(currURL)) {
 		pageType = 'profile';
-	} else if ((RESUtils.regexes.comments.test(currURL)) || (RESUtils.regexes.friendsComments.test(currURL)) || (RESUtils.regexes.nosubComments.test(currURL))) {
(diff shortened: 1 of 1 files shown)


lib/core/utils.js
```diff
@@ -278,42 +278,47 @@ RESUtils.isReddit = function() {
...
 };
 RESUtils.isPageType = function(/*type1, type2, type3, ...*/) {
-	var thisPage = RESUtils.pageType();
-	return Array.prototype.slice.call(arguments).some(function(e) {
-		return (e === 'all') || (e === thisPage);
+	var href = document.location.href;
+	return Array.prototype.slice.call(arguments).some(function(type) {
+		return (type in RESUtils.regexes &&
+			RESUtils.regexes[type].test(href)) ||
+			type === 'linklist';
```
erikdesjardins: Eh, I can't see this being anything but slower: `isPageType()` gets called at least once per module, so it'll end up doing way more regex tests than the previous implementation.

Otherwise this is just a slightly more complex way to do it, so I don't think `isPageType()` needs to be changed. (If nothing else, this implementation matches `linklist` on every page, so that needs to be fixed.)

mc10: The original code matched `linklist` as well (you can check the `pageType` code), but I agree, I'll revert these changes.

PR #2349, opened 2015-07-09
Compatibility changes

mc10:
In particular, this removes the `Function.prototype.bind` polyfill that we should no longer need, given that Safari 5.1 is far, far away.


andytuba:
I suspect there's about 2 users who still use Safari 5.1, and they're probably on crappy old hardware and can't upgrade to any other browse that we do support.  :feelsgood:

PR #2351, opened 2015-07-10
Increase max # of rows per dashboard widget

erikdesjardins:
Reddit can serve up to 100 entries per request, so why not?

It won't be fun to manually increment it, but I guess you can always set the default to what you want and then add widgets.

[via reddit](https://www.reddit.com/r/Enhancement/comments/3cstep/feature_request_change_the_widget_cap_to_a_higher/)

### DIFF
--- a/lib/modules/dashboard.js
+++ b/lib/modules/dashboard.js
@@ -83,6 +83,7 @@ modules['dashboard'] = {
 			}
 		}
 	},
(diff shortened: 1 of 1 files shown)


mc10:
Looks good to me, but if anyone complains about the `MAX_ROWS` we can always tweak it later.

PR #2352, opened 2015-07-11
Get nightmode.css up to date

githue:
Tested in both legacy_search and the new search page.

### DIFF
--- a/lib/modules/nightmode.css
+++ b/lib/modules/nightmode.css
@@ -354,7 +354,7 @@ body.res-nightmode,
 .res-nightmode .dropdown.lightdrop a.choice:hover {
 	background-color: #111;
 }
-.res-nightmode .side {
+.res-nightmode .guiders_button {
 	color: #fff;
 }
 .res-nightmode .sidebox,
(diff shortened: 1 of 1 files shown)


lib/modules/nightmode.css
```diff
@@ -715,6 +715,9 @@
 .res-nightmode #quickMessage select {
 	background-color: #aaa;
 }
+.res-nightmode #search input[type=text] {
+	color: rgb(255,255,255) !important;
```
githue: Had to use !important to override reddit's inline style of `color:white`.

githue:
I decided to go through and update as many styles as I could, there were so many little things that had to be done.

One thing I'm not sure about was removing the unique text colors on domains and timestamps, tell me if they need to be put back in. I thought they were unnecessary, distracting, and created more work for moderators to change.

andytuba:
If anyone complains about the domain/timestamp disappearing, then I'll add it to /r/resupdates/about/stylesheet and they can opt in to that feature by adding a body class.

PR #712, opened 2014-01-13
Backup & Restore Module

thybag:
Hello,

Hopefully I'm not being to annoying putting in all these random pull requests. 

Had a play with creating a 1 click backup & restore module for RES (as described in https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/300 ). Currently its a bit rough around the edges + I've yet to fully put it through its paces in terms of cross browser testing, but i thought i may as well put in a pull for visibility even if its decided eventually not to include the functionality (due to cross browser issues/support headaches it could cause).

It utilities- the data URI's for downloading the file (just stringified JSON) + the filereader API to read the file back in. Everything appears to be working in chrome.

PR #2353, opened 2015-07-12
#712 (Backup & Restore) updated/cleaned up

erikdesjardins:
Closes #712.

Someone should test this in Safari.


lib/modules/backupAndRestore.js
```diff
@@ -0,0 +1,78 @@
...
+					return;
+				}
+				// Reimport data
+				for (var k in settings) {
+					RESStorage.setItem(k, settings[k]);
+				}
+
+				alert('Your RES settings have been imported. Reloading reddit.');
+				location.hash = '';
+				setTimeout(location.reload.bind(location), 500);
```
mc10: Is the `bind` here necessary?

erikdesjardins: It is, apparently:
image

lib/modules/backupAndRestore.js
```diff
@@ -0,0 +1,78 @@
...
+		}
+	});
+
+	function backup() {
+		// Generate copy of RES settings for download
+		var settings = JSON.stringify(RESStorage);
+		var link = document.createElement('a');
+		if (BrowserDetect.isChrome()) {
+			// Encoding can make the URI long enough to crash Chrome; it doesn't appear to care if it's unencoded
+			link.href = 'data:text/plain;charset=utf-8,' + settings;
```
JordanMilne: You can get around this by [storing the json in a `Blob`](http://jsfiddle.net/koldev/cw7w5/) and setting `.href` to the `Blob`'s URI instead.

erikdesjardins: Ooooh, that's perfect.

PR #2354, opened 2015-07-12
.gitignore: remove IE folders, exclude IntelliJ .iml files

erikdesjardins:


andytuba:
Huh, I thought mc10 cleared that out last year.

mc10:
Oh we deleted the files but I remember sticking those folders in gitignore to make sure they'd never come back.

PR #2355, opened 2015-07-12
Remove my sourceSnudown unescape hack

erikdesjardins:


lib/modules/sourceSnudown.js
```diff
@@ -95,7 +83,7 @@ addModule('sourceSnudown', function(module, moduleID) {
 					var thisResponse = JSON.parse(response.responseText),
 						sourceText = null,
 						userTextForm = $('<div class="usertext-edit viewSource"><div><textarea rows="1" cols="1" name="text" readonly></textarea></div><div class="bottom-area"><div class="usertext-buttons"><button type="button" class="cancel">hide</button></div></div></div>');
```
mc10: You should probably just attach a `$` in front of the variable name to distinguish it as a jQuery variable.

erikdesjardins: Yeah I was gonna do that but like half of the variables in that function are jQuery.
But I guess it couldn't hurt.

I wonder if JSCS can enforce this kind of prefixing.

PR #2358, opened 2015-07-13
Improved CSS font declarations on Batch icons.

githue:
To reduce the chance of batch icons inheriting undesirable styles, and to save a few lines, use the font shorthand property. Also removed a bold declaration from the 'to-top' icon because it was almost unrecognisable.

For a specific example see the sidebar at www.reddit.com/r/csshelp and notice the italic usernames and tag icons. This change helps to ensure the icons don't inherit properties like italic, bold, and others.

Further explanation here: https://developer.mozilla.org/en/docs/Web/CSS/font

How the font declaration looks when expanded:
font-shorthand-css

In this case 'normal' works as a catch-all for style, weight, variant and stretch.

The minimum amount of properties you can put in `font` are size and family, and Batch icons should always define both anyway.


andytuba:
you guys are so great

PR #2360, opened 2015-07-14
Comment box falling below sidebar

### LINKED ISSUE #2359: Naut working too well with RES
http://redd.it/3cxhm3

1. Have commentingAs enabled.
2. Have macroButtons disabled.
3. Watch textarea disappear.

I fixed this same issue for an older version of Naut way back when: #1452, but it seems to have come back. There are probably several ways to solve this, but I want to find the magic bullet so it never returns.

~~It may be related to this separate issue which I reproduced a few days ago, but now can't (release version of RES). http://redd.it/3ccutb~~ Don't know about this one.

githue:
Here's my detailed analysis of this issue and how I found a solution: http://output.jsbin.com/wemonaw/

As an extra safeguard for the future, it will be important to be very careful with using floats inside form.usertext. Try to avoid using floats when they're not needed, for example div.commentingAs only needs to be floated when span.RESMacroWrappingSpan is present, so I've added an if statement to check for the latter.

This PR is the closest I can get to a permanent solution to this long-standing problem, it will be interesting to see how it holds up in  the wild.

Fixes #2359

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -346,12 +346,18 @@ body.wiki-page form#editform, form.usertext {
 div.usertext-edit {
 	width: auto;
 	max-width: 500px;
+	padding: 0px;
+}
+/* Correct reddit's miscalculation of textarea width when editing a comment. */
+.usertext-body > .md {
+	padding-right: 1px;
 }
 .usertext-edit .md,
 .wiki-page #editform textarea {
 	clear: left;
 }
-/* Fix wiki edit form. */
+/* Improve size calculation for textareas. */
+.usertext-edit textarea,
 #wiki_page_content,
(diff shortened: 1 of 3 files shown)


githue:
Forgot to mention that only Chrome/webkit seems to need this fix, Firefox doesn't show the same issue afaik.

githue:
Update: after more testing I discovered that right floats don't trigger this bug, so I've made a new commit based on new knowledge. I'm very confident that this new commit will solve the issue for good.

[The test case is also updated.](http://output.jsbin.com/wemonaw/)

githue:
I just discovered this isn't exclusive to naut, in fact the same problem has always existed on vanilla reddit when replying to another comment (not root comments). RES was already taking steps to avoid the issue, but it wasn't perfect.

PR #2361, opened 2015-07-16
filteReddit support for new search layout

andytuba:
Added/refactored a bunch of "get attributes/elements of this post/comment" to RESUtils.thing

filteReddit uses RESUtils.thing().

Needs testing for syntax errors, works on new search layout, still works on old search layout and other post listings, and will work on inbox comments.  Probably need to regression test other things that use changed RESUtils.thing methods.

### DIFF
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -366,9 +366,9 @@ RESUtils.$things = function(container) {
 	return $(container).find(RESUtils.thing.prototype.thingSelector);
 };
 RESUtils.things = function(container) {
-	return RESUtils.$things(container)
-		.map(function(index, ele) {
-			return RESUtils.thing(ele);
+	return RESUtils.$things(container).get()
+		.map(function(ele, index) {
+			return new RESUtils.thing(ele);
 		});
 };
 RESUtils.thing = function(element) {
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -366,6 +366,9 @@ RESUtils.things = function(container) {
 	return $(container).find(RESUtils.thing.prototype.thingSelector);
 };
 RESUtils.thing = function(element) {
+	if (element instanceof RESUtils.thing) {
+		return element;
+	}
 	if (!(this instanceof RESUtils.thing)) {
 		return new RESUtils.thing(element);
 	}
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -408,7 +408,7 @@ RESUtils.thing.prototype = {
 	},
 
 	isPost: function() {
-		return this.entry.classList.contains('link') || this.entry.classList.contains('search-result-link');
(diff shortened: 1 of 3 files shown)


lib/core/utils.js
```diff
@@ -361,11 +361,20 @@ RESUtils.thingsContainer = function(body) {
 
 	return container;
 };
-RESUtils.things = function(container) {
+RESUtils.$things = function(container) {
 	container = container || RESUtils.thingsContainer();
 	return $(container).find(RESUtils.thing.prototype.thingSelector);
 };
+RESUtils.things = function(container) {
```
mc10: I'm a bit confused... what's the difference between these two functions? Is there any reason why we'd call the first?

lib/modules/filteReddit.js
```diff
@@ -769,21 +760,16 @@ modules['filteReddit'] = {
 			(RESUtils.currentSubreddit('all')) || (RESUtils.currentDomain()) || (RESUtils.currentMultireddit('me/f/all')),
 			onSavedPage = modules['filteReddit'].excludeSaved.test && modules['filteReddit'].excludeSaved.test(location.href);
 
-		Array.prototype.slice.call(entries).forEach(function(entry) {
-			var postSubreddit, currSub;
+		Array.prototype.slice.call(things).forEach(function(thing) {
```
mc10: You probably don't need the `Array.prototype.slice.call` part, since `.things` should return an array.

lib/core/utils.js
mc10: Also, maybe call `.get` here to convert it into a normal array?

lib/modules/filteReddit.js
andytuba: I changed `RESUtils.things` to return an array of plain objects and added `RESUtils.$things` to return a jQuery object.  So, I should just be able to call `.forEach` :+1:

lib/core/utils.js
andytuba: `.things` returns an array of thing objects (the view wrapper for things elements with accessors for various properties/elements); `.$things` returns a jquery set of elements which can be wrapped in `.things`.  Does that make sense or should we pick some different names?

andytuba: and yeah, this really ought to return a normal array

mc10:
Also you might want to check the JSHint errors--you might have introduced a few bugs.

lib/core/utils.js
mc10: Do we ever need `$things`? I feel like we'd only ever need the array of the elements all wrapped (though I guess I didn't look too closely to see where we used it in code).

andytuba: It's only used in NER so I didn't have to rewrite .thing or a lot of code that expects a regular jQuery element set.

andytuba:
wouldn't surprise me if this breaks things.

PR #2372, opened 2015-07-24
Universal class for subreddit-specific toggle buttons

githue:
subredditInfo.js relied on filteReddit.js to apply the correct style to the subscribe button. This PR creates a universal class based on fancy-toggle-button which is module-independent, and removes doubled-up CSS.

I named the class res-subreddit-btn because all the relevant buttons perform subreddit actions: subscribe, shortcut, dashboard, filter. This doesn't touch the +friends button because that uses reddit's class names.

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -795,15 +795,16 @@ p.moduleListing {
 .res-subreddit-btn {
 	font: bold 10px verdana, sans-serif; /* allow line-height to be inherited */
 	color: white;
+	display: inline-block;
 	margin: 0 5px 5px 0;
 	padding: 1px 6px;
 	border: 1px solid rgb(68,68,68);
-	background: rgb(123,184,80) url(https://redditstatic.s3.amazonaws.com/bg-button-add.png) 0 100% repeat-x;
+	background-image: linear-gradient(to bottom, rgb(123,184,80), rgb(117,168,73));
 	border-radius: 3px;
 	cursor: pointer;
 }
 .res-subreddit-btn.remove {
-	background: rgb(207,97,101) url(https://redditstatic.s3.amazonaws.com/bg-button-remove.png) 0 100% repeat-x;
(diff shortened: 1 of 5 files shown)


andytuba:
awyeah.jpg

lib/core/res.css
```diff
@@ -790,6 +790,21 @@ p.moduleListing {
...
 }
+/* Universal subreddit toggle button (subscribe, filter, shortcut, etc) */
+/* Borrowed from fancty-toggle-button */
+.res-subreddit-btn {
+	font: bold 10px verdana, sans-serif; /* allow line-height to be inherited */
+	color: white;
+	margin: 0 5px 5px 0;
+	padding: 1px 6px;
+	border: 1px solid rgb(68,68,68);
+	background: rgb(123,184,80) url(https://redditstatic.s3.amazonaws.com/bg-button-add.png) 0 100% repeat-x;
```
andytuba: I'd love to take this opportunity to remove that background-url.

githue:
I'll add the old class back as a legacy thing, the themes I checked didn't show any issues.

lib/core/res.css
githue: We can just replace it with linear-gradient, I don't know why there was an image in the first place.

githue:
Naut is back to normal. I went through a bunch of style-heavy subs and found two that still look different: /r/askreddit and /r/wow. The former can't be helped because it uses inelegant techniques to position the buttons, and the latter just looks bigger than it should because it's forcing the line-height.

Also replaced the image URL with a linear-gradient.

githue:
Is there a better name for the class instead of res-subreddit-button? What about res-fancy-toggle-btn to convey that it's meant to emulate reddit's buttons?

erikdesjardins:
+1 for res-fancy-toggle-button

andytuba:
res-fancy-toggle-button would be cool, a little more general-use than subreddit-button.

PR #2377, opened 2015-07-27
Big Nightmode Update

githue:
* Makes nightmode compatible with different modules and reddit features
 * e.g. Visited link color, sticky color, username colors with username highlighter disabled.
* Removes duplicate and unnecessary styles
 * The color of the NSFW tag was set in two places, yet it doesn't need to be included at all
 * Alternating background colors on link listings isn't something reddit or RES does by default, why should nightmode introduce it? Should be an option in styletweaks that works with and without nightmode.
* Removes overreaching styles
 * e.g. selectedEntry should only add BG color on .md-container, not every .md inside the container.
* Groups more selectors to reduce number of lines
* Adds new styles that were missing. e.g. white drop shadows, black text on black, etc
* Doesn't break existing nightmode themes like /r/space, /r/askreddit and /r/askscience. However /r/videos looks quite different, but that's because their stylesheet is actually having an effect. :/

Give it a twirl and see what you think.

### DIFF
--- a/lib/modules/nightMode.js
+++ b/lib/modules/nightMode.js
@@ -136,9 +136,6 @@ addModule('nightMode', function(module, moduleID) {
 
 		// If night mode is on and the sub isn't compatible, disable its stylesheet.
 		if (module.isNightModeOn() && !module.isNightmodeCompatible) {
-			// Hide header images since sub isn't night mode compatible and therefore they
-			// may be bright images, etc.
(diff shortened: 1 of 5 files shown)


andytuba:
I'm inclined to go for broke and merge this, then use /r/resupdates to hotpatch any major bugs reported.

githue:
/r/resupdates?

andytuba:
Check out the new Stylesheet module's options -- RES loads its subreddit stylesheet everywhere by default.

andytuba:
@andytuba TODO changelog

PR #2380, opened 2015-07-27
Youtube regex fix

### LINKED ISSUE #2379: Youtube (non-link) expandos are easily fakable
Due to the use of regexes in the handling of youtube expandos, it's easy to create youtube links that, when viewed using an expando show one thing, but when clicked on directly show another. I made a test post outlining how this works: https://www.reddit.com/r/puckipedia_tests/comments/3erl0x/res_youtube_breakage/

(note: this does not work on link expandos, because those use the oembed metadata in the pages)

erikdesjardins:
Fixes #2379

### DIFF
--- a/lib/modules/hosts/youtube.js
+++ b/lib/modules/hosts/youtube.js
@@ -8,7 +8,7 @@ modules['showImages'].siteModules['youtube'] = {
 	handleLink: function(elem) {
 		var def = $.Deferred();
 
-		var hashRe = /^https?:\/\/(?:www\.|m\.)?youtube\.com\/watch\?.*v=([\w\-]+)/i;
+		var hashRe = /^https?:\/\/(?:www\.|m\.)?youtube\.com\/watch.*?[?&]v=([\w\-]+)/i;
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/youtube.js
```diff
@@ -8,7 +8,7 @@ modules['showImages'].siteModules['youtube'] = {
 	handleLink: function(elem) {
 		var def = $.Deferred();
 
-		var hashRe = /^https?:\/\/(?:www\.|m\.)?youtube\.com\/watch\?.*v=([\w\-]+)/i;
+		var hashRe = /^https?:\/\/(?:www\.|m\.)?youtube\.com\/watch.*?[?&]v=([\w\-]+)/i;
```
mc10: Think we can narrow down the highlighted part a bit more, maybe with `\/watch\?(?:[^?&]+&)*?v=([\w\-]+)`. I don't know how valuable that would be though.

mc10:
Huh after playing around with it my regex seems a bit brittle if the URL isn't perfect, like `https://www.youtube.com/watch?&av=b1WWpKEPdT4&v=dQw4w9WgXcQ` would fail to work properly.

PR #2384, opened 2015-07-28
User Highlight priority should be: friend > OP > mod > etc...

githue:
userhighlight-priority

Priority of colors, as seen on vanilla reddit.

### DIFF
--- a/lib/modules/userHighlight.js
+++ b/lib/modules/userHighlight.js
@@ -161,29 +161,29 @@ modules['userHighlight'] = {
 		this.options['generateHoverColors'].callback = modules['userHighlight'].generateHoverColors;
 		if ((this.isEnabled()) && (this.isMatchURL())) {
 			this.colorTable = {
+				'admin': {
+					color: this.options.adminColor.value,
+					hoverColor: this.options.adminColorHover.value
+				},
+				'firstComment': {
+					color: this.options.firstCommentColor.value,
+					hoverColor: this.options.firstCommentColorHover.value
+				},
 				'friend': {
 					color: this.options.friendColor.value,
 					hoverColor: this.options.friendColorHover.value
-				},
-				'submitter': {
-					color: this.options.OPColor.value,
-					hoverColor: this.options.OPColorHover.value
 				},
 				'moderator': {
 					color: this.options.modColor.value,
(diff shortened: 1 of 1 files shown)


lib/modules/userHighlight.js
```diff
@@ -186,12 +186,12 @@ modules['userHighlight'] = {
 					hoverColor: this.options.firstCommentColorHover.value
 				}
 			};
-			if (this.options.highlightOP.value) {
-				this.highlight('submitter');
-			}
 			if (this.options.highlightFriend.value) {
```
mc10: Shouldn't we be flipping this, doing mod, then submitter, then friend? Seems like the latest rule should trump all the rest.

lib/modules/userHighlight.js
```diff
@@ -161,14 +161,14 @@ modules['userHighlight'] = {
 		this.options['generateHoverColors'].callback = modules['userHighlight'].generateHoverColors;
 		if ((this.isEnabled()) && (this.isMatchURL())) {
 			this.colorTable = {
```
mc10: I think it's best to just alphabetize the elements here.

lib/modules/userHighlight.js
githue: In the screenshot the styles are ordered top-to-bottom in order of highest priority to lowest, it's from DevTools which re-orders selectors for convenience. If you were to look at the raw CSS it should look like:

```
.friend {}
.submitter {}
.moderator {}
.admin {}.
```

The if statements should be in that order too. Sorry for the confusion.

andytuba:
This'll be nice to clean up.   I personally would have sorted it the other way around: admin > mod > friend > op, in order of most exceptional to most regular. What's your thinking/rationale/source for the order here?

erikdesjardins:
@andytuba the title is in reverse order (i.e. ">" = "is superseded by"), so it's the same as your suggestion, except OP overrides Friend (a la reddit).
image

andytuba:
Oh, okay, thanks for the clue :)

githue:
Yeah I didn't put much thought into the title, Erik is right.

andytuba:
@andytuba TODO changeloge

PR #2387, opened 2015-07-30
Chrome Omnibox support in RES

joms:
This adds the support for using Chrome Omnibox, which allows the user to write *r*, press space and then enter the name of a subreddit to navigate to it. There is also an option for opening the subreddit in a new tab.

As I have no knowledge of other browsers, I have no idea if this functionality can exist in any other browsers than Chrome.


allthefoxes:
Is this really needed? Users interested could easily set this up in chrome search options yes?

andytuba:
Nifty! -- but I'd rather see this as a standalone extension or a guide to a custom search engine, particularly so users can enable/disable the functionality at will.

PR #2389, opened 2015-08-02
Shortcuts Easy Delete

githue:
When dragging and dropping shortcuts to the trash bin you have to drop them on a tiny 16x16 icon on the far corner of the page, why not make the icon really big and in the middle of the page instead?

Demo 1 (normal use): https://imgrush.com/CrrCGY95fNUf

Demo 2 (how users can discover it): https://imgrush.com/MTPhAfcOmfWY

It includes a little tip under the 'trash zone' about how to rearrange shortcuts.

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -23,7 +23,8 @@ body.res-console-open {
 	-o-transition: bottom 0.4s;
 	transition: bottom 0.4s;
 }
-#RESConsoleContainer em { font-style: italic; }
+body.res em { font-style: italic; }
+body.res strong { font-style: bold; }
 #RESConsoleContainer.slideIn {
 	visibility: visible;
 	bottom: 1.5%;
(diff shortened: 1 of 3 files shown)


andytuba:
Nifty!  I'd like an advanced option to disable the trash bin to sate any potential complainers if it's not too burdensome for dev time -- looks like you factored stuff so that's easy to add in, too.

githue:
Should the advanced option revert to the old version with the small icon in the subreddit bar, or should it disable drag+drop deletion entirely? I should've mentioned that the PR essentially just repositions the existing icon and makes it bigger, because I didn't think anyone would prefer the 16x16 icon over this.

What would you prefer?

andytuba:
I'm thinking advanced icon should just be to show or not show the drag'n'drop bin. I agree it's silly  to keep a tiny click target.

githue:
Tidied up the 'add shortcut' form since it will lead users to the trash bin.

shortcuts-easy-delete-add

PR #2391, opened 2015-08-03
Fix pre tags, and wiki UL tags in nightmode

link07:
This makes pre tags, and Wiki UL tags readable

Before, the UL tags had the same color as the background, and the pre tags had a white background with blue text on top, both of which made it hard to read them.


andytuba:
I appreciate your contribution, but these issues have already been fixed in master.  If you're interested in fixing other bugs (which would be great!), please clone RES to your computer and [load a developer build of RES](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/README.md#building-development-versions-of-the-extension).

PR #2395, opened 2015-08-04
Add support for image caption linebreaks

landaire:
In the showImage module image caption linebreaks are currently not supported, making lists difficult to read. This PR adds a new util function, $.safeHtmlLinebreaks, which first sanitizes the input string then replaces linebreaks with a `<br>` tag.

I only tested this with an imgur album ([this post in particular](https://www.reddit.com/r/interestingasfuck/comments/3fqlf1/famous_last_words/)) so if need be I can try to test other services as well.

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -1261,6 +1261,7 @@ h4.imgCaptions, .md h4.imgCaptions {
 }
 div.imgCaptions, .md div.imgCaptions {
 	font-size: 11px;
+	white-space: pre-wrap;
 }
 .imgCredits {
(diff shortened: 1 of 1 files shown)


erikdesjardins:
Could this be accomplished with CSS `.imgCaptions { white-space: pre-wrap; }` instead?

landaire:
It actually can be! I'll rewrite the patch to use that instead.

andytuba:
> squash commits

Classy, but optional :)

andytuba:
TODO: changelog

PR #2396, opened 2015-08-05
Settings console design cleanup

githue:
Minor layout issues have emerged in the settings console; tables difficult to read, long option names cut off by toggle buttons, elements out of alignment, inconsistent spacing between elements, green highlight color isn't colorblind friendly, and so on.

This tweaks some styles to make things more readable and clean, without changing the overall look too much. Some unused CSS was also cut out.

It also changes the layout model to [flex](http://www.w3.org/TR/css-flexbox-1/), which is ideal for the settings console, it only required the addition of one div which I've called 'optionSetting'. Using flex means you no longer have to worry about managing margins and fiddling with floats, flex takes care of it for you.

settings-console-design-labels

Screenshots showing before and after of the full settings console (updated):

Proposed look (latest):
settings-console-design-compare-after-04

Current look:
settings-console-design-compare-before

Suggestions are welcome and encouraged.

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -257,34 +257,28 @@ body.res-console-open {
 }
 .RESCloseButton {
 	position: absolute;
-	top: 7px;
-	right: 7px;
-	font: 12px Verdana, sans-serif;
+	top: 8px;
+	right: 8px;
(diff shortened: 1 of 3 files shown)


andytuba:
@andytuba todo changelog

PR #2397, opened 2015-08-06
Update XPI main file path in gulpfile.js

danieljl:
Since commit 24d772caa2fa056f73bbe01bbe8393a9e8b72b35 `gulp -b firefox` does not create main file because XPI/lib/main.js was moved to XPI/index.js but gulpfile.js was not updated.


andytuba:
Thanks!

mc10:
My bad, thanks for the catch!

PR #2400, opened 2015-08-06
Fixed oddshot.tv expando

nicememe:
- Fixed strange "onedrive" typo
- Updated AWS URL, old URL no longer supported.
- Updated RegEx to work better with Oddshot's new URL format.


andytuba:
Thanks!

PR #2404, opened 2015-08-11
XPI: Rename Firefox extension

mc10:
It'd be a lot nicer if our extension file wasn't some randomly-generated string of characters, but I'm not sure how the add-on upgrade process works. Thoughts @honestbleeps @andytuba?

### DIFF
--- a/XPI/package.json
+++ b/XPI/package.json
@@ -1,9 +1,10 @@
 {
-	"name": "reddit_res",
+	"name": "reddit-enhancement-suite",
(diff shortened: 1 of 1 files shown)


andytuba:
I think the ID has to be the same for upgrades. If hb doesn't know, we should go ask in irc.mozilla.org#addons.

callahad:
You need to keep the `jid1-xUfzOsOFlzSOXg@jetpack` id for upgrades to work. Sorry :(

PR #2407, opened 2015-08-12
jscs: Enforce tab indentation

mc10:
Did not realize this wasn't part of the build. This'll be nice to enforce.


erikdesjardins:
We should probably wait to merge this until after #2401 because resolving the conflicts will be considerably easier on this end.

andytuba:
sorry, gonna be lazy for a ittle bit and focus on functional stuff -- let's merge this right after 4.6.1

mc10:
That's fair.

PR #2412, opened 2015-08-16
Enhance Notifications UX

githue:
**Changes to appearance**

* Never cover up the user bar or content (so there's no pressing need to close a notification)
* Message text color changed to black (so it's easier to read)
* Blue button moved over to the left (to make it associated with the preceding message)
* Cog moved to the left of the title (more balanced, and won't be shifted around from long titles)
* Uses the titlebar to add strong contrast (so it stands out in various subreddit themes)

Before / after:
notifications-compare

**New behaviour for non-stickied notifications**

The fade out animation was made much slower (several seconds) and can be interrupted with the cursor. This way the user can see when a non-sticky notification is about to disappear and decide whether they want it to stay up for longer.

Video [16s]: https://imgrush.com/vCEpII7etPnk

**Option descriptions rewritten, and a new option**

The options descriptions for 'closeDelay' and 'sticky' were rewritten to be more descriptive. There's also a new option for setting the fadeOutLength, which defaults to five seconds.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -111,6 +111,7 @@ changelog v4.6.0
 	- Cleaned up gulp (thanks @mc10)
 	- Refactored gulp a bit (thanks @andytuba)
 	- Refactored gulp a *lot* (thanks @erikdesjardins)
+	- Bugfixes to gulp scripts (thanks @danieljl)
 	- Update and maintain readme (thanks @allthefoxes and @mc10)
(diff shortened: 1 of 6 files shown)


andytuba:
tacked on a few tweaks, whatcha thing?
 https://github.com/githue/Reddit-Enhancement-Suite/compare/notificationCompact...andytuba:2412
screen shot 2015-08-15 at 8 24 28 pm

mc10:
Snazzy!

PR #2417, opened 2015-08-16
filteReddit: Avoid unnecessary [].slice wrapping

mc10:
`RESUtils.things` should already return an array, so this is unnecessary.


andytuba:
lgtm

Issue #1125, opened 2014-05-25
Educate users about hardIgnore


andytuba:
Show a notification when the user ignores a redditor (either via user info popup's [Ignore] button or by setting ignored toggle on user tagger form) educating the user about "ignored" behavior and how to set hardToggle.  link to hardToggle option, show at once most a day?

PR #2418, opened 2015-08-16
UserTagger: fix a number of layout/styling issues

### LINKED ISSUE #2339: User Tag dialog overflows
Whenever it's long enough that the preview wraps.
image

githue:
Fixes #2339

[screenshot below]

Notice the color change for the toggleButton, I think it's better this way because users are used to red meaning off.

### DIFF
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -137,24 +137,27 @@ addModule('userTagger', function(module, moduleID) {
 			}
 			// add tooltip to document body...
 			userTaggerToolTip = RESUtils.createElement('div', 'userTaggerToolTip', 'RESDialogSmall');
-			var thisHTML = '<h3>Tag User</h3><div id="userTaggerToolTipContents" class="RESDialogContents clear">';
+			var thisHTML = '<h3>Tag User</h3><div id="userTaggerToolTipContents" class="RESDialogContents">';
 			thisHTML += '<form name="userTaggerForm" action=""><input type="hidden" id="userTaggerName" value="">';
-			thisHTML += '<label for="userTaggerTag">Tag</label> <input type="text" id="userTaggerTag" value="">';
-			thisHTML += '<div id="userTaggerClose" class="RESCloseButton">&times;</div>';
-			thisHTML += '<label for="userTaggerColor">Color</label> <select id="userTaggerColor">';
+			thisHTML += '<div><label for="userTaggerTag">Tag</label> <input type="text" id="userTaggerTag" value=""></div>';
+			thisHTML += '<div><label for="userTaggerColor">Color</label> <select id="userTaggerColor">';
 			for (var color in bgToTextColorMap) {
 				var bgColor = (color === 'none') ? 'transparent' : color;
 				thisHTML += '<option style="background-color: ' + bgColor + '; color: ' + bgToTextColorMap[color] + ' !important;" value="' + color + '">' + color + '</option>';
 			}
-			thisHTML += '</select>';
-			thisHTML += '<label for="userTaggerPreview">Preview</label> <span id="userTaggerPreview"></span>';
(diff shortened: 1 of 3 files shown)


githue:
Update:
usertagger-layout-02

githue:
Travis doesn't like scripts inside hrefs I guess, suggestions?

erikdesjardins:
You could remove the `href` attribute entirely at that point, which has the added benefit of making the link unclickable.

Maybe also:

```css
a:not([href]) {
    color: inherit;
    text-decoration: none;
}
```

githue:
Thank you.

githue:
I have an idea for replacing the color select menu with a chart (using radios hidden inside labels). So users can have more colors to choose from.

https://imgrush.com/O5NzgfQy3GJL

Thoughts?

mc10:
Ehh that's such a cludgy way of matching the selector. Is there a way to avoid the `:not([href])` portion?

githue:
I've decided to move the color selection option to a separate branch because it's not really related to this PR.

Meanwhile, I've added a fix for the user tagger dialog opening outside the page when opened from the side bar. See the moderator list for an example.

githue:
I think this will help with #1125 because it makes it more obvious that the option can be customized.

githue:
Whoops! I forgot to sync master properly before doing the merge and push, so it added unintended commits. How should I resolve this?

Ajedi32:
@githue You could always rebase: `git rebase master` (Then `git log --oneline --graph` to make sure everything looks as expected, and `git push -f` to update the PR.)

lib/modules/userTagger.js
```diff
@@ -137,24 +137,27 @@ addModule('userTagger', function(module, moduleID) {
 			}
 			// add tooltip to document body...
 			userTaggerToolTip = RESUtils.createElement('div', 'userTaggerToolTip', 'RESDialogSmall');
-			var thisHTML = '<h3>Tag User</h3><div id="userTaggerToolTipContents" class="RESDialogContents clear">';
+			var thisHTML = '<h3>Tag User</h3><div id="userTaggerToolTipContents" class="RESDialogContents">';
```
andytuba: I'd love to see this moved into templates.html while you're in here, too.

githue: Done.

lib/modules/userTagger.js
```diff
@@ -225,6 +165,54 @@ addModule('userTagger', function(module, moduleID) {
...
+
+		// Define $tagger.
+		$tagger = $('#userTaggerToolTip');
+		// Add generated elements.
+		$tagger.find('.res-usertag-ignore').append(ignoreToggleButton).append(ignoreSettingsLink);
+		// Populate color selection with options.
+		$colorSelect = $tagger.find('#userTaggerColor');
+		for (var color in bgToTextColorMap) {
+			var bgColor = (color === 'none') ? 'transparent' : color;
+			$colorSelect.append('<option style="background-color: ' + bgColor + '; color: ' + bgToTextColorMap[color] + ' !important;" value="' + color + '">' + color + '</option>');
```
erikdesjardins: You should be able to use [a list](https://mustache.github.io/mustache.5.html#Sections) in the template to accomplish this, i.e. something like (haven't actually tested it):
```html
<select id="userTaggerColor">
  {{#colors}}
    <option style="background-color: {{bgColor}}; color: {{textColor}} !important;" value="{{color}}">{{color}}</option>
  {{/colors}}
</select>
```
and then when you call `template.html()`, pass in an object of this shape:
```js
template.html({
  colors: [
    { bgColor, textColor, color }
  ]
})
```

andytuba: No ES6 in this branch :sob:

```javascript
template.html({
  bgColor: bgColor,
  textColor: textColor,
  color: color
});
```

PR #2419, opened 2015-08-17
Notifications: cross browser compatibility for animation

githue:
As per: http://caniuse.com/#feat=css-animation and https://developer.mozilla.org/en-US/docs/Web/CSS/animation#Browser_compatibility

Mainly for Android browser and Safari.

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -840,10 +840,11 @@ body.res-console-open {
 	-webkit-animation: notificationFadeOut, 3s, ease-in, 3s, 1, normal, forwards, running;
 	animation: notificationFadeOut, 3s, ease-in, 3s, 1, normal, forwards, running;
 }
+@-webkit-keyframes notificationFadeOut {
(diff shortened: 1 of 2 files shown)


lib/core/res.css
```diff
@@ -837,11 +837,8 @@ body.res-console-open {
...
 .RESNotification.timerOn {
-	animation-duration: 3s; /* default */
-	animation-delay: 5s; /* default */
-	animation-timing-function: ease-in;
-	animation-fill-mode: forwards;
-	animation-name: notificationFadeOut;
+	-webkit-animation: notificationFadeOut, 3s, ease-in, 3s, 1, normal, forwards, running;
+	animation: notificationFadeOut, 3s, ease-in, 3s, 1, normal, forwards, running;
 }
 @keyframes notificationFadeOut {
```
erikdesjardins: You're probably gonna need `@-webkit-keyframes` too.

PR #2420, opened 2015-08-17
Make nightmode look a little nicer

githue:
* Adjusted some colours to be less harsh.
* **Removed custom expando sprites** and replaced with the simple CSS 'filter'. Filter is a relatively new and experimental style, but I feel that it's worth using because it saves an extra http request, and if it fails for some users they only have to deal with bright icons. [caniuse chart](http://caniuse.com/#feat=css-filters).
* Changed a bit of inline CSS in upload.js.

Animated gif showing differences (new version has darker colors on text):
nightmode-pretty-compare

### DIFF
--- a/lib/modules/nightmode.css
+++ b/lib/modules/nightmode.css
@@ -57,7 +57,6 @@
 .res-nightmode.liveupdate-app .content {
 	background-color: rgb(22,22,22);
 }
-.res-nightmode .entry .buttons li a,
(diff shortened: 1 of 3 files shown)


erikdesjardins:
Very nice!

I believe the only browser we "support" that doesn't support `filter` is Opera 12.1, but I don't think that's a big enough deal to matter.

IMO we really need to drop Opera 12 support (after 4.6.1?), it'll only hold us back.

PR #2422, opened 2015-08-17
Usability improvements for notifications settings

githue:


lib/modules/notifications.js
```diff
@@ -58,7 +59,7 @@ addModule('notifications', function(module, moduleID) {
 					value: false
 				},
 				{
-					name: 'cooldown',
+					name: 'closeDelay',
```
andytuba: notificationType rows get mapped to objects using the field names, so instances of `.cooldown` would need to get updated if this were changed.

that said, `cooldown` is how many milliseconds before this notification type is allowed to be shown again

lib/modules/notifications.js
```diff
@@ -37,6 +37,7 @@ addModule('notifications', function(module, moduleID) {
 		notificationTypes: {
 			description: 'Manage different types of notifications',
 			type: 'table',
+			advanced: true,
```
andytuba: good idea

lib/modules/notifications.js
```diff
@@ -26,7 +26,7 @@ addModule('notifications', function(module, moduleID) {
 		closeDelay: {
 			type: 'text',
 			value: 3000,
-			description: 'In milliseconds, length of time until a notification begins to disappear. Can be customized per notification type.'
+			description: 'In milliseconds, length of time until a notification begins to disappear. Can be customized per notification type in advanced options.'
```
andytuba: not relevant because cooldown isn't actually a closeDelay

PR #2424, opened 2015-08-18
visitedStyle changes

githue:
Currently visitedStyle acts a little strangely; it doesn't affect links in self text or the side bar, it overrides links in .tagline, and even when disabled styleTweaks.js applies a blanket visited style to all comments anyway (to help with expandable images).

I found that showImages.js adds a name prefixed with `img` to all expandable image links, so we can target them with the selector `a[name^="img"]`, no need to style every normal link. The only shortfall is that when visitedStyle is disabled, duplicate images don't have a name so they don't change color.

I'm not sure if it's a good idea to apply this to the side bar because I found a number of themes that aren't very strict with their side bar link colors. Fortunately I haven't seen any serious issues, and themers don't have to do much to avoid it.

Here's a screenshot of my proposal:
visited-style-01-proposal

Screenshot from master for comparison:
visited-style-01-master

### DIFF
--- a/lib/modules/nightmode.css
+++ b/lib/modules/nightmode.css
@@ -822,22 +822,17 @@ body.res-nightmode,
 	border-right-color: rgb(118,133,148);
 }
 /* moderator style takes higher priority than friend */
-.res-nightmode .tagline .friend,
-.res-nightmode .tagline .friend:visited {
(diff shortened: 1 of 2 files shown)


githue:
Alternative: take out the option altogether and have it apply to self text and comments only.

Issue #2096, opened 2015-03-11
link to subreddit when listed as "xpost" in title


andytuba:
https://www.reddit.com/r/Enhancement/comments/2ynzh1/feature_request_would_it_be_possible_to_have_res/ and a dozen other requests

potential implementations: 

* post info/actions hover pop-up when hovering the title, shown below the title
* wrap subreddit name in a span, onclick open subreddit (yuck, re implementing links in JavaScript) 
* split post title link into multiple links, but this sounds dangerous

erikdesjardins:
Related: #1229

andytuba:
> * wrap subreddit name in a span, onclick open subreddit (yuck, re implementing links in JavaScript)
> * split post title link into multiple links, but this sounds dangerous

compromise idea: 

1. wrap subreddit name in a span
1. onclick span change title link to that url
1. onclick link revert title link to original url

erikdesjardins:
There's also #2425. It's a different approach, but I like it, I think it would be less surprising.

andytuba:
Oh right! Yeah, I think I like that best.

erikdesjardins:
#2425

PR #2425, opened 2015-08-20
xPostLinks module

br4nnigan:
compares the post titles with a regular expression (looking for strings similar to "x-posted from r/gaming") and creates links to the x-posted subs in the tagline, like this (also removed the comma after creating this screenshot)

this

did so cause i always hated copy pasting the channel with the usual closing bracket after the sub and also copying link texts is a pain itself.

hope u like!

### DIFF
--- a/lib/modules/xPostLinks.js
+++ b/lib/modules/xPostLinks.js
@@ -52,5 +52,5 @@ addModule('xPostLinks', function(module, moduleID) {
 				}
 			}
 		}
-	};
+	}
 });
--- a/lib/modules/xPostLinks.js
+++ b/lib/modules/xPostLinks.js
@@ -2,9 +2,7 @@ addModule('xPostLinks', function(module, moduleID) {
 	module.moduleName = 'X-post Links';
 	module.category = 'Posts';
 	module.description = 'Create links to x-posted subreddits in the taglines.';
-	module.include = [
-		/^https?:\/\/([a-z]+)\.reddit\.com/i
-	];
+	module.include = [ 'linklisting', 'comments' ];
 
 	module.go = function() {
 		if ((module.isEnabled()) && (module.isMatchURL())) {
--- a/lib/modules/xPostLinks.js
+++ b/lib/modules/xPostLinks.js
@@ -10,21 +10,29 @@ addModule('xPostLinks', function(module, moduleID) {
 		if ((module.isEnabled()) && (module.isMatchURL())) {
 
 			createLinks();
-			window.addEventListener('neverEndingLoad', createLinks);
+			window.addEventListener('neverEndingLoad', onNeverEndingLoad);
 		}
 	};
 
-	function createLinks() {
+	function onNeverEndingLoad ( evt ) {
 		var
-		titles        = document.querySelectorAll('a.title'),
+		siteTables = document.body.querySelectorAll('.sitetable'),
+		lastItem = siteTables.item(siteTables.length - 1);
+
+		createLinks( lastItem );
+	}
+
+	function createLinks( container ) {
+		var
+		titles        = (container || document.body).querySelectorAll('a.title'),
 		xpost_re      = /x-?post/i,
 		xpost_from_re = /(x-?post\w*\b)(\s+\S+)?\s+(\/?r\/)?([a-zA-Z0-9]+)/i;
 
 		for (var i = 0, title, tagline, entry, userattrs, textNode, subreddit, add_after, xpost_from, xpost_from_a, sub; i < titles.length; i++) {
 
-			title = titles[i];
+			title = titles.item(i);
 
-			if ( xpost_re.test(title.innerHTML) && !title.classList.contains('x-post') ){
+			if ( xpost_re.test(title.innerHTML) ){
 
 				xpost_from = title.innerHTML.match(xpost_from_re);
 
@@ -57,7 +65,6 @@ addModule('xPostLinks', function(module, moduleID) {
 								add_after.parentNode.appendChild(xpost_from_a);
 							}
 						}
-						title.classList.add('x-post');
 					}
 				}
 			}
--- a/lib/modules/xPostLinks.js
+++ b/lib/modules/xPostLinks.js
@@ -1,0 +1,66 @@
+addModule('xPostLinks', function(module, moduleID) {
+	module.moduleName = 'X-post Links';
+	module.category = 'Posts';
+	module.description = 'Create links to x-posted subreddits in the taglines.';
+	module.include = [
+		/^https?:\/\/([a-z]+)\.reddit\.com/i
+	];
+
+	module.go = function() {
+		if ((module.isEnabled()) && (module.isMatchURL())) {
+
+			module.createLinks();
+			window.addEventListener('neverEndingLoad', module.createLinks);
+		}
+	};
+
+	module.createLinks = function () {
+		var
+		titles        = document.querySelectorAll('a.title'),
+		xpost_re      = /x-?post/i,
+		xpost_from_re = /(x-?post\w*\b)(\s{0,1}\S+)?\s([r/]+)([a-zA-Z0-9]+)/i;
+
+		for (var i = 0, title, tagline, entry, userattrs, textNode, subreddit, add_after, xpost_from, xpost_from_a, sub; i < titles.length; i++) {
+
+			title = titles[i];
+
+			if ( xpost_re.test(title.innerHTML) && !title.classList.contains('x-post') ){
+
+				xpost_from = title.innerHTML.match(xpost_from_re);
+
+				if ( xpost_from && xpost_from.length >= 5 ){
+
+					sub = xpost_from[4];
+					entry = title.parentNode.parentNode;
+
+					if ( entry && entry.classList.contains('entry') ){
+
+						textNode = document.createTextNode(' x-posted from ');
+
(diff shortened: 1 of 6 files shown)


allthefoxes:
That's pretty cool. :+1:

andytuba:
Neat-o -- thanks for contributing this! People keep asking for it, glad to finally deliver.

lib/modules/xPostLinks.js
```diff
@@ -0,0 +1,66 @@
+addModule('xPostLinks', function(module, moduleID) {
+	module.moduleName = 'X-post Links';
+	module.category = 'Posts';
+	module.description = 'Create links to x-posted subreddits in the taglines.';
+	module.include = [
+		/^https?:\/\/([a-z]+)\.reddit\.com/i
```
andytuba: This can just be `[ 'linklisting', 'comments' ]` -- since it only needs to run on pages where you see posts.

lib/modules/xPostLinks.js
```diff
@@ -0,0 +1,66 @@
...
+					entry = title.parentNode.parentNode;
+
+					if ( entry && entry.classList.contains('entry') ){
+
+						textNode = document.createTextNode(' x-posted from ');
+
+						xpost_from_a           = document.createElement('a');
+						xpost_from_a.href      = 'https://www.reddit.com/r/'+sub;
+						xpost_from_a.className = 'subreddit hover';
+						xpost_from_a.innerHTML = '/r/'+sub;
```
andytuba: `.textContent` for safety

lib/modules/xPostLinks.js
```diff
@@ -0,0 +1,66 @@
...
+
+						userattrs = entry.querySelector('.userattrs');
+						subreddit = entry.querySelector('.subreddit');
+						tagline   = entry.querySelector('.tagline');
+
+						add_after = subreddit || userattrs || tagline && tagline.lastChild;
+
+						if ( add_after ){
+							if ( add_after.nextSibling ){
```
andytuba: We have an `RESUtils.insertAfter` utility method for this sort of thing, but no worries if you want to keep the same codebase between your script and RES's.

lib/modules/xPostLinks.js
```diff
@@ -0,0 +1,66 @@
...
+		var
+		titles        = document.querySelectorAll('a.title'),
+		xpost_re      = /x-?post/i,
+		xpost_from_re = /(x-?post\w*\b)(\s+\S+)?\s+(\/?r\/)?([a-zA-Z0-9]+)/i;
+
+		for (var i = 0, title, tagline, entry, userattrs, textNode, subreddit, add_after, xpost_from, xpost_from_a, sub; i < titles.length; i++) {
+
+			title = titles[i];
+
+			if ( xpost_re.test(title.innerHTML) && !title.classList.contains('x-post') ){
```
andytuba: Unless we have a good reason to style xposts, I think it would be better to store "this post has been processed already" as a data attribute ... or not at all.

lib/modules/xPostLinks.js
```diff
@@ -0,0 +1,66 @@
...
+		if ((module.isEnabled()) && (module.isMatchURL())) {
+
+			module.createLinks();
+			window.addEventListener('neverEndingLoad', module.createLinks);
+		}
+	};
+
+	module.createLinks = function () {
+		var
+		titles        = document.querySelectorAll('a.title'),
```
andytuba: Could you implement the `= function(container) { var titles = (container || document.body).querySelectorAll(...)` pattern? helps avoid re-scanning elements that have already been processed.

lib/modules/xPostLinks.js
```diff
@@ -0,0 +1,66 @@
...
+
+	module.go = function() {
+		if ((module.isEnabled()) && (module.isMatchURL())) {
+
+			module.createLinks();
+			window.addEventListener('neverEndingLoad', module.createLinks);
+		}
+	};
+
+	module.createLinks = function () {
```
erikdesjardins: Very minor: `createLinks` doesn't have to be a module property, it can just be `function createLinks() {...}`

lib/modules/xPostLinks.js
honestbleeps: also note innerHTML use here. I know it's a get and not a set, but might it make more sense to test against textContent anyway?

lib/modules/xPostLinks.js
erikdesjardins: And if you do decide to diverge, we have a better way of watching for new elements than an event listener (using `RESUtils.watchForElement`), though that might be more work than it's worth.

erikdesjardins:
Could you also add this to Opera12 and Safari's manifests (paths [here](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/hosts/example.js#L10-L14)), and take a look at the build errors?

br4nnigan:
Thank you guys for all the feedback! Will try to incorporate everything.

lib/modules/xPostLinks.js
br4nnigan: watchForElement seems super useful especially

lib/modules/xPostLinks.js
br4nnigan: That's cool but it also needs to include the /user/ pages. What's the keyword for that, or better how could i possibly find out myself?

br4nnigan:
Allright it's all done. Do i have to do anything now? I'm sorry, i've never done this sort of collaboration before :) I'm enjoying it a lot :) Thank you guys/girls.

lib/modules/xPostLinks.js
```diff
@@ -0,0 +1,61 @@
...
+
+				sub        = title.textContent.match(subreddit_re);
+				xpost_from = title.textContent.match(xpost_from_re);
+
+				sub        = sub && sub.length >= 1 && sub[1];
+				xpost_from = xpost_from && xpost_from.length >= 5 && xpost_from[4];
+
+				if ( sub || xpost_from ){
+
+					sub = sub || xpost_from;
```
erikdesjardins: I think `xpost_from` should be prioritized, to prevent this from beating it: `Something something /r/notxpost [x-posted from /r/actualxpost]`.

lib/modules/xPostLinks.js
```diff
@@ -0,0 +1,61 @@
...
+					entry = title.parentNode.parentNode;
+
+					if ( entry && entry.classList.contains('entry') ){
+
+						textNode = document.createTextNode(' x-posted from ');
+
+						xpost_from_a           = document.createElement('a');
+						xpost_from_a.href      = 'https://www.reddit.com/r/'+sub;
```
erikdesjardins: `href` can be just `'/r/' + sub`, to avoid redirecting people from other subdomains to `www.`.

erikdesjardins:
No, thank *you* :)

There's nothing else you need to do (aside from other code tweaks), it's all in our hands.

At some point this will get merged - possibly not for a while - and eventually it'll be part of a release.

lib/modules/xPostLinks.js
br4nnigan: I've been aware of the problem you mention, however the other way creates other problem cases. I've put some more thought into it and came to this conclusion:
I think we can agree that if after "x-post" something with the pattern of r/string is found, we can be pretty sure we found gold. That's priority 1.
Suggested priority 2: if we find one or two words after the "x-post" part we can be pretty sure it's either something like " from askReddit" or just the subreddit. 
Problemematic are cases with more than 2 words after the x-post. We could guess the subreddit is probably the last word, but i've already encountered a case where this would fail. I'd say we just skip those. Should be pretty rare.
I made some tests which showed prio 1 would be 80-90% of all cases, prio 2 the rest. In about 200 x-post finds none failed (only tested the regex programatically, no manual evaluation)

try yourself
https://www.reddit.com/search?q=x-post+subreddit%3Abooks&restrict_sr=&sort=top&t=all

    var el = document.querySelectorAll(".search-title");
    var xpost_re      = /(x-?post\S*)(.+)/i;
    var xpost_from_re = /^(\s+\S+)?\s+\/?([a-zA-Z0-9_]+)([\)\]}])|($)/i;
    var subreddit_re  = /r\/([a-zA-Z0-9_]+)/i; 
    var s;
    var success = {
        "xpost_re": 0,
        "xpost_from_re": 0,
        "subreddit_re": 0
    };
    for (var i = 0; i < el.length; i++) {
        
        if ( el[i].textContent.match(xpost_re) ){
            success["xpost_re"]++;
            s = el[i].textContent.match(xpost_re)[2];
            if ( s.match(subreddit_re) ){
                success["subreddit_re"]++;
            } 
            else if ( s.match(xpost_from_re) ){
                success["xpost_from_re"]++;
            }
        } 
    }
    console.log(success);

will commit later

lib/modules/xPostLinks.js
br4nnigan: I know i had some reason not to do that, however i can't recall it... i'll change it anyway.

lib/modules/xPostLinks.js
erikdesjardins: Yeah, that sounds logical. :+1:

erikdesjardins:
Fixes #2096

erikdesjardins:
Merged in #2861.

Thanks @br4nnigan!

image

br4nnigan:
Awesome! Very welcome.

PR #2426, opened 2015-08-21
Only use legacy gearIcon in user bar

githue:
I believe this was the intended purpose of this function, because it's only checking the background property of the icon in the user bar. The legacy class shouldn't affect every icon on the page since they could be styled differently.

See /r/naut and /r/askreddit using master and notice the broken icons in userTagger, commentPreview and NER. I'm not sure which other subreddits to check.


erikdesjardins:
Yeah, that was the intent, and I believe our original concern was subs using Naut so :+1:

PR #2428, opened 2015-08-21
Properly align the pause and up buttons in #NREFloat

Goctionni:


andytuba:
pixel pushing is best pushing

PR #2429, opened 2015-08-22
Update to NER notifications

githue:
* Rewrote the message of the 'pauseAfterEvery' notification.
* Removed unnecessary cogs from messages, users can use the cog in the title.
* Don't show 'paused' notification when 'pauseAfterEvery' notification is shown, because it's unnecessary and confusing to show them both at once.
* Removed notifyWhenPaused option, because users can use the checkbox in the notification instead.

### DIFF
--- a/lib/modules/neverEndingReddit.js
+++ b/lib/modules/neverEndingReddit.js
@@ -15,12 +15,6 @@ modules['neverEndingReddit'] = {
 			value: true,
 			description: 'Automatically load new page on scroll (if off, you click to load)'
 		},
-		notifyWhenPaused: {
-			type: 'boolean',
-			value: true,
-			description: 'Show a reminder to unpause Never-Ending Reddit after pausing',
-			advanced: true
-		},
 		reversePauseIcon: {
 			type: 'boolean',
 			value: false,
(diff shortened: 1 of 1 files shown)


lib/modules/neverEndingReddit.js
```diff
@@ -271,14 +262,15 @@ modules['neverEndingReddit'] = {
...
-			var notification = [];
-			notification.push('Time for a break!');
-			notification.push('Never-Ending Reddit was paused automatically. ' + modules['settingsNavigation'].makeUrlHashLink('neverEndingReddit', 'pauseAfterEvery', '', 'gearIcon'));
-			notification = notification.join('<br><br>');
-			setTimeout(modules['notifications'].showNotification.bind(RESUtils, notification, 5000));
+			this.togglePause('pauseAfterEvery');
+			var message = '<p>Time for a break!</p>';
+			message += '<p>Never-Ending Reddit has been paused because you\'ve passed ' + this.pauseAfterPages + ' pages. You can change this number in RES settings.</p>';
+			modules['notifications'].showNotification({
+				moduleID: modules['neverEndingReddit'].moduleID,
```
erikdesjardins: You can add in `optionKey: 'pauseAfterEvery'` to make the gear link to that option.

PR #2434, opened 2015-08-27
Correct positioning for dropdown lists

githue:
- Stops dropdown lists from going outside of the viewport.
- Corrects height of toggleButtons in dropdowns.
- Corrects cursor for top-right gearIcon in Firefox.

Before / after:
dropdownlist-position

### DIFF
--- a/lib/modules/hover.js
+++ b/lib/modules/hover.js
@@ -413,22 +413,16 @@ addModule('hover', function(module, moduleID) {
 	HoverDropdownList.prototype = new Hover();
 	HoverDropdownList.prototype.templateName = 'RESHoverDropdownList';
 	HoverDropdownList.prototype._updatePosition = function($container) {
-		var XY = RESUtils.getXYpos(this._target);
-
-		var height = $(this._target).height();
-
-		var containerWidth = $container.outerWidth();
-
-		var left = XY.x;
-		if (left + containerWidth > window.innerWidth) {
-			left = window.innerWidth - containerWidth - 10;
-		}
-
-		$container.css({
-			top: XY.y + (height * 1.1) + 10,
-			left: left,
-			position: 'fixed',
-			zIndex: 2147483646
-		});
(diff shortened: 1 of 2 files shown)


lib/modules/hover.js
```diff
@@ -413,22 +413,16 @@ addModule('hover', function(module, moduleID) {
...
-		}
-
-		$container.css({
-			top: XY.y + (height * 1.1) + 10,
-			left: left,
-			position: 'fixed',
-			zIndex: 2147483646
-		});
+		var trigger = $(this._target)[0];
+		var menu = $container[0];
```
erikdesjardins: Do we need to avoid jQuery here? The following appears to work the same:

```javascript
var pos = RESUtils.getXYpos(this._target);
var t = pos.y + this._target.offsetHeight,
    r = document.body.scrollWidth - pos.x - this._target.offsetWidth,
    l = pos.x;
$container.css({
	right: l + $container.outerWidth() < document.body.scrollWidth ? 'auto' : r,
	left: l + $container.outerWidth() < document.body.scrollWidth ? pos.x : 'auto',
	zIndex: 2147483646,
	position: 'fixed',
	top: t + 2
});
```

githue: No need to avoid jQuery, I originally changed it to simplify it for myself, but we can def use jQuery there. I'll make a new PR with the cleaner code, after retesting.

erikdesjardins: Eh, I didn't refresh to see that it was already merged, you don't really need to do it at this point.

githue: Alright, I'll keep it in mind for other things though.

andytuba: yeah, I figured it didn't matter all that much functionally when I merged, just code style.

erikdesjardins: Yeah I'm probably too nitpicky about that, anyway.

PR #2438, opened 2015-08-29
Fixed username sorting of tagged users in dashboard

githue:
* sorting tagged users by username now works even after re-sorting other columns.
* added 'Page' before pagination buttons, and fixed their positioning.
* tableTools no longer affects dashboard. Was requiring an extra click to sort by username.

### DIFF
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -86,7 +86,7 @@ modules['userTagger'] = {
 			css += '#userTaggerToolTip .toggleButton.enabled .toggleOn { background-color: #ddd ; color: #636363; }';
 			css += '#userTaggerToolTip .toggleButton.enabled .toggleOff { background-color: #d02020; color: #fff; }';
 			css += '#userTaggerToolTip .toggleButton .toggleOff { background-color: #ddd; color: #636363; } ';
-			css += '#userTaggerTable th { -moz-user-select: none; -webkit-user-select: none; -o-user-select: none; user-select: none; }';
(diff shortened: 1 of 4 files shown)


lib/modules/userTagger.js
```diff
@@ -506,14 +507,11 @@ modules['userTagger'] = {
...
+				break;
 			default:
-				// sort users, ignoring case
-				taggedUsers.sort(function(a, b) {
-					return (a.toLowerCase() > b.toLowerCase()) ? 1 : (b.toLowerCase() > a.toLowerCase()) ? -1 : 0;
-				});
-				if (this.descending) {
-					taggedUsers.reverse();
-				}
+				modules['userTagger'].sortUsername(taggedUsers, this.decending);
```
andytuba: typo -- `descending`

lib/modules/userTagger.js
```diff
@@ -506,14 +507,11 @@ modules['userTagger'] = {
 					taggedUsers.reverse();
 				}
 				break;
+			case 'username':
+				modules['userTagger'].sortUsername(taggedUsers, this.descending);
```
andytuba: may as well  simply `case 'username': // falls through` and delete the rest of this case.

githue: JSHint gives a warning when I do that.

> `Expected a 'break' statement before 'default'.`

andytuba: Oh, sorry, JSHint is expecting this:

    case 'username':
        /* falls through */
    default:

PR #2440, opened 2015-08-29
Support for Miiverse posts.

sfoop:
This makes posts from miiverse.nintendo.net show up like posts from Twitter do. Nintendo provides embed HTML code when viewing a post on Miiverse which is a div tag and a script tag. The modules/hosts/miiverse.js just generates the same div & script tags, and places it in a GENERIC_EXPANDO.

### DIFF
--- a/Chrome/manifest.json
+++ b/Chrome/manifest.json
@@ -144,6 +144,7 @@
 				"modules/hosts/github.js",
 				"modules/hosts/onedrive.js",
 				"modules/hosts/oddshot.js",
+				"modules/hosts/miiverse.js",
 				"core/init.js"
 			],
 			"css": [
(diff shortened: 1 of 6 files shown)


andytuba:
Unfortunately, security concerns from Mozilla prevent RES from simply fetching Nintendo's embed script. RES's Twitter expando code includes special handling for Firefox to embed the contents without the nice styling: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/hosts/twitter.js#L76-L78

Is it possible to fetch just the content without using the embed script?

sfoop:
The quick answer is you just need to put an iframe tag inside the div, where the src URL is the Miiverse post (which is a data- attrib on the div) with /embed added to path. I'd have to look into trying that later.

andytuba:
Iframe solution sounds good!

sfoop:
I updated the code to generate an IFRAME for Firefox. I didn't actually test it in Firefox though (I've never done extension development there), but I forced the code to run in Safari and it worked as expected.

erikdesjardins:
Released in 4.6.1 62d604e22097fb1f8d3db8cca42a48973252d650

Issue #2219, opened 2015-05-06
RES interacting strangely with /r/keys


allthefoxes:
https://www.reddit.com/r/RESissues/comments/350nh4/bug_post_titles_from_rkeys_seem_to_have_some_text/

https://www.reddit.com/r/keys+randomprecision1

Also, /r/keys does not exist according to subredditHover

jsayol:
This happens when the subreddit name coincides with a function in ```Array.prototype```. I did some testing and you can see the same behaviour with subreddits like [map](https://www.reddit.com/r/map+randomprecision1) or [sort](https://www.reddit.com/r/sort+randomprecision1).

The fix is easy: instead of using object[element] [here](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/389d54cb0cc8d08db7668e0eb6604e049c920794/lib/modules/subredditTagger.js#L87-L88) to check if an element exists, just use object.hasOwnProperty(element).

I'm creating the pull request now.

PR #2442, opened 2015-08-30
Fix for issue #2219 on GitHub

jsayol:
Fix for issue #2219. It also fixes the subreddit hover, which was closely related.


andytuba:
looks legit.  Could somebody test this to verify, or I'll do it in the next few days?

andytuba:
Thanks!

PR #2443, opened 2015-08-30
Highlighted NSFW expandos in nightmode

githue:
In nightmode the CSS filter property on expandos doesn't play well with the NSFW stripes, because it converts them to grayscale:

nsfw-expando-night-gray

I couldn't find a solution to make the stripes work, short of adding the old sprite back in.

As a workaround I added some extra filters and removed the stripes, turning the expando itself red:
nsfw-expandos-night

The [RES expandos](https://s3.amazonaws.com/b.thumbs.redditmedia.com/tBwwK20XXxtpgudWx1L7bDXla-iotv-JA0jgA0Y-FVs.png) have lighter borders and +/- indicators than the text and video expandos, so they're harder to see. It's still usable, but could be better.

### DIFF
--- a/lib/modules/nightmode.css
+++ b/lib/modules/nightmode.css
@@ -341,6 +341,14 @@
 	-webkit-filter: grayscale(100%) invert(80%);
 	filter: grayscale(100%) invert(80%);
 }
+/* from white & blue to black and red-ish */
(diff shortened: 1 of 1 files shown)


githue:
Participants from #2146 may be interested.

PR #2444, opened 2015-09-02
Added an extra class for selected comments

githue:
As requested [here](https://www.reddit.com/r/Enhancement/comments/2odnoi/cssis_there_a_reason_the_active/) and [here](https://www.reddit.com/r/Enhancement/comments/3j3jnu/feature_request_put_reskeynavactiveelement_on_the/).

This doesn't change any existing behavior or styling, it just adds a class to `.thing` and leaves it up to themers to do what they will.

I made a demo page with some styling to show what this makes possible (you'll need to get the PR obviously): https://www.reddit.com/r/resdemo/comments/3jbs3b/nested_comments/

Currently, if you want to get the above result without an active class on `.thing`, then you'd need to rewrite a significant portion of the comments' layout, which is a PITA.

### DIFF
--- a/lib/core/res.css
+++ b/lib/core/res.css
@@ -1741,11 +1741,13 @@ a spinner with a static ring and no overlay.*/
 	display: none; /* hide by default */
 	padding: 10px;
 }
-
 #RESStyleSheetTipPane-contents ul {
-	list-style: disc inside;
+	list-style: initial;
+	margin: 0 0 0 1em;
 }
-
+#RESStyleSheetTipPane-contents li p {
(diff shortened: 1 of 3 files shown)


honestbleeps:
I did a quick check of this manually (instead of pulling your branch, manually adding activeThing in the console -- this demo breaks keyboard navigation completely.

am I doing something wrong?

githue:
I can still use J, K, a, and z. Using Chrome here.

honestbleeps:
Yeah my mistake. Replied without sufficient brain.

PR #2449, opened 2015-09-10
Hover cards open more reliably, and use fadeElementOut

githue:
For Subreddit Info and User Info I like to have a short open delay and longer fade delay, but currently that causes this kind of thing to happen: https://imgrush.com/opq6CoX2YGVs

In that situation close() fires after open(), which prevents the card from ever appearing, unless you hover off the link and back over it. My solution is to check if the latest event type is `mouseenter`, and if so, close and reopen the card.

I also replaced the animated shrinking effect on close() with a simple fade, which affects all hover objects including the cog menu. I hope that's okay.

### DIFF
--- a/lib/modules/hover.js
+++ b/lib/modules/hover.js
@@ -168,7 +168,7 @@ addModule('hover', function(module, moduleID) {
 		},
 		target: function(element) {
 			if (typeof element !== 'undefined') {
-				this.close(true);
+				this.close();
 				this._target = element;
 			}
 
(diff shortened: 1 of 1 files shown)


lib/modules/hover.js
```diff
@@ -309,6 +309,12 @@ addModule('hover', function(module, moduleID) {
 				.off('mouseleave', this.cancelShow);
 		},
 		startShowTimer: function() {
+			this.mouseEvent = 'enter';
+
+			if (this.visible && this.mouseEvent === 'enter') {
```
andytuba: bit odd to check for mouseEvent when you've just set it..

githue: Whoops! Must have had tunnel vision. This.visible should be sufficient, I'll take another look.

andytuba:
Fade does feel better than shrink.  I kinda like the minor usability win of "this is where this tooltip came from" but I think this is better than the weird looking effect that hapens to the internals during shrink.

PR #2450, opened 2015-09-12
Move saved-RES tab to save space, improve UX

githue:
Created some inline tabs to save space in the main tab list. The URL remains the same so old links to it will still work. Cleaned up the formatting too.

[screenshot below]

Not sure if I'm using history() correctly since I haven't used it before. Not sure if data param is necessary.


erikdesjardins:
The NER indicator doesn't get hidden when you change tabs; otherwise it looks good.
image

It appears that the `history.pushState` data can be null or undefined, but it doesn't really matter since we don't use it.

githue:
saved-comments-v2

Updated the UI and refactored the code.

* Now using templates.html
* Makes the whole header of each comment a link to the comment.
* Gets the subreddit name and puts it in the header (is the regex ok?).
* When you unsave a comment, it keeps the comment until you reload the page, this allows you to recover from a miss-click.
* Hovering over "X days ago" shows a tooltip with the exact date.

lib/core/templates.html
```diff
@@ -573,5 +573,50 @@ <h3 class="RESHoverTitle" data-hover-element="0"></h3>
...
+						<li><b>{{savePostKey}}</b>: save a submission.</li>
+						<li><b>{{saveCommentKey}}</b>: save a comment (to your reddit account).</li>
+						<li><b>{{saveRESKey}}</b>: save a comment with RES.</li>
+					</ul>
+				</div>
+				<div class="res-saveComments-list"></div>
+			</div>
+		</div>
+
+		<div id="saveComments-comment">
```
erikdesjardins: Would be cool if this was all in the same template, in a `{{#comments}}` list.
Then you could eliminate `$wrapper.find('.res-module-description').hide();`, and wrap the description in `{{^comments}}{{/comments}}`, which will only be displayed if `comments` is falsy or an empty list.

lib/modules/saveComments.js
```diff
@@ -167,122 +159,118 @@ addModule('saveComments', function(module, moduleID) {
...
-				savedCommentsContent.appendChild(clearLeft);
-			}
+		$wrapper = $('#res-saveComments');
+
+		if (Object.getOwnPropertyNames(storedComments).length) {
+			$wrapper.find('.res-module-description').hide();
+		}
+
+		if (!modules['keyboardNav'].isEnabled()) {
+			$wrapper.find('.res-module-tip').hide();
```
erikdesjardins: This can be done in the template with `{{#keyboardNavEnabled}}...{{/keyboardNavEnabled}}`.

That is: if `foo` isn't an array, then `{{#foo}}bar{{/foo}}{{^foo}}baz{{/foo}}` acts like `foo ? 'bar' : 'baz'`.

PR #2463, opened 2015-09-27
Small cleanups

callahad:


honestbleeps:
thanks for these...

regarding the manifest changes, we'll need to do some testing to see if this would cause the infamous permission dialog problem and disable the extension for everyone on Chrome...

I think it may, unfortunately, which means we should consider whether the cleanup in the manifest is worth the merge or not.

the manifest change we should certainly take, getURL should've been used, for sure.

honestbleeps:
I should add, just for clarity: the manifest changes are unquestionably better, it's just a matter of whether it will cause us a gigantic flurry of tech support requests due to the security popup people will see about "new" permissions.

callahad:
That makes sense -- want me to split this into two PRs, or would you like to just cherry pick the getURL commit while we test the other?honestbleeps/Reddit-Enhancement-Suite wrote:I should add, just for clarity: the manifest changes are unquestionably better, it's just a matter of whether it will cause us a gigantic flurry of tech support requests due to the security popup people will see about "new" permissions.

mc10:
@callahad Not honestbleeps, but I think it's probably better to split into two PRs so we can at least merge one of them in. Cherry picking is kind of annoying.

callahad:
Opened #2464 instead. Closing this. We can revisit the manifest later. I suspect the explicit `http://` permissions will need to get updated as TLS becomes more common, so maybe we could roll this in whenever that happens. :)

honestbleeps:
thanks for understanding!  there's also the option of optimal permissions (will FF's WebExtension thing support those too?), but those get sort of dicey and are a bit buggy in Chrome apparently (we get "I said no to twitter, and now I can't get it to ask me again!" all the time, and we can't replicate that problem but it obviously happens because it comes up with some frequency.

PR #2464, opened 2015-09-28
Use getURL() instead of manual URL construction

callahad:
This is more idiomatic, and it works in Firefox today.

The current code will not work in Firefox until these bugs are resolved:

https://bugzilla.mozilla.org/1208761:
  chrome.i18n is undefined in content scripts

https://bugzilla.mozilla.org/1208763:
  Treat chrome-extension:// and moz-extension:// schemes as equivalent

This obsoletes #2463


honestbleeps:
thanks!

PR #2465, opened 2015-09-28
Don't presume window is the global object for js

callahad:
The bundled Favico lib exports itself by setting `this.Favico = Favico`.

The orangered module then presumes that Favico will be available on the
global `window` object, but that's not the case for Firefox's sandbox.
From [Bug 1208775][]'s discussion, it sounds like we might WONTFIX this.

Simply omitting `window.` works equally well in Chrome and Firefox.

[Bug 1208775]: https://bugzilla.mozilla.org/show_bug.cgi?id=1208775#c3

### DIFF
--- a/Opera/opera-footer.js
+++ b/Opera/opera-footer.js
@@ -1,6 +1,6 @@
 // It's not great that favico assumes this is the global
 // and that it gets "fixed" at this stage, but it's not terrible.
-window.Favico = this.Favico;
+var Favico = this.Favico;
 
 // Close self-invoking anonymous function
 })();
--- a/lib/modules/orangered.js
+++ b/lib/modules/orangered.js
@@ -197,7 +197,7 @@ addModule('orangered', function(module, moduleID) {
 		$('head link[rel="shortcut icon"], head link[rel="icon"]').attr('href', faviconDataurl);
 
 		// Init Favico
-		favicon = new window.Favico();
+		favicon = new Favico();
 
 		if (module.options.resetFaviconOnLeave.value) {
 			// Prevent notification icon from showing up in bookmarks


andytuba:
Could you also update https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/Opera/opera-footer.js#L3 so `var Favicon = this.Favicon`?  In Opera12, all of RES runs inside a huge IIFE.

callahad:
I literally haven't touched Opera 12 in years, and don't have any experience or knowledge re: debugging add-ons for it, so this is untested. But logically, it should work. Force-pushed a new rev of the patch.

andytuba:
Cool, thanks!  I'm double-check it runs before pushing Opera release.

callahad:
Thanks! :+1:

callahad:
As far as I can tell, Opera 12 doesn't seem to work at all. Neither self-built master nor v4.5.4 from the [Opera add-ons site](https://addons.opera.com/en/extensions/details/reddit-enhancement-suite-2/?display=en) run successfully.

This might be a larger policy discussion, but... maybe it's time to drop support for Opera 12? It hasn't seen a security patch in over a year, comments indicate that RES [hasn't worked since at least April](https://addons.opera.com/en/extensions/details/reddit-enhancement-suite-2/?display=en&reports#feedback-container), and virtually no one seems to be complaining [in the issues](https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues?utf8=%E2%9C%93&q=is%3Aissue+opera).

Heck Given the lack of updates, I'd personally feel it's irresponsible to continue supporting users on that platform, but that's your call. :)

Short term, maybe we could merge this since it certainly can't break Opera support any more than it's currently broken, and revisit Opera altogether in another bug?

callahad:
bump?

andytuba:
Thanks for tidy-up! 

Opera12 was working for a little bit recently in dev. I got it working again mostly for entertainment value and partly to be nice to the remaining few dozen users. Maybe I'll get it running again if it's simple.

PR #2472, opened 2015-10-07
Cleaned Up Links by Removing Redirects

theGeekPirate:


allthefoxes:
well..i mean..

ok..

theGeekPirate:
@allthefoxes Treat all text the same way you would code.

I would remove the www's as well, but those are handled individually by most HTTP servers, and I wouldn't want to risk them not working in the future because of it.

honestbleeps:
a trailing slash is something debatable, not canonically correct or incorrect.

arguments against the trailing slash:

- not technically necessary / extra character

arguments for the trailing slash:

- gives context, denotes the URL is definitely not a direct link to a file
- denotes the URL may likely be a link to a "section" on a website, or a collection of resources (not necessarily a directory)

Whether or not one finds the trailing slash at the end of a URL aesthetically pleasing or displeasing is something subjective.

Because this is a pretty subjective area, and one of nitpicking and not code quality, I'll wait until at least one of the other maintainers of this project weighs in before either merging or closing.

theGeekPirate:
https://developer.apple.com/devcenter/safari/index.action redirects to https://developer.apple.com/safari/, but the instructions contained are no longer correct as the site seems to have changed. Needs to be updated by someone who understands this process.

https://developer.apple.com/programs/safari also now redirects to https://developer.apple.com/programs/, and should be looked into by someone more knowledgeable, as I cannot find where to sign up for a free account as instructed.

For the meantime, those sections are incorrect.

allthefoxes:
I would treat it more like code if it were code. 

Personally I just see it as a pointless commit.

theGeekPirate:
@allthefoxes It is a form of code, as it affects where the URL leads to. http://example.com/blog does not necessarily lead to http://example.com/blog/, not to mention forcing redirects where they aren't needed.

honestbleeps:
I'd prefer it we keep the conversation respectful. There are valid reasons / preferences for both formats.

Regarding example.com/blog versus example.com/blog/ - whether or not one forces unneeded redirects to the other is, again, subjective.  Many sites 301 and add the trailing slash, rather than remove it.

This is why I've tried to stay neutral here and be both respectful and objective.  

Per [google's own article on the topic of trailing slashes](http://googlewebmastercentral.blogspot.com/2010/04/to-slash-or-not-to-slash.html), they intentionally do not recommend which is preferred (trailing slash or no):

> Steps for serving only one URL version

> What if your site serves duplicate content on these two URLs:

> http://<your-domain-here>/<some-directory-here>/
> http://<your-domain-here>/<some-directory-here>

> meaning that both URLs return 200 (neither has a redirect or contains rel=”canonical”), and you want to change the situation?

> Choose one URL as the preferred version. If your site has a directory structure, it’s more conventional to use a trailing slash with your directory URLs (e.g., example.com/directory/ rather than example.com/directory), but you’re free to choose whichever you like.

Sites may 301 redirect from example.com/foo to example.com/foo/ or vice versa - it's up to the site administrator's preference and neither can be relied upon to be canonical.

theGeekPirate:
@honestbleeps Right, which is what my last commit has changed, it simply avoids all redirects, regardless of a trailing slash or not (with preference to without).

theGeekPirate:
The larger issue at this point is that your documentation is incorrect RE: Apple Developer program, as I commented above.

mc10:
I'm personally not a fan of removing the trailing slash off root URLs, as copying them from the address bar will just copy them with a `/`, and the `/` is not actually absent&mdash;browsers just hide it. For trailing slashes off URLs like /r/Enhancement, I think the general guideline is what the site defaults to, which (in the case of Reddit) is to keep the trailing slash.

PR #2475, opened 2015-10-09
Surface module documentation

andytuba:
because it's gotten a bit hard to find in the last year


lib/reddit_enhancement_suite.user.js
```diff
@@ -42,140 +42,9 @@ RESMetadata.updatedURL = 'http://redditenhancementsuite.com/whatsnew.html?v=' +
...
-function escapeHTML(str) {
-	return (str === undefined || str === null) ?
-		null :
-		str.toString().replace(/[&"<>]/g, function(m) {
-			return escapeLookups[m];
-		});
-}
-
-// This alias is to account for Opera having different behavior...
-if (typeof navigator === 'undefined') {
```
andytuba: This is already in Opera header

PR #2481, opened 2015-10-12
In-line viewer support for imgur.com/gallery URLs

listentous:
Updated handleLink() to make /gallery URLs use the album viewer (if ID is < 6 chars) or (otherwise) the image viewer.

radd.it has been converting URLs this way for over a year and I've seen no exceptions.

### DIFF
--- a/lib/modules/hosts/imgur.js
+++ b/lib/modules/hosts/imgur.js
@@ -62,10 +62,13 @@ modules['showImages'].siteModules['imgur'] = {
 
 		if ((groups = siteMod.galleryHashRe.exec(href))) {
 			hash = groups[1];
-			def = siteMod._api('gallery/' + encodeURIComponent(hash));
+
+			// was: def = siteMod._api('gallery/' + encodeURIComponent(hash));
+			if (hash.length < 6) href = 'https://imgur.com/a/' + hash;	// imgur IDs < 6 chars are always albums
+			else href = 'https://i.imgur.com/' + hash + '.png';			// extension doesn't matter
 		}
 
-		else if ((groups = siteMod.albumHashRe.exec(href))) {
+		if ((groups = siteMod.albumHashRe.exec(href))) {
 			if (modules['showImages'].options['prefer RES albums'].value === true) {
 				hash = groups[1];
 				elem.imgHash = hash;


andytuba:
imgur used 5-character IDs for both images and albums up until about 2 years ago, so it's not guaranteed (especially when looking through /r/subreddit/top) that a 5-char ID will be an album. That's why RES asks the imgur gallery API endpoint to clarify.

listentous:
Fair enough.  I will also admit that it's not the most future-proof solution either as albums will probably creep up to 6 digits.

Hmm.  Since we know the rough cut-off date of when images stopped having 5-digit IDs, could this idea be implemented with an additional check on the posts created_utc?  Is that data available in handleLink()?  If it was posted in 2015, the odds of it being an older 5-digit ID has to be acceptably-low

I want all my imgur links to have expandos, damneet!  The gallery API doesn't seem to accomplish it.

andytuba:
Incidentally, Imgur's IDs are actually hashes instead of incremented IDs. With the hashing scheme used to generate IDs, they should always be an odd number of characters. I don't know how suffused the hash space currently is, but 36^5 is allotment to albums and pre-2013 images and 36^7 for images since 2013 leaves a little bit of wiggle room.

handleLink could get the post date via `RESUtils.thing(link)` -- I'd want to add in another method to fetch that -- but unsure if it's necessary yet.  I'd hazard the chances are higher than I'd like, since there are a lot of pre-2013 images that might still get reposted with the same ID. 

Can you link me any imgur.com/gallery/:id URLs which the latest dev build of RES isn't handling properly?

listentous:
Wait-- were /gallery URLs even a thing 2 years ago?  I recall them being slightly-more recent than that.  If they were introduced after images switched to the 6-digit ID/ hash then my method would always work, yeah yeah?

/gallery URLs not expandoing: https://redd.it/3orjj4  https://redd.it/3opdfo

andytuba:
Usersub /gallery was introduced about two years ago, yeah.

The gallery ID is the same as the original image/album ID, e.g. http://imgur.com/12345 when published to gallery becomes http://imgur.com/gallery/12345.  So, if you uploaded the image since 2013 and published to gallery, then you'll have definitely have a URL like /gallery/1234567.  Lower probability but still possible is taking an image uploaded pre-2013 getting published/pushed into the gallery, so it *could* have a /gallery/12345 URL.  All albums, of course, will be /a/12345 -> /gallery/12345.

listentous:
It's probably a safe assumption that /gallery/ and the 6-digit pic IDs were implemented at the same time.  Makes sense to me-- imgur doesn't want to mix up album and pic IDs either.  If that's true, then my method will work 100% of the time.

Even old pics like https://i.imgur.com/lce0i.jpg can't use the /gallery/ URL.  https://imgur.com/gallery/lce0i automatically forwards you to the old style http://imgur.com/lce0i  (and even forces HTTP!)

Your call but I think this solution is more reliable than even I originally assumed.  Plus it eliminates some pings between RES and imgur.

listentous:
Wanna see something fun?  https://redd.it/3qaoj7  Pic link to a imgur.com/gallery URL w/ a 5-digit ID.

What's fun about it?  Even imgur's gallery API gets it wrong and thinks it's an album.

erikdesjardins:
What version of the API are you using? Current dev build handles it fine (uses imgur's v3 API).
image

listentous:
Using the latest release version.   In my browser that page looks a bit different.

screencapture-www-reddit-com-r-oldschoolcool-comments-3qaoj7-a_moment_in_time_with_dizzy_gillespie_1980-1445878305854

Closing this pull request since it's not going to be used.

PR #2490, opened 2015-10-19
Fix voteWeight and hiding archived posts' arrows

### LINKED ISSUE #2486: Vote weights broken again
I've been using a self-built version of RES directly from this repository to get back the vote weights. Today I noticed that they don't work any more. I looked a bit through the code and found that the arrow class changed from upmod to up and from downmod to down, respectively (see https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/userTagger.js#L344). Changing RES to account for that did not restore the functionality. Unfortunately I'm not an extension developer so I don't have the tools to debug this further. Hence, I'm submitting this report.

erikdesjardins:
Fixes #2486.

### DIFF
--- a/lib/modules/styleTweaks.js
+++ b/lib/modules/styleTweaks.js
@@ -220,8 +220,8 @@ modules['styleTweaks'] = {
 				RESUtils.addCSS('.res .compressed .expando-button { display: block; }');
 			}
 			if (this.options.hideUnvotable.value) {
-				RESUtils.addCSS('.unvoted .arrow[onclick*=unvotable] { visibility: hidden }');
-				RESUtils.addCSS('.voted .arrow[onclick*=unvotable] { cursor: normal; }');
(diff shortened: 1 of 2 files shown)


13steinj:
You may wanna wait, since supposedly /u/madlee may change some more things regarding voting, that could potentially even break this fix, who knows.

erikdesjardins:
Yeah, sounds like a good idea - thanks for the heads-up.

13steinj:
Yup. It should be safe when the voting js changes get pushed to the repo though.

13steinj:
@erikdesjardins [It should be safe](https://www.reddit.com/r/redditdev/comments/3p8zxd/was_the_change_in_the_voting_javascript_closed/cw5a5p4), though you may wanna double check.

erikdesjardins:
Yep, still works :+1:

PR #2493, opened 2015-10-22
sticky expandos (#2376) round two

andytuba:
needs debug


andytuba:
wrong repo, probably stale and useless or something

PR #2501, opened 2015-11-01
Fix unsubscribing from threads via notification button 

githue:
The button wasn't getting thread ID correctly. `subObj.id` > `commentID`


andytuba:
Thanks!

PR #2503, opened 2015-11-02
Automatically toggle nightMode after loading neverEndingReddit page

### LINKED ISSUE #2502: nightmode: automatically toggle automatic nightmode when the time comes
I think it would be pretty neat if nightmode automatically toggled when the time comes.  Usually I'll be browsing down on page 30+ of RES and I'll click a comment only to see night mode has been enabled -- and it is likely past my bed time.  

I can't help but notice there's an event that never ending reddit will fire when it has finished loading a page - I think this is the perfect event to check for whether automatic nightmode should be enabled or not... I think for the most part, the only time I wish it would toggle automatically is when using never ending reddit.

It would be possible to figure out the time it should be enabled and set a timeout for that time -- but that would probably be more effort for what it's worth.  I'm creating a PR with the method I had in mind...

tswaters:
- adds an event handler for neverEndingLoad after the module loads
- event handler is removed if the module gets disabled
- when the event fires, perform a call to 'handleAutomaticNightMode'

Fixes #2502 

Also, I should mention, I updated the event firing to use the more modern API -- `initEvent` is [showing a deprecation warning on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent)

I tried this out locally on FireFox -- setting the time to a minute in the future with nightmode off, waiting for a minute and scrolling down -- it seemed to work pretty well.

### DIFF
--- a/lib/modules/nightMode.js
+++ b/lib/modules/nightMode.js
@@ -64,6 +64,16 @@ addModule('nightMode', function(module, moduleID) {
 		// If the module is turned off, disable night mode completely
 		if (!toggle) {
 			module.disableNightMode();
+			// ensure the event handler listening for siteTable changes has been removed.
+			if (module.siteTableSubscription) {
+				module.siteTableSubscription();
+				module.siteTableSubscription = null;
+			}
+		}
+		else {
+			if (!module.siteTableSubscription) {
+				module.siteTableSubscription = RESUtils.watchForElement('siteTable', module.afterNeverEndingLoad);
+			}
 		}
 	},
 	loadDynamicOptions: function() {
@@ -116,6 +126,14 @@ addModule('nightMode', function(module, moduleID) {
 				}
 			);
 		}
+	},
+	afterLoad: function () {
+		// add a subscription to site table loads - save the return so we can remove it later.
+		module.siteTableSubscription = RESUtils.watchForElement('siteTable', module.afterNeverEndingLoad);
+	},
+	afterNeverEndingLoad: function () {
+		// once RES loads a page, do a check to see if automatic nightmode should toggle.
+		modules['nightMode'].handleAutomaticNightMode();
 	},
 	isNightModeOn: function() {
 		return module.isEnabled() && module.options.nightModeOn.value;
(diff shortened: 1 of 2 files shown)


lib/modules/neverEndingReddit.js
```diff
@@ -493,8 +493,8 @@ modules['neverEndingReddit'] = {
 							modules['neverEndingReddit'].NERFail(noresultsfound);
 						}
 
-						var e = document.createEvent('Events');
-						e.initEvent('neverEndingLoad', true, true);
+						// fire an event so other modules can respond to page loads.
+						var e = new Event('neverEndingLoad');
```
andytuba: Sorry for the lack of comment. This event is for the benefit of other extensions, not RES internal modules.

lib/modules/nightMode.js
```diff
@@ -117,6 +119,14 @@ addModule('nightMode', function(module, moduleID) {
 			);
 		}
 	},
+	afterLoad: function () {
+		// after loading, create event handler for when neverEndingEeddit finishes a page.
+		window.addEventListener('neverEndingLoad', modules['nightMode'].afterNeverEndingLoad);
```
andytuba: It would be tidier to listen for RES's siteTable.. syntax is something like `RESUtils.watch.siteTable.add(callback)` -- I can grab it tonight if you don't find it first.

andytuba:
This is a cool idea, thanks for the contribution and the infrastructural update! We should regression check the `new Event` change in the other browsers, too.

tswaters:
There's no way to remove the callbacks aside from modifying `RESUtils.watchers.siteTable`... so if someone disables nightMode the callback would stay until they refreshed.

A cool pattern to deal with this sort of thing I've seen was with atom.io -- all the events they attach return a Disposable which can be used to remove the listener.

I could enhance the `watchForElement` api... right now it doesn't return anything, so it could return a function that when called would remove the callback (by splicing on the index).... stay tuned!

andytuba:
It would be great to replace the watchers with a better emitter like you described. It is an old and somewhat clunky pattern.

tswaters:
Ok, well, poor-man's disposable pattern coming right up!

Also, I should note there seems to be a few bugs around the initial toggle to true.  It seems isNightModeOn was returning true even though it wasn't on.  I had to refresh for it to come around.

PR #2505, opened 2015-11-03
UserInfo friend toggle button fix

### LINKED ISSUE #2447: The friend button in user info doesn't always work
Two issues:

1. The toggle() function on [+friend] sends a href value, instead of just the username. So if the link is `<a href="/u/githue">`, then onclick will be `toggle('/u/githue')` instead of `toggle('githue')`.
2. The isFriend var is checking for the presence of the class 'friend', which means it only works for links inside `.tagline`. The check won't work for links in self posts and sidebars.

I don't know how the second issue should be solved.

githue:
Fixes #2447.

Sets correct attribute values on the friend toggle button.

Clicking the friend button expires the cache on the associated user which forces a new check the next time userInfo is opened.

Makes the throbber work the way it was originally intended to. The throbber element needs to be a variable before being passed to def.notify().

### DIFF
--- a/lib/modules/userInfo.js
+++ b/lib/modules/userInfo.js
@@ -98,32 +98,18 @@ addModule('userInfo', function (module, moduleID) { $.extend(true, module, {
 	},
 	showUserInfo: function(def, obj, context) {
 		def.notify('Loading...', '<span class="RESThrobber />');
-		var isFriend = obj.classList.contains('friend'),
-			test = obj.href.match(modules['userTagger'].usernameRE),
-			thisUserName;
+		var test = obj.href.match(modules['userTagger'].usernameRE),
+			thisUserName,
+			throbber = '<span class="RESThrobber"></span>';
 
 		if (test) {
 			thisUserName = test[1];
 		}
 		module.lastOpenedAuthorLink = obj;
-
-		var header = $('<div />');
-		header.append('<a href="/user/' + escapeHTML(thisUserName) + '">/u/' + escapeHTML(thisUserName) + '</a> (<a href="/user/' + escapeHTML(thisUserName) + '/submitted/">Links</a>) (<a href="/user/' + escapeHTML(thisUserName) + '/comments/">Comments</a>)');
-		RESUtils.getUserInfo(function(userInfo) {
-			var friendButton, myID;
-			myID = 't2_' + userInfo.data.id;
-			if (isFriend) {
-				friendButton = '<span class="fancy-toggle-button toggle" style="display: inline-block; margin-left: 12px;"><a class="option active remove" href="#" tabindex="100" onclick="return toggle(this, unfriend(\'' + obj.textContent + '\', \'' + myID + '\', \'friend\'), friend(\'' + obj.textContent + '\', \'' + myID + '\', \'friend\'))">- friends</a><a class="option add" href="#">+ friends</a></span>';
-			} else {
-				friendButton = '<span class="fancy-toggle-button toggle" style="display: inline-block; margin-left: 12px;"><a class="option active add" href="#" tabindex="100" onclick="return toggle(this, friend(\'' + obj.textContent + '\', \'' + myID + '\', \'friend\'), unfriend(\'' + obj.textContent + '\', \'' + myID + '\', \'friend\'))">+ friends</a><a class="option remove" href="#">- friends</a></span>';
-			}
-			header.append(friendButton);
-			def.notify(header, null);
-		});
-
-		def.notify(header, null);
+		def.notify(thisUserName, throbber);
 
 		RESUtils.cache.fetch({
+			key: thisUserName.toLowerCase(),
 			method: 'GET',
 			endpoint: '/user/' + thisUserName + '/about.json',
 			callback: module.renderAuthorInfo.bind(module, def, obj)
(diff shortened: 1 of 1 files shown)


lib/modules/userInfo.js
```diff
@@ -97,33 +97,18 @@ addModule('userInfo', function (module, moduleID) { $.extend(true, module, {
...
-			}
-			header.append(friendButton);
-			def.notify(header, null);
-		});
-
-		def.notify(header, null);
+		def.notify(thisUserName, throbber);
 
 		RESUtils.cache.fetch({
+			key: thisUserName.toLowerCase(),
```
erikdesjardins: In this case, this is unnecessary since the endpoint URL will be different for each user (if no key is specified, the key is generated from the endpoint - e.g. `RESUtils.cache.-user-erikdesjardins-about-json`).

githue: Ah, I see.

lib/modules/userInfo.js
```diff
@@ -142,6 +126,33 @@ addModule('userInfo', function (module, moduleID) { $.extend(true, module, {
...
 		data.name = data.name.toLowerCase();
+
+		var header = $('<div />');
+		header.append('<a href="/user/' + escapeHTML(data.name) + '">/u/' + escapeHTML(data.name) + '</a> (<a href="/user/' + escapeHTML(data.name) + '/submitted/">Links</a>) (<a href="/user/' + escapeHTML(data.name) + '/comments/">Comments</a>)');
+		
+		// Add friend button to header.
+		RESUtils.getUserInfo(function(me) {
+			var friendButton, myID = 't2_' + me.data.id;
+			// Check cached value first, or fall back to initial server response.
+			if (data.res_cache_friend || (data.is_friend && data.res_cache_friend === undefined)) {
```
erikdesjardins: `data.res_cache_friend` should be unnecessary (it should send a new request when it's expired, so `data.is_friend` would always be correct), but it's a bit broken (#2506).

githue: Oh that will be nice. It felt a bit dirty modifying properties, but it was the only thing that worked for me.

lib/modules/userInfo.js
```diff
@@ -142,6 +126,29 @@ addModule('userInfo', function (module, moduleID) { $.extend(true, module, {
...
+			header.append(friendButton);
+			def.notify(header, 'Got friend status, loading details...');
+			header.find('.toggle').click(function (e) {
+				if (e.target.tagName !== 'A') return;
+				// Expire the cache to re-check friend status.
+				RESUtils.cache.expire({key: 'RESUtils.cache.-user-' + data.name + '-about-json'});
+			});
+		});
+
+		def.notify(header, 'Getting friend status...');
```
githue: It looks like we may not need either of those def.notify calls in renderAuthorInfo(). Correct? Nothing changes when I remove them.

erikdesjardins: Yep, correct.
The one on this line would do something, but it's not noticeable since the promise is resolved immediately after (on [line 185](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2505/files#diff-a4534d46d42c0c1d56651e222f88c168R185)).
The one on [line 142](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2505/files#diff-a4534d46d42c0c1d56651e222f88c168R142) always occurs after the promise is resolved, so it does nothing.

lib/modules/userInfo.js
```diff
@@ -142,6 +126,29 @@ addModule('userInfo', function (module, moduleID) { $.extend(true, module, {
...
+				friendButton = '<span class="fancy-toggle-button toggle" style="display: inline-block; margin-left: 12px;"><a class="option active remove" href="#" tabindex="100" onclick="return toggle(this, unfriend(\'' + data.name + '\', \'' + myID + '\', \'friend\'), friend(\'' + data.name + '\', \'' + myID + '\', \'friend\'))">- friends</a><a class="option add" href="#">+ friends</a></span>';
+			} else {
+				friendButton = '<span class="fancy-toggle-button toggle" style="display: inline-block; margin-left: 12px;"><a class="option active add" href="#" tabindex="100" onclick="return toggle(this, friend(\'' + data.name + '\', \'' + myID + '\', \'friend\'), unfriend(\'' + data.name + '\', \'' + myID + '\', \'friend\'))">+ friends</a><a class="option remove" href="#">- friends</a></span>';
+			}
+			header.append(friendButton);
+			def.notify(header, 'Got friend status, loading details...');
+			header.find('.toggle').click(function (e) {
+				if (e.target.tagName !== 'A') return;
+				// Expire the cache to re-check friend status.
+				RESUtils.cache.expire({key: 'RESUtils.cache.-user-' + data.name + '-about-json'});
```
erikdesjardins: One last thing: this could be `{ endpoint: '/user/' + data.name + '/about.json' }` - it'll generate the key the same way as `fetch`.

It's my bad, my previous comment was a bit misleading.

githue:
Each time renderUserInfo() was executed, it was accessing the logged in user's JSON file, this change fixes that. Added a cache.expire() on 116 because otherwise the cached version would load before loggedInID got defined.

PR #2510, opened 2015-11-08
Bring in changes from Chrome/background.js.

mryanmurphy:
Missing sendResponse in delete cookies broke account switcher.


andytuba:
Thanks!

PR #2513, opened 2015-11-10
Removed autoplay and replaced embed.radd.it

listentous:
To save money on its SSL cert, embed.radd.it is no longer supported via HTTPS.  Updated code to use the new ?embed format instead.  Autoplay is no longer supported.  I've removed it from radd.it's options.

All embedded links are currently broken, so please include this in the next release!

### DIFF
--- a/lib/modules/hosts/raddit.js
+++ b/lib/modules/hosts/raddit.js
@@ -7,13 +7,6 @@ modules['showImages'].siteModules['raddit'] = {
 			value: true,
 			type: 'boolean'
 		},
-		// No longer supported.
-		// 'autoplay sidebar playlister': {
-		// 	dependsOn: 'show playlister toggle',
-		// 	description: 'Start playing the radd.it sidebar embed when opening it',
-		// 	type: 'boolean',
-		// 	value: true
-		// },
 		'always show sidebar playlister': {
 			description: 'Always show the radd.it embed in the sidebar when applicable',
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/raddit.js
```diff
@@ -7,12 +7,13 @@ modules['showImages'].siteModules['raddit'] = {
 			value: true,
 			type: 'boolean'
 		},
-		'autoplay sidebar playlister': {
-			dependsOn: 'show playlister toggle',
-			description: 'Start playing the radd.it sidebar embed when opening it',
-			type: 'boolean',
-			value: true
-		},
+		// No longer supported.
```
andytuba: May as well just delete these lines unless you intend to add them back soonish.  We've got git history.

lib/modules/hosts/raddit.js
```diff
@@ -26,7 +20,7 @@ modules['showImages'].siteModules['raddit'] = {
 			groups = hashRe.exec(elem.href);
 
 		if (groups) {
-			def.resolve(elem, '//embed.radd.it' + groups[1]);
+			def.resolve(elem, '//radd.it' + groups[1] + (groups[1].indexOf('?') == -1 ? '?embed' : '&embed'));
```
mc10: Minor nit, but you should use `===` instead of `==`.

PR #2519, opened 2015-11-15
remove reddit expando button for duplicate images

andytuba:
for consistency with removing reddit expandos when RES provides the expando

before:

screen shot 2015-11-15 at 12 53 43 pm

after:

screen shot 2015-11-15 at 12 51 22 pm


andytuba:
for consistency with removing reddit expandos when RES provides the expando

PR #2520, opened 2015-11-16
gulp build generates browser manifests

andytuba:
Update each browse manifest and in-code version number with values from package.json

**FYI** `npm install` after pulling -- this adds new dependencies.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -26,26 +26,26 @@ var rootBuildDir = 'dist',
 			// The file for addFileToManifest to modify when adding new hosts/modules
 			manifest: 'Chrome/manifest.json',
 			manifestProperties: [
-				{ key: 'title',			needle: /("title": ")(?:.*)(")$/ },
-				{ key: 'description', 	needle: /("description": ")(?:.*)(")$/ },
-				{ key: 'version',	 	needle: /("version": ")(?:[\d\.])+(")/ }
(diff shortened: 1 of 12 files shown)


gulpfile.js
```diff
@@ -94,10 +124,42 @@ gulp.task('build', function(cb) {
 			gulp.src(paths.src)
 				.pipe(gulp.dest(path.join(getBuildDir(browser), paths.dest)));
 		});
+
+		if (config[browser].manifest) {
+			del(path.join(getBuildDir(browser), config[browser].manifest));
```
erikdesjardins: Does this need to be deleted? Naively I'd assume that gulp would happily overwrite it.
Also, it's async; it should return a promise.

andytuba: Gulp dest seemed to be throwing "could not write, file exists" errors.. Oh yeah, I suppose I should wait to add new files into old delete is complete.

erikdesjardins: That's strange, it seems to work fine for me without it :/
Although I have encountered this before; I think I did something weird to fix it (like appending `./` to the dest path)

andytuba: Yuck, del accepts a callback instead of returning a promise.

erikdesjardins: [It should](https://www.npmjs.com/package/del#del-patterns-options), we probably need to update to 2.1.0

PR #2523, opened 2015-11-18
Add Hide Scores module

denizdogan:
This new module hides the score for submissions and/or comments until the user has cast their vote. The idea is to prevent users from being influenced by the votes of others before deciding for themselves.

Everything is disabled by default. There is an option to always hide scores, even after the user has voted.


erikdesjardins:
Looks good, but this should probably be part of voteEnhancements; I don't think we need a new module for it.

denizdogan:
I've now moved it into 'Vote Enhancements' and fixed a couple of minor bugs. Another function that we could add would be to always hide the user's own submission/comment scores. Shouldn't be too much work.

denizdogan:
I will make a new branch and pull request for this feature

PR #2526, opened 2015-11-19
Fix #2497 - Opening a link (consisting of only a hash) in a new tab opens the background page

### LINKED ISSUE #2497: Opening a link (consisting of only a hash) in a new tab opens the background page
https://www.reddit.com/r/RESissues/comments/3qt3e5/bug_selecting_a_markdown_link_with_an_invalid_url/

Basically, because RES has to hit the API from the [background page](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/ed66de559a2d3aac86a445221a6dcd0a1ba28dd5/Chrome/background.js#L177) to open new tabs, a link to `#foo` will go to `_generated_background_page.html#foo` when it's opened in a new tab using keyboard shortcuts.

For some reason, `chrome.tabs.create` seems to resolve relative URLs (`/bar`) based on the URL of the tab sending the request, but hashes (`#foo`) based on the background page's URL. ¯\\_(ツ)_/¯

denizdogan:
This is a simple fix for #2497

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -684,6 +684,12 @@ addModule('keyboardNav', {
 		var thisURL = link.getAttribute('href'),
 			button = (this.options.followLinkNewTabFocus.value) ? 0 : 1;
 
+		// we do this because chrome.tabs.create opens URLs consisting solely of fragments (#foo)
+		// relative to the background page instead of the current location
+		if (thisURL.substring(0, 1) === '#') {
+			thisURL = location.href + thisURL;
+		}
+
 		if (this.options.commentsLinkNewTab.value) {
 			var thisJSON = {
 				requestType: 'keyboardNav',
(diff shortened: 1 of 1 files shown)


lib/modules/keyboardNav.js
```diff
@@ -684,6 +684,12 @@ addModule('keyboardNav', {
 		var thisURL = link.getAttribute('href'),
 			button = (this.options.followLinkNewTabFocus.value) ? 0 : 1;
 
+		// we do this because chrome.tabs.create opens URLs consisting solely of fragments (#foo)
+		// relative to the background page instead of the current location
+		if (thisURL.substring(0, 1) === '#') {
+			thisURL = location.href + thisURL;
```
andytuba: This will break if the href already contains a hash. How about `location.pathname + location.search + thisURL` instead?

denizdogan: @andytuba: Yes, that seems like the most reasonable thing to do

PR #2528, opened 2015-11-19
Simplify module categories

### LINKED ISSUE #2374: Better module categorizing
Module categories are a little arbitrary right now. Could they be improved?

Reminder: modules can belong to multiple categories.

githue:
Closes #2374.

Brings the total number of categories down from 11 to 10. A loose rule I came up with for fighting ambiguity was that for each category, at least half of its modules should be exclusive. If the majority of modules for one category aren't exclusive to it, then either the category is too vague or some modules don't belong there.

The new category list:

1. About RES
2. My account
3. Users
4. Comments
5. Submissions
6. Subreddits
7. Appearance
8. Browsing
9. Productivity
10. Core

These were replaced or removed because I felt they were too specific (filters) or too vague (posts):

* Editing
* Style
* Filters
* Posts
* UI

### DIFF
--- a/lib/modules/commentPreview.js
+++ b/lib/modules/commentPreview.js
@@ -1,7 +1,7 @@
 addModule('commentPreview', {
 	moduleID: 'commentPreview',
 	moduleName: 'Live Comment Preview',
-	category: 'Editing',
+	category: 'Comments',
 	options: {
 		enableBigEditor: {
 			type: 'boolean',
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -1,7 +1,7 @@
 addModule('keyboardNav', {
 	moduleID: 'keyboardNav',
 	moduleName: 'Keyboard Navigation',
-	category: [ 'UI' ],
+	category: [ 'Browsing' ],
 	options: {
 		mediaBrowseMode: {
 			type: 'boolean',
--- a/lib/modules/settingsConsole.js
+++ b/lib/modules/settingsConsole.js
@@ -1,6 +1,6 @@
 addModule('settingsConsole', function(module, moduleID) {
 	module.moduleName = 'Settings Console';
-	module.category = 'Core';
+	category: 'Core';
 	module.description = 'Manage your RES settings and preferences';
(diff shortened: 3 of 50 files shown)


lib/modules/keyboardNav.js
```diff
@@ -1,7 +1,7 @@
 addModule('keyboardNav', {
 	moduleID: 'keyboardNav',
 	moduleName: 'Keyboard Navigation',
-	category: [ 'UI' ],
+	category: [ 'Browsing' ],
```
andytuba: Not subreddits?

andytuba: Oops wrong line

lib/modules/settingsConsole.js
```diff
@@ -1,6 +1,6 @@
 addModule('settingsConsole', function(module, moduleID) {
 	module.moduleName = 'Settings Console';
-	module.category = 'Core';
+	category: 'Core';
```
andytuba: Module.Category=

lib/modules/keyboardNav.js
githue: Why subreddits?

andytuba:
I like the new names, seem more human. Good consolidation, definitely a few rough edges tidied up.

lib/modules/keyboardNav.js
andytuba: Got confused, disregard

githue:
Hopefully it saves a few users having to ask where stuff is.

lib/modules/commentPreview.js
```diff
@@ -1,7 +1,7 @@
 addModule('commentPreview', {
 	moduleID: 'commentPreview',
 	moduleName: 'Live Comment Preview',
```
andytuba: This module should really be renamed "Live Snudown Preview" since it applies to posts, wiki, and messages in addition to comments.

thybag: Possibly worth just going with "Live Preview"? 
My suspicion would be that a lot of RES user's probably don't know what Snudown is.

mc10:
This should be properly merged into `master` as well.

PR #2530, opened 2015-11-19
Fix #2261 - Show /me/f/all link in subreddit manager

### LINKED ISSUE #2261: Show "/me/f/all" link in subreddit manager
This would be useful for gold users.

Show "me/f/all" link in subreddit manager setting for https://www.reddit.com/#!settings/subredditManager

denizdogan:
I'm not sure using class names from the HTML element is the best way to check if we're currently on /me/f/all, but it works.

### DIFF
--- a/lib/modules/subredditManager.js
+++ b/lib/modules/subredditManager.js
@@ -882,7 +882,7 @@ addModule('subredditManager', {
 			} else if ($('html').is('.res-me-f-all')) {
 				subLower = 'f/all';
 			}
-			specialButtonSelected[subLower] = 'RESShortcutsCurrentSub'; // class name used for ac
+			specialButtonSelected[subLower] = 'RESShortcutsCurrentSub'; // class name used for active button
 
 			var shortCutsHTML = '';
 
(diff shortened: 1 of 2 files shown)


lib/modules/subredditManager.js
```diff
@@ -871,14 +876,20 @@ addModule('subredditManager', {
 			/* this probably isn't the best way to give the option, since the mechanic is drag/drop for other stuff..  but it's much easier for now... */
 			$(this.staticShortCutsContainer).html('');
 			var specialButtonSelected = {};
-			var subLower = (RESUtils.currentSubreddit()) ? RESUtils.currentSubreddit().toLowerCase() : 'home';
-			specialButtonSelected[subLower] = 'RESShortcutsCurrentSub';
+			var subLower = 'home'; // default
+			if (RESUtils.currentSubreddit()) {
+				subLower = RESUtils.currentSubreddit().toLowerCase();
+			} else if ($('html').is('.res-me-f-all')) {
```
andytuba: Hmm, users can disable the module that adds that class. Might should add a url regex in utils.js to check for this case.

andytuba: It might be a good idea for currentSubreddit to return "/f/all".

denizdogan:
@andytuba: Looking at the code for `currentSubreddit` makes me think, wouldn't it be better to rethink that function entirely? Currently it seems like it does two different things depending on whether an argument is given. If we make it also check if we're on /f/all, it'll start getting messy.

PR #2531, opened 2015-11-20
Add support for Derpibooru image host

### LINKED ISSUE #2527: Support for Derpibooru Media Host
[](/twibeam) I've started frequenting [/r/mylittlepony](https://www.reddit.com/r/mylittlepony) recently, and one of the more popular image hosts there (and on [all the other MLP subs](https://www.reddit.com/r/wehavetoomanymlpsubs)) is [Derpibooru](https://derpiboo.ru/).

[](/twisquint) I'd like to add support for that site to RES's image expand feature, but I want to make sure a PR for that would be welcome before I start working on it. Derpibooru [has an easy-to-use API](https://derpiboo.ru/pages/api), and I've already located [the place in RES where new media hosts are placed](https://github.com/honestbleeps/Reddit-Enhancement-Suite/tree/master/lib/modules/hosts), so I'm pretty confident this wouldn't be too difficult to implement, just give me the go ahead and I'll get to work. [](/ajhappy-in)

Ajedi32:
Fixes #2527

### DIFF
--- a/lib/modules/hosts/derpibooru.js
+++ b/lib/modules/hosts/derpibooru.js
@@ -179,6 +179,7 @@ addLibrary('mediaHosts', 'derpibooru', {
 
 	deferredImageRequests: [],
 	batchDebounceTime: 250,
+	batchMaxSize: 15,
 
 	doFetchBatch: function() {
 		var mod = modules['showImages'].siteModules['derpibooru'];
@@ -201,11 +202,14 @@ addLibrary('mediaHosts', 'derpibooru', {
 			imageID: imageID
 		});
 
-		RESUtils.debounce(
-			'showImages.derpibooru.doFetchBatch',
-			mod.batchDebounceTime,
-			mod.doFetchBatch
-		);
+		var debounceName = 'showImages.derpibooru.doFetchBatch';
+		if (mod.deferredImageRequests.length == mod.batchMaxSize) {
+			clearTimeout(RESUtils.debounceTimeouts[name]);
+			delete RESUtils.debounceTimeouts[name];
+			mod.doFetchBatch();
+		} else {
+			RESUtils.debounce(debounceName, mod.batchDebounceTime, mod.doFetchBatch);
+		}
 
 		return def.promise();
 	},
--- a/lib/modules/hosts/derpibooru.js
+++ b/lib/modules/hosts/derpibooru.js
@@ -2,11 +2,8 @@ addLibrary('mediaHosts', 'derpibooru', {
 	name: 'Derpibooru',
 	domains: [
 		'derpiboo.ru',
-		'www.derpiboo.ru',
 		'derpibooru.org',
-		'www.derpibooru.org',
 		'trixiebooru.org',
-		'www.trixiebooru.org'
 	],
 	pathMatcher: /^\/(?:images\/)?(\d+)$/i,
 
--- a/lib/modules/hosts/derpibooru.js
+++ b/lib/modules/hosts/derpibooru.js
@@ -7,20 +7,20 @@ addLibrary('mediaHosts', 'derpibooru', {
 	],
 	pathMatcher: /^\/(?:images\/)?(\d+)$/i,
 
-	detect: function(_href, elem) {
-		var mod = modules['showImages'].siteModules['derpibooru'];
+	detect(_href, elem) {
+		const mod = modules['showImages'].siteModules['derpibooru'];
 
 		return mod.pathMatcher.test(elem.pathname);
 	},
 
-	fetchJSON: function(url) {
-		var def = $.Deferred();
+	fetchJSON(url) {
+		const def = $.Deferred();
 
 		RESEnvironment.ajax({
 			method: 'GET',
 			url: url,
-			onload: function(response) {
-				var json;
+			onload(response) {
+				let json;
 				try {
 					json = JSON.parse(response.responseText);
 				} catch (error) {
@@ -28,7 +28,7 @@ addLibrary('mediaHosts', 'derpibooru', {
 				}
 				def.resolve(json);
 			},
-			onerror: function(response) {
+			onerror(response) {
 				def.reject();
 			}
 		});
@@ -38,19 +38,19 @@ addLibrary('mediaHosts', 'derpibooru', {
 
 	// Fetches image information from the Derpibooru API for the specified ID
 	// numbers without trying to follow redirects in the case of duplicate images.
-	fetchMultipleImageDataIgnoringDups: function(imageIDs) {
-		var mod = modules['showImages'].siteModules['derpibooru'];
-		var def = $.Deferred();
+	fetchMultipleImageDataIgnoringDups(imageIDs) {
+		const mod = modules['showImages'].siteModules['derpibooru'];
+		const def = $.Deferred();
 
-		var apiURL = 'https://' + mod.domains[0] + '/api/v2/images/show.json?id_numbers=' + encodeURIComponent(imageIDs.join(','));
+		const apiURL = 'https://' + mod.domains[0] + '/api/v2/images/show.json?id_numbers=' + encodeURIComponent(imageIDs.join(','));
 
-		mod.fetchJSON(apiURL).then(function(json) {
+		mod.fetchJSON(apiURL).then((json) => {
 			if (Array.isArray(json.images)) {
 				def.resolve(json.images);
 			} else {
 				def.reject();
 			}
-		}, function() {
+		}, () => {
 			def.reject();
 		});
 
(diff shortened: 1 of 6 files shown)


lib/modules/hosts/derpibooru.js
```diff
@@ -0,0 +1,38 @@
+addLibrary('mediaHosts', 'derpibooru', {
+	name: 'Derpibooru',
+	domains: ['derpiboo.ru'],
+	matchRe: /^https?:\/\/(?:www\.)?derpiboo.ru\/\d+$/i,
```
erikdesjardins: Should probably allow query strings or hashes after the id, maybe something like `/^https?:\/\/(?:www\.)?derpiboo.ru\/\d+(?:$|[?#])/i`.

lib/modules/hosts/derpibooru.js
```diff
@@ -0,0 +1,38 @@
...
+			onerror: function(response) {
+				def.reject();
+			}
+		});
+		return def.promise();
+	},
+
+	handleInfo: function(elem, info) {
+		elem.type = 'IMAGE';
+		elem.src = info.image;
```
erikdesjardins: Should check here or above and reject if `!info.image`, since sometimes the response doesn't include one.

lib/modules/hosts/derpibooru.js
```diff
@@ -0,0 +1,38 @@
...
+			},
+			onerror: function(response) {
+				def.reject();
+			}
+		});
+		return def.promise();
+	},
+
+	handleInfo: function(elem, info) {
+		elem.type = 'IMAGE';
```
erikdesjardins: Might as well do `elem.caption = info.description;` and add the fix I described in #2532

lib/modules/hosts/derpibooru.js
erikdesjardins: Actually this seems to happen if an image is a duplicate; if you're feeling adventurous you could make another request if the response includes `duplicate_of` (some examples of that [here](https://www.reddit.com/domain/derpiboo.ru)).

lib/modules/hosts/derpibooru.js
Ajedi32: Yeah, that's starting to get a bit complicated for a Regex. Certainly still doable, but it might be cleaner if I used a real URL parser. Is https://developer.mozilla.org/en-US/docs/Web/API/URL an option, or does RES need to support older browsers?

lib/modules/hosts/derpibooru.js
Ajedi32: Ah, thanks. I was having trouble finding a place with a lot of Derpibooru images to test on. (/r/mylittlepony usually only has a lot of Derpibooru images immediately after a new episode comes out). I'll definitely give the `duplicate_of` bit a shot.

lib/modules/hosts/derpibooru.js
Ajedi32: Yeah, seems reasonable...

lib/modules/hosts/derpibooru.js
andytuba: RES is only for modern browsers: last few versions of Chrome, Firefox, OperaBlink, and last two or three versions of Safari. Safari 6 and Opera 12 get a token nod.

lib/modules/hosts/derpibooru.js
andytuba: I usually test media hosts on their domain listing: https://www.reddit.mom/domain/derpiboo.ru

lib/modules/hosts/derpibooru.js
Ajedi32: Yeah, I think I'll probably just use [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) then. It's supported in everything but IE.

Ajedi32:
To do: captions. (Test with fixes from #2534)

lib/modules/hosts/derpibooru.js
```diff
@@ -0,0 +1,95 @@
...
+		var def = $.Deferred();
+
+		var apiURL = new URL(elem.href);
+		apiURL.pathname += '.json';
+
+		mod.fetchJSON(apiURL.href).then(function(json) {
+			if (json.image) {
+				def.resolve(elem, json);
+			} else {
+				if (json.duplicate_of) {
```
andytuba: They don't do the redirect on the backend? That woulda been nice for API consumers..

Ajedi32: No, they don't. My guess is they wanted to make it possible to check the status of duplicate images.

Ajedi32:
Okay, well _that_ was a lot harder than expected. The new implementation has about 5x more LOC than the original. :sweat_smile: I think I've got this down now though.

A few notes on the implementation here I feel are worth mentioning.

First, the code attempts to roll strings of API requests spaced more than 0.25 seconds apart into a single batch request which fetches all the data for those images at once. The batch size is limited to 15 images by default, since that is the default page size for Derpibooru's search page.

The batching is implemented by running a search using the IDs of the images in question. (E.g `id_number:1006970 OR id_number:522355 OR ...`). This has two limitations which complicate things quite a bit. Firstly, as I mentioned before, by default a maximum of 15 images per request can be retrieved this way. Secondly, the search just flat out fails to find images which are either marked as a duplicate, or hidden by the [default filter](https://derpiboo.ru/filters/100073) on Derpibooru. For those types of images, separate API requests needs to be made for the images individually.

The next complication is duplicate images. For images marked as a duplicate, only the ID of the duplicate is given by the API. A separate request then needs to be made to fetch the image with the provided ID.

One note on errors handling. One type of error I'm currently just ignoring is the case of deleted images. Right now, I'm simply rejecting the deferred, which causes the following error in the console: `showImages: error detecting image expando for https://www.derpibooru.org/654479`. Let me know if you'd rather I handle that differently.

One final note is that you may have noticed I removed the code for retrying failed requests in the most recent commit. This is because it didn't seem to be necessary anymore in my testing (after I wrote the code to start batching requests). It is conceivable though that for some users it might still be needed though, so let me know if you'd rather I leave that code in.

[](/rhuhhuh) And that's it! :tada:

Ajedi32:
Oh yeah, one other thing I just remembered. Captions in this PR are implemented assuming #2534 has already been merged, so they won't work without it. If that's a problem, let me know.

lib/modules/hosts/derpibooru.js
yaotung: #2531

yaotung:
#2531

yaotung:
#2531

lib/modules/hosts/derpibooru.js
Ajedi32: @yaotung ??? :confused:

Ajedi32:
@yaotung ??? :confused:

erikdesjardins:
@Ajedi32 looks like a bot, I'll get someone on it

Ajedi32:
@erikdesjardins Why would someone make a bot to spam references to random pull requests though? Eh, weird.

Ajedi32:
Ah, looks like there's a changelog. Updated.

This should be ready to merge BTW. Let me know if there's anything you need me to change before that happens.

lib/modules/hosts/derpibooru.js
```diff
@@ -0,0 +1,216 @@
+addLibrary('mediaHosts', 'derpibooru', {
+	name: 'Derpibooru',
+	domains: [
+		'derpiboo.ru',
+		'www.derpiboo.ru',
+		'derpibooru.org',
+		'www.derpibooru.org',
+		'trixiebooru.org',
+		'www.trixiebooru.org'
```
erikdesjardins: A bit late but: `showImages` ignores subdomains when checking the domain, so `www.` can be removed...I'll add a note to `example.js`.

lib/modules/hosts/derpibooru.js
```diff
@@ -0,0 +1,216 @@
...
+	],
+	pathMatcher: /^\/(?:images\/)?(\d+)$/i,
+
+	detect: function(href, _elem) {
+		var mod = modules['showImages'].siteModules['derpibooru'];
+		href = new URL(href);
+
+		return mod.domains.indexOf(href.host) != -1 && mod.pathMatcher.test(href.pathname);
```
erikdesjardins: Checking the domain in `detect()` is redundant (I know, it's in every single module, even `example.js`...I'll fix that too), and you can use `elem.pathname` without having to create a URL object, since it's an `<a>` element.

Ajedi32: Thanks for the tip. I'll fix this over the weekend.

lib/modules/hosts/derpibooru.js
Ajedi32: Ah, okay. Yeah, definitely mention that in `example.js`; better documentation will help prevent others from making the same mistake in the future.

Ajedi32:
Fixed. Looks like there's some merge conflicts now though. Hold on, I'm gonna rebase.

Ajedi32:
There we go. Took me a minute to figure out the new build process, but everything should be working now.

Ajedi32:
I just found out there actually *is* a batch request API for Derpibooru, it's just undocumented. This feels much cleaner.

I also rebased, and tweaked a few things to take advantage of the new ES6 features available with Babel.

erikdesjardins:
Merged in #2851.

Thanks @Ajedi32! (sorry about the delay)

PR #2536, opened 2015-11-22
Don't ignore errors thrown from DeviantArt handleInfo method

### LINKED ISSUE #2533: Minor: DeviantArt Media Host will silently swallow errors raised in its `handleInfo` method
Minor issue I discovered while working on #2531: if an error is thrown in DeviantArt's handleInfo method, the error will be silently caught and ignored thanks to https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/hosts/deviantart.js#L16-L21

Ajedi32:
Fixes #2533

### DIFF
--- a/lib/modules/hosts/deviantart.js
+++ b/lib/modules/hosts/deviantart.js
@@ -13,12 +13,13 @@ addLibrary('mediaHosts', 'deviantart', {
 			url: apiURL,
 			// aggressiveCache: true,
 			onload: function(response) {
+				var json;
 				try {
-					var json = JSON.parse(response.responseText);
-					def.resolve(elem, json);
+					json = JSON.parse(response.responseText);
 				} catch (error) {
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/deviantart.js
```diff
@@ -13,12 +13,13 @@ addLibrary('mediaHosts', 'deviantart', {
 			url: apiURL,
 			// aggressiveCache: true,
 			onload: function(response) {
+				var json;
 				try {
-					var json = JSON.parse(response.responseText);
-					def.resolve(elem, json);
+					json = JSON.parse(response.responseText);
```
Ajedi32: Just in case it's not clear, resolve here was calling the handleInfo method, so exceptions thrown from that method were being caught by this try catch block here.

PR #2537, opened 2015-11-22
Avoid potential double-posting on CTRL+enter

lamby:
This change avoids potential double-posting on CTRL+enter although this means that you lose your comment in the event of an error when posting (internet down, etc. etc.). Not entirely convinced but hopefully this will encourage a more robust solution.

### DIFF
--- a/lib/modules/commentTools.js
+++ b/lib/modules/commentTools.js
@@ -376,6 +376,12 @@ addModule('commentTools', {
 					function(e) {
 						var currentForm = $(e.target).closest('form');
 						var saveButton = currentForm.find('.save')[0] || currentForm.find('#wiki_save_button')[0] || $('.BEFoot button')[0];
+
+						// Avoid potential double-posting
+						if (currentForm.hasClass('usertext')) {
+							$(e.target).val('');
+						}
+
 						RESUtils.click(saveButton);
 					}
 				);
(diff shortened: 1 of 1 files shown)


andytuba:
I wonder if there's something more robust we could hook into like the visibility of a "saving..." throbber.  

It'd be good to also do that more holistically, like with a click handler on the save button to prevent default if saving.

lamby:
@andytuba

> hook into like the visibility of a "saving..." throbber.
[..]
> separate option for this behavior.

Why a separate option if it were robust? I mean, is there really a genuine use-case to be able to press ctrl+enter very fast to get duplicate comments..?

andytuba:
> if it were robust

I'm thinking about potential failures, like reddit 502's, and now your text is gone.

lamby:
> I'm thinking about potential failures, like reddit 502's, and now your text is gone

Exactly. Do I not mention this above? :)

>  in case something unanticipated breaks 

That certainly makes sense, but I was mostly asking for clarity of communication in some kind of  hypothetical 100% robust and bug-free scenario.

andytuba:
Ah. I'm thinking more short-term than robust :p

lamby:
This version, at least, won't delete your comment..

erikdesjardins:
Sorry about the delay.

At least, it seems that at some point reddit implemented this on their end; if I repeatedly click "submit" or press ctrl-enter, I get "please wait a few seconds and try again" and only one comment is submitted.

PR #2410, opened 2015-08-15
Fix the userbar sometimes taking up 100% of the header height

erikdesjardins:
This issue:
image

PR #2538, opened 2015-11-23
Move most CSS to .scss, use body classes for optional CSS

erikdesjardins:
Depends on #2301.

Well, this fixes #2410, that's for sure. And it's *fast*.


lib/core/init.js
```diff
@@ -176,6 +183,29 @@ RESUtils.init._loadModules = function () {
 	RESUtils.init.await._deferred('sourceLoaded').resolve(); // kick everything off
 };
 
+RESUtils.init.addOptionsBodyClasses = function(mod, modId) {
```
erikdesjardins: This should probably be moved. But where?

yaotung:
#2538

PR #2548, opened 2015-12-03
Fix for userTaggerToolTip appearing offscreen

corylulu:
When userTags are edited to when on the very right side of the screen, the tooltip appears partially offscreen. This fix checks for that and moves the tooltip to the left of the userTag instead.

### DIFF
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -595,10 +595,11 @@ addModule('userTagger', {
 			}
 			document.getElementById('userTaggerColor').selectedIndex = 0;
 		}
-		if (thisXY.x + 334 < window.innerWidth) {
+		this.userTaggerToolTip.style.display = "block";
+		if (thisXY.x + this.userTaggerToolTip.offsetWidth < window.innerWidth) {
 			this.userTaggerToolTip.setAttribute('style', 'display: block; top: ' + thisXY.y + 'px; left: ' + thisXY.x + 'px; z-index: 20;');
 		} else {
-			this.userTaggerToolTip.setAttribute('style', 'display: block; top: ' + thisXY.y + 'px; right: ' + (window.innerWidth - thisXY.x - 20)  + 'px; z-index: 20;');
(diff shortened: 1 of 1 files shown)


lib/modules/userTagger.js
```diff
@@ -595,7 +595,11 @@ addModule('userTagger', {
 			}
 			document.getElementById('userTaggerColor').selectedIndex = 0;
 		}
-		this.userTaggerToolTip.setAttribute('style', 'display: block; top: ' + thisXY.y + 'px; left: ' + thisXY.x + 'px; z-index: 20;');
+		if (thisXY.x + 334 < window.innerWidth) {
```
githue: The width won't always be 334, use offsetWidth to get the actual width.

lib/modules/userTagger.js
```diff
@@ -595,7 +595,11 @@ addModule('userTagger', {
...
+			this.userTaggerToolTip.setAttribute('style', 'display: block; top: ' + thisXY.y + 'px; left: ' + thisXY.x + 'px; z-index: 20;');
+		} else {
+			this.userTaggerToolTip.setAttribute('style', 'display: block; top: ' + thisXY.y + 'px; right: ' + (window.innerWidth - thisXY.x - 20)  + 'px; z-index: 20;');
```
githue: Same for the tag icon, instead of '20', try obj.offsetWidth.

lib/modules/userTagger.js
corylulu: I originally did this, but offsetWidth returns 0 because it's not visible when called, it returns properly using jquery, but then it doesn't pass the built in test. I'll revise this however and find a suitable solution.

corylulu:
:haircut:

corylulu:
:haircut: @githue

corylulu:
:haircut: @githue

lib/modules/userTagger.js
erikdesjardins: You are definitely allowed to use jQuery; it failed because of the double quotes.
image

corylulu: Oh, well regardless, I have it working now without it.

corylulu:
:haircut:

PR #2563, opened 2015-12-13
XPI/index: Fix ESLint issues

mc10:
Mostly converting `let` to `const` for unchanged variables.

### DIFF
--- a/XPI/index.js
+++ b/XPI/index.js
@@ -5,40 +5,40 @@
 // require('sdk/preferences/service').set('javascript.options.strict', false);
 
 // Import the APIs we need.
-let pageMod = require('sdk/page-mod');
-let Request = require('sdk/request').Request;
-let self = require('sdk/self');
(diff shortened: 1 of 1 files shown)


XPI/index.js
```diff
@@ -120,8 +120,8 @@ tabs.on('activate', function() {
 });
 
 function getActiveWorker() {
-	let tab = tabs.activeTab;
-	for (let i in workers) {
+	const tab = tabs.activeTab;
+	for (const i in workers) {
```
erikdesjardins: Even though this is correct, Firefox (42) seems to only work if it's `let` ...dammit Firefox.
There are two other `for (const ...` that need to be changed too.

mc10: Odd, I see http://stackoverflow.com/questions/31987465/ecmascript-2015-const-in-for-loops too. I'll change it to fix Firefox, but hopefully Firefox fixes their implementation of `const`.

PR #2564, opened 2015-12-13
Rename Firefox and Safari folders

mc10:
For consistency with Chrome and Opera. Safari still gets transformed to `RES.safariextension` after `gulp`, but Firefox becomes `firefox`, not `XPI`.


erikdesjardins:
There are a few lines in the readme that reference these paths, IIRC.

mc10:
I believe I covered all the spots in the README.

PR #2550, opened 2015-12-06
Nightmode fixes, improvements, additions

githue:
Notable changes include fixing some over-specific selectors (https://github.com/stevenhoward/Reddit-Enhancement-Suite/commit/afc71ddfa61f3399cc0bc08d1bd57c6719f74005#commitcomment-14802877), improving the brightness of links, and a slight change to the appearance of buttons (1px outset border).

This gif shows difference between the old and new link colors:
nightmode-pretty-02-compare

There's a whole bunch of other updates which aren't very noteworthy, too.

PR #2541, opened 2015-11-27
Local images

erikdesjardins:
Fixes #2115 

I left named images alone (i.e. `https://redditstatic.s3.amazonaws.com/mailgray.png`), should they packaged with RES as well?

Tested in Chrome, Firefox, and Safari.

PR #2566, opened 2015-12-14
SCSS Nightmode fixes, improvements, additions

githue:
Same as #2550, but for master instead of 4.6.

### DIFF
--- a/lib/css/modules/_nightMode.scss
+++ b/lib/css/modules/_nightMode.scss
@@ -205,7 +205,7 @@
 	}
 
 	.tabmenu.formtab .selected a {
-		color: rgb(255,255,255);
+		color: rgb(255, 255, 255);
 		background-color: rgb(95, 153, 207);
 	}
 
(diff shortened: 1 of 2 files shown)


lib/css/modules/_nightMode.scss
```diff
@@ -258,16 +245,22 @@
...
-		&.upmod { background-position: 0 0 !important; }
-		&.downmod { background-position: 0 -14px !important; }
+	
+	/* Note: arrow class is used on other elements besides vote buttons. */
+	.arrow {
+		&.up,
+		&.upmod,
+		&.down,
+		&.downmod {
+			background-image: url(https://b.thumbs.redditmedia.com/WZDoQsDzUySVTErcweuA3k2HH1HU5-BZNnBXOR4FDwc.png);
```
andytuba: @erikdesjardins Would it be easier to merge this first or #2541?

erikdesjardins: Rebasing #2541 should be pretty straightforward; might as well do this first.

PR #2568, opened 2015-12-15
Fix night mode background colour of code in permalinked posts

ScoreUnder:
This has been a problem for me since forever. (Not anything to do with the recent night mode changes). The CSS selectors for code/pre tags aren't specific enough all the time.

I went into the inspector and copied out every selector which gave the code a light background colour, so I'm not sure if a couple of these are redundant, but what I do know is that it darkens the background for me as expected.

### DIFF
--- a/lib/css/modules/_nightMode.scss
+++ b/lib/css/modules/_nightMode.scss
@@ -468,9 +468,13 @@
 
 	.md code,
 	.md pre,
+	.link .md pre,
 	.content .md pre,
+	.new-comment .md pre,
+	.usertext.border .md pre,
 	.link .md :not(pre) > code,
-	.new-comment .md :not(pre) > code {
+	.new-comment .md :not(pre) > code,
+	.usertext.border .md :not(pre) > code {
 		color: inherit;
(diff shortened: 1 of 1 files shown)


allthefoxes:
I believe this is already fixed and ready to go for 4.60

githue:
Nightmode should be using the same selectors that reddit uses for code and pre tags, and reddit keeps adding more. As long as this follows the same rules then it should be good.

It's difficult to check right now because reddit's servers are failing.

ScoreUnder:
>I believe this is already fixed and ready to go for 4.60
I checked out the release branch (there doesn't seem to be a develop branch) and I can't find a similar fix.
The release branch also hasn't had master merged into it so the nightmode.css file would go away anyway.

erikdesjardins:
@githue is right, this seems to be novel - look at a permalinked comment or a new comment (with gold):

image

@ScoreUnder `master` is develop, FYI

githue:
:+1: Looks good.

erikdesjardins:
Thanks @ScoreUnder!

PR #2569, opened 2015-12-15
Fix pinHeader

githue:
On [/r/gonewild](http://reddit.com/r/gonewild) the spacer height wasn't being calculated correctly because it wasn't checking for margins. jQuery saves the day with outerHeight(true):
gonewild

On [/r/web_design](http://reddit.com/r/web_design) and [/r/lifeprotips](http://reddit.com/r/lifeprotips) the header would be in the middle of the viewport instead of at the top (but only if you enable custom subreddit styles *site-wide*):
web_design

Note: this doesn't completely fix the last two subs because the header has no background color, due to its height being zero. There's no graceful way to fix that.

I removed the timeout loop that waits for the logo image to finish loading, because Reddit now has a height attribute on logo images.

Pinning the user bar was broken because it was setting `top: 0` without unsetting `bottom`. (Edit: that was because I removed the height that was set somewhere else)

### DIFF
--- a/lib/modules/betteReddit.js
+++ b/lib/modules/betteReddit.js
@@ -579,7 +579,7 @@ addModule('betteReddit', function(module, moduleID) {
 		var userBarElement = document.getElementById('header-bottom-right');
 		var thisHeight = $('#header-bottom-right').height();
 		RESUtils.addCSS('#header-bottom-right:hover { opacity: 1 !important;  }');
-		RESUtils.addCSS('#header-bottom-right { height: ' + parseInt(thisHeight + 1, 10) + 'px; }');
+		//RESUtils.addCSS('#header-bottom-right { height: ' + parseInt(thisHeight + 1, 10) + 'px; }');
 		// make the account switcher menu fixed
 		window.addEventListener('scroll', RESUtils.debounce.bind(RESUtils, 'scroll.betteReddit', 300, handleScroll, userBarElement), false);
 		pinCommonElements();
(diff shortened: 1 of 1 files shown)


lib/modules/betteReddit.js
```diff
@@ -595,71 +593,52 @@ addModule('betteReddit', function(module, moduleID) {
...
-		}
-
-		// add a dummy <div> to the document for spacing
-		var spacer = document.createElement('div');
-		spacer.id = 'RESPinnedHeaderSpacer';
 
-		// without the next line, the subreddit manager would make the subreddit bar three lines tall and very narrow
-		RESUtils.addCSS('#sr-header-area {left: 0; right: 0;}');
-		spacer.style.height = $('#header').outerHeight() + 'px';
+		if (header === null) return; // Reddit is having issues.
```
erikdesjardins: Might as well log an error here

lib/modules/betteReddit.js
```diff
@@ -595,71 +593,52 @@ addModule('betteReddit', function(module, moduleID) {
...
+		// We can't change the z-index because themes often rely on it being a certain value.
+		RESUtils.addCSS('#header { top: 0; left: 0; right: 0; bottom: auto; }');
+
+		// Spacer height should be equal to header (including padding, margins, etc).
+		$(spacer).height(headerHeight);
+		document.body.insertBefore(spacer, header.nextSibling);
+
+		// Hide sidebar checkbox.
+		setTimeout(function () {
+			modules['styleTweaks'].setSRStyleToggleVisibility(false, 'pinHeader');
```
erikdesjardins: Probably want to avoid doing this, if possible (unless it's shown somewhere else). Does it overlap with the header or something?

andytuba:
@githue Would this be a good candidate for v4.6.2?

githue:
Yeah I think so. Even though the code has been rewritten the effective changes are rather minor, i.e. uses `outerHeight(true)` instead of `height()`, adds `top:` to `#header`, and removes `left: 0; right: 0;` from `#sr-header-bar`.

andytuba:
I think Safari also suffers a timing problem with attempting to determine and cache the header height/position before it's set up, so it'll be nice to resolve that along with this.

PR #2571, opened 2015-12-18
RES is adding too many tabmenus

githue:
Utility function RESUtils.getHeaderMenuList() was adding a new tabmenu every time it was called.

extra-tabmenus

The name of the function was misleading because it's not only returning the tabmenu, but it's creating one too. There's only one time it needs to create a new element and that's in showImages.js.

With the fix: 
extra-tabmenus-fix

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -323,7 +323,17 @@ addModule('showImages', function(module, moduleID) {
 	var viewImageButton;
 
 	function createImageButtons() {
-		var mainMenuUL = RESUtils.getHeaderMenuList();
+		var mainMenuUL = $('#header-bottom-left ul.tabmenu')[0];
+		// Create new tabmenu on these pages, regardless if one already exists.
+		if ((RESUtils.regexes.search.test(location.href)) ||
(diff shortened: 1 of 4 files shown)


lib/modules/showImages.js
```diff
@@ -323,7 +323,17 @@ addModule('showImages', function(module, moduleID) {
...
+		var mainMenuUL = $('#header-bottom-left ul.tabmenu')[0];
+		// Create new tabmenu on these pages, regardless if one already exists.
+		if ((RESUtils.regexes.search.test(location.href)) ||
+			(/\/about\/(?:reports|spam|unmoderated)/.test(location.href)) ||
+			(location.href.indexOf('/modqueue') !== -1) ||
+			(location.href.toLowerCase().indexOf('/dashboard') !== -1))
+		{
+			mainMenuUL = RESUtils.createElement('ul', null, 'tabmenu viewimages');
+			document.getElementById('header-bottom-left').appendChild(mainMenuUL);
+			mainMenuUL.style.display = 'inline-block'; // Override dashboard's subreddit style.
```
githue: It would be nice to fix this in the subreddit stylesheet, and remove the inline style.

PR #2572, opened 2015-12-19
Improved commentHoverBorder

Zunon:
I noticed that commentHoverBorder moves elements when you hover, may I suggest that you swap instead of using border, use outlines as that isn't considered when an elements dimensions are calculated, here's a quick codepen I threw up to demonstrate this. http://codepen.io/Zunon/full/Lpvzeg/

### DIFF
--- a/lib/css/modules/_commentBoxes.scss
+++ b/lib/css/modules/_commentBoxes.scss
@@ -41,6 +41,6 @@
 	}
 
 	&.res-commentHoverBorder .content .comment:hover {
-		border: 1px solid #99AAEE !important;
+		outline: 1px solid #99AAEE !important;
 	}
 }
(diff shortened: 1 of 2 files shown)


lib/css/modules/_commentBoxes.scss
```diff
@@ -41,6 +41,6 @@
 	}
 
 	&.res-commentHoverBorder .content .comment:hover {
-		border: 1px solid #99AAEE !important;
+		outline: 1px solid #99AAEE !important;
```
erikdesjardins: Hmmm. This will double-up on the existing gray border instead of just changing its color, and it doesn't play nice with commentRounded.

It might be preferable to use `border-color: #99AAEE !important`, which does (virtually) the same thing without those downsides. If a subreddit removes all of the borders it won't work, but that's still the status quo.

erikdesjardins:
Declining in favour of `border-color` (in #2893). Thanks!

PR #2574, opened 2015-12-21
gulp: Use unnamed arg for add-module/add-host

mc10:
This makes the command a bit less attractive code-wise, but since the `file` arg is required anyways, it makes more sense to not make it named.


andytuba:
:+1:

mc10:
Huh guess gulp treats the second argument as a task name... oh well. Closing for now unless I can find an alternative.

PR #2578, opened 2015-12-26
Add Hide Scores module

denizdogan:
(This is a previously submitted pull request, #2523, but under a new branch, instead of my master branch as before.)

----

This new module hides the score for submissions and/or comments until the user has cast their vote. The idea is to prevent users from being influenced by the votes of others before deciding for themselves.

Everything is disabled by default. There is an option to always hide scores, even after the user has voted.


erikdesjardins:
CSS handling has been greatly improved since you opened the first PR, so if it isn't too much trouble, could you bring this up to date with those changes?

You can look at `highlightScores` for an example, but basically you'll add `bodyClass: true` to the options, add your CSS to `lib/css/modules/_voteEnhancements.scss`, and you won't need to use `addCSS`.

denizdogan:
@erikdesjardins Yeah, I noticed that quite a bit has changed since I made the PR. I'll do my best to update all my pull requests to make the progress easier.

PR #2580, opened 2015-12-30
Prevent voteWeight from breaking when users have no tags

erikdesjardins:
`JSON.parse('null') === null`, so safeJSON doesn't save us.

Broken in master and 4.6.0.


andytuba:
I'll cherry pick this into release/4.6.0 tomorrow.

PR #2585, opened 2016-01-02
Clarify module descriptions and options

githue:
Is my wording for cleanComments fair?


andytuba:
Looks good to me. Once Page Navigator gets something more than a "Back to top" arrow, we can update it.

Issue #2544, opened 2015-11-29
How to build xpi via jpm


darkred:
I want to build the latest xpi via jpm. (win10x64)

Is the following the correct procedure?
_I have already installed node.js v5.1.0 x64_

```
git clone https://github.com/mozilla/jpm
cd jpm
npm install 
npm link
cd ..
git clone https://github.com/honestbleeps/Reddit-Enhancement-Suite
cd XPI
-----> Do I also need:  npm install  ?  
jpm xpi
```

andytuba:
Yes, you need to `npm install` and, every time you pull / edit, then `gulp build -b Firefox && cd dist/XPI && jpm xpi`

darkred:
So, if I got this right,
I do this the 1st time:


```
git clone https://github.com/mozilla/jpm
cd jpm
npm install 
npm link
cd ..
git clone https://github.com/honestbleeps/Reddit-Enhancement-Suite
cd XPI
npm install
```
and after every time I do a `git pull` in the Reddit-Enhancement-Suite directory, 
I have to do (have gulp installed and then) :
```
gulp build -b Firefox 
cd dist/XPI 
jpm xpi
```

<br>
Ιs `gulp build -b Firefox ` really necessary in order to build the xpi (not run it in a new profile)? 
Isn't `jpm` sufficient?
And, what's the purpose of the above command? 
I only found In [here](http://gruntjs.com/using-the-cli) that: 
```
--base, -b
Specify an alternate base path. By default, all file paths are relative to the Gruntfile.
```

mc10:
`gulp build -b Firefox` packages all of the files and moves them into the proper `dist` folder. The link you provided actually refers to Grunt, not gulp (confusing, especially since we still have both build systems). So yes, it is necessary to build the XPI.

darkred:
Thanks for your reply.
But, after

```
git clone https://github.com/honestbleeps/Reddit-Enhancement-Suite
cd XPI
npm install
```
I get `npm WARN EPACKAGEJSON reddit_res@4.5.4 No repository field`


_(using npm 3.3.12  (node.js 5.1.0 x64) in win 10 x64)_

andytuba:
Sorry, I missed a line earlier. 

You should run npm install from the RES root directory, not XPI.

darkred:
Thanks.
`npm install` showed various WARN and ERR messages:
http://pastebin.com/raw.php?i=471JZVQ4

[...]

PR #2596, opened 2016-01-06
Add Python and node.js version to readme

erikdesjardins:
Ref: https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2544#issuecomment-169068859

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ RES is built with [gulp](http://gulpjs.com/).
 
 First time installation:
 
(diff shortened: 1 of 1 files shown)


mc10:
We should also mention that [building on Windows is kind of a pain](https://github.com/nodejs/node-gyp#installation); see also nodejs/node-gyp#629.

13steinj:
@mc10 Extremely stupid question: can I build on Linux and transfer it over to Windows no problem?

mc10:
Should be fine.

PR #2600, opened 2016-01-08
Exclude showImages on reddit.com/subreddits

githue:
And /subreddits/search. Not only are there no images to show, but it's adding a second tabmenu on that page.


13steinj:
Just to chime in: Can't there be links via the descriptions? E.g. https://www.reddit.com/subreddits/search?q=13steinj (third from the bottom as of typing)

githue:
You're right, I hadn't noticed that. However, showImages doesn't look for images in that text, or in the sidebar, or submission guidelines.

13steinj:
Oh, never mind then. Just assumed it looked everywhere without actually checking the source.

PR #2602, opened 2016-01-08
CSS cleanup for subreddit themes

githue:
Found these CSS rules while developing a new theme, they appear to be unnecessary and only complicate things.

`max-height: 20px;` is doubled up.

`h1.redditname { overflow: auto; }` only serves to add scrollbars when it overflows.

`.comment .tagline { display: inline; }`, can't think of a good reason to have this, and it causes taglines to not extend all the way to the right.

The inline styles on .subButtons are presumably to prevent tampering, but A) it's not very effective, and B) even Reddit doesn't try to protect the subscribe button, which goes inside .subButtons.

Userbar hider is doing way more work than necessary.

### DIFF
--- a/lib/css/modules/_userTagger.scss
+++ b/lib/css/modules/_userTagger.scss
@@ -1,7 +1,3 @@
-.comment .tagline {
-	display: inline;
-}
-
 #userTaggerToolTip {
 	display: none;
 	position: absolute;
--- a/lib/css/modules/_subredditManager.scss
+++ b/lib/css/modules/_subredditManager.scss
@@ -157,7 +157,6 @@
 
 #RESShortcutsViewport {
 	width: auto;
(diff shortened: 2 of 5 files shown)


lib/css/modules/_userTagger.scss
```diff
@@ -1,7 +1,3 @@
-.comment .tagline {
-	display: inline;
```
erikdesjardins: I thought this might have been the fix for comments getting pushed below the sidebar, but it looks like it's been here forever: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blame/5bef0d2a2edc6fcf06fe760ffa2cb27d52d3eb94/lib/modules/userTagger.js#L51

So I don't see a reason to keep it either.

PR #2603, opened 2016-01-10
Use npm scripts (and local gulp) for building, misc build improvements

erikdesjardins:
Depends on #2693; [diff](https://github.com/erikdesjardins/Reddit-Enhancement-Suite/compare/more-lint...local-gulp).

This changes `gulp watch` to `npm start` (and other stuff, see the README), which should eliminate errors caused by global gulp being out-of-sync, and make it easier to debug.

It also upgrades our test infrastructure (see `lib/core/__tests__`) and shrinkwraps our dependencies (so if you add new ones, you'll need to rebuild `npm-shrinkwrap.json`).

@allthefoxes When this is merged, you'll have to use `npm run build -- chrome,firefox,safari && mv dist/zip/* /wherever`.

@erikdesjardins Note to self: appveyor and coveralls


lib/css/modules/_userTagger.scss
```diff
@@ -111,6 +111,11 @@ a.voteWeight {
 
 .RESUserTag {
 	color: black;
+
+	// unselectable in comments/markdown for usability
+	.md & {
```
andytuba: I wonder how long until somebody asks for this back.. but hey, we can give them a CSS snippet or a bodyclass or something.

erikdesjardins: That's existing behaviour, I just moved it from another file...I think.

erikdesjardins: Yeah https://github.com/erikdesjardins/Reddit-Enhancement-Suite/commit/2eadf20aea566c3e49204a888bf33f234b6a8c2c#diff-a34ff6560c6da1e0b6480724e857032aL2026

Unless you meant album captions/images, in which case I agree.

PR #2609, opened 2016-01-11
New custom filter: Logged in as [user]

morgil:
It can be useful to apply custom filters only when (not) logged in as a certain user, so I created a corresponding filter.

### DIFF
--- a/lib/modules/filteReddit.js
+++ b/lib/modules/filteReddit.js
@@ -586,6 +586,23 @@ addModule('filteReddit', function(module, moduleID) {
 						return thing.classList.contains('over18');
 					}
 				},
+				loggedInAs: {
+					name: 'When logged in as a certain user',
+					defaultTemplate: function(patt) {
+						return {type: 'loggedInAs', patt: patt || ''};
+					},
+					fields: [
+						'logged in as /u/', {type: 'text', id: 'loggedInAs', validator: RegExp}
+					],
+					evaluate: function(thing, data, config) {
+						var myName = RESUtils.loggedInUser();
+						if (!myName) {
+							return false;
+						}
+						var pattern = new RegExp('^('+data.loggedInAs+')$', 'i');
+						return pattern.test(myName);
+					}
+				},
 				toggle: {
 					name: 'Custom toggle',
 					defaultTemplate: function() {


lib/modules/filteReddit.js
```diff
@@ -586,6 +586,23 @@ addModule('filteReddit', function(module, moduleID) {
...
+					},
+					fields: [
+						'logged in as /u/', {type: 'text', id: 'loggedInAs', validator: RegExp}
+					],
+					evaluate: function(thing, data, config) {
+						var myName = RESUtils.loggedInUser();
+						if (!myName) {
+							return false;
+						}
+						var pattern = new RegExp('^('+data.loggedInAs+')$', 'i');
```
andytuba: How about a simple `data.toLowerCase() === `myName`?  Remember this code runs on every single thing.

andytuba:
Neat!

lib/modules/filteReddit.js
morgil: That's fine by me, I just copied that line from other filters where a RegExp is used (e.g. https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/filteReddit.js#L519)

morgil: @gamefreak, is there a reason why you used regular expressions instead of string comparisons? In my opinion, this should rather depend on the regexpFilters setting.

andytuba: oh nevermind regexp is great for this, allows people to use wildcards like  Account* covers Account1, Account2, Account3

morgil: I agree, but I still believe that this should depend on whether the user has activated regexpFilters. But this pull request is probably not the right place to discuss this. Maybe I can find some time for it later this week.

andytuba: Eh, seems safe enough without checking that option: reddit usernames can't contain special regexp characters anyway.

morgil: The only problematic location might be where domain names are tested. I don't think there will be a situation where this becomes relevant (if no one starts using iximgur.com or similar on reddit), but it's still unintuitive.

andytuba: Re: domain names -- you mean for the custom filters by domain?  

I should dig in a little deeper before I give a final say, too.

morgil: Yes, as in https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/filteReddit.js#L381
Looking at http://stackoverflow.com/q/2140627, RegExp actually seems to be one of the less ugly methods for case-insensitive string comparison (when escaping control characters). It still breaks for languages such as turkish, where uppercase i is İ and lowercase I is ı - see https://jsfiddle.net/csu3pf9a/1/. Actual language-independent case-insensitive string comparison seems to be impossible.

erikdesjardins:
Merged in #2868.
Thanks!

PR #2611, opened 2016-01-11
set hoverDelay to stated default

aperson:
The description says the default value is 800, when it is actually set to 1000.


KamranMackey:
LGTM :+1:

mc10:
Alternatively, change the description to be 1000, as everyone using RES right now defaults to 1000.

andytuba:
:+1: nix value from the description, default value should be surfaced from the code instead of incorrectly duplicated in description.

aperson:
Done.

andytuba:
Thanks!

PR #2616, opened 2016-01-12
Fix handling of image hrefs - images in albums link to the album

### LINKED ISSUE #2605: Clicking image in album doesn't go to album
https://www.reddit.com/r/Enhancement/comments/40hgpo/clicking_an_inline_image_thats_part_of_an_album/

erikdesjardins:
Fixes #2605

Should be able to cherry-pick this to 4.6.1 without conflict.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -988,7 +988,7 @@ addModule('showImages', function(module, moduleID) {
 
 			var imageAnchor = document.createElement('a');
 			imageAnchor.classList.add('madeVisible');
-			imageAnchor.href = sourceImage.src;
+			imageAnchor.href = sourceImage.href || sourceImage.src;
 			if (module.options.openInNewWindow.value) {
 				imageAnchor.target = '_blank';
 			}


mc10:
This is close (probably close enough), but the link doesn't point to the specific image in the album, which RES does for any other images when you press prev/next. Do you know where the code is for next/prev links?

erikdesjardins:
That's done explicitly by the individual siteModules ([e.g.](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/ac20a876fa7fd36e8ba21ee7472d3f7b3955cea9/lib/modules/hosts/imgur.js#L255)) - so I don't see why it wouldn't work (`addImage()` is called for each image in the album). I'll check again.

gfairchild:
This has been a bug with Vidble albums for months. Will this PR fix it for Vidble albums as well?

erikdesjardins:
Vidble albums have always [intentionally linked to the image itself](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/1adca0c5585505570706a589110db05038f38dfa/lib/modules/hosts/vidble.js#L62), unrelated to this issue. I could fix that separately, I guess.

gfairchild:
It makes sense to fix that. Basically, all albums should function the same, regardless of host. Clicking on the image to take the user directly to the album anchored on the clicked image is the correct behavior.

PR #2618, opened 2016-01-12
RESUtils.forEachChunked dynamic

### LINKED ISSUE #2488: convert forEachChunked to WorstThanHipster's implementation
https://github.com/LiterallyWTH/reddit-moderator-toolbox/commit/80ad3810e1d8bae71218b58aafbce99753284b27

> to keep the framerate at a target framerate, that you can set but i default it to 30.
> No chunk size is needed.  You can set options but it's just array and callback.  It defaults to a chunk size of 10 but ramps up (or slows down) very quickly
> No dependencies, all vanilla, so you can paste it wherever to try it out.

LiterallyWTH:
RESUtils.forEachChunked - Dynamic

Closes #2488

### DIFF
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -2043,7 +2043,7 @@ Of course the items' process times don't matter nearly as much as how hard the b
 */
 RESUtils.forEachChunked = function(array, nope, nuh_uh, process, options){
 	if (typeof process !== 'function') return;
-	var	arr = Array.from(array);
+	var	arr = Array.from(array),
 		start,
 		stop,
 		fr,
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -2026,26 +2026,64 @@ RESUtils.toolTipTimers = {};
 /*
 Iterate through an array in chunks, executing a callback on each element.
 Each chunk is handled asynchronously from the others with a delay betwen each batch.
-If the provided callback returns false iteration will be halted.
-*/
-RESUtils.forEachChunked = function(array, chunkSize, delay, call) {
-	if (typeof array === 'undefined' || array === null) return;
-	if (typeof chunkSize === 'undefined' || chunkSize === null || chunkSize < 1) return;
-	if (typeof delay === 'undefined' || delay === null || delay < 0) return;
-	if (typeof call === 'undefined' || call === null) return;
-	var counter = 0,
-		length = array.length;
+As it stands, nope and nuh_uh are old variables not used for the time being for backwards compatibility.
+This will change the timeout dynamically based on current screen performance in an effort
+to work as fast as possibly without blocking the screen.  There are a lot of tweeks available:
 
-	function doChunk() {
-		for (var end = Math.min(length, counter + chunkSize); counter < end; counter++) {
-			var ret = call(array[counter], counter, array);
-			if (ret === false) return;
-		}
-		if (counter < length) {
-			setTimeout(doChunk, delay);
-		}
+size: doesn't matter much, but if a particularly easy loop is regularly reaching 100's per frame
+you could eeek out a tiny gain by starting at 100's per frame.
+
+framerate: is the important one.  We may want 60, but mostly we want our scrolling and
+mouseclicks to register in a timely manner.  24 is hardly noticable I've found, 
+even as low as 8-16 is tolerable for mouse clicks while the page loads.
+
+nerf: is trickey.  It basically keeps things behaving smoothly.  If you have alist that has items that processes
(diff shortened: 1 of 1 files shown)


lib/core/utils.js
```diff
@@ -2026,26 +2026,57 @@ RESUtils.toolTipTimers = {};
...
+			stop = now();
+			fr = 1000/(stop - start);
+			size = Math.ceil(size*(1 + (fr/framerate - 1)*nerf));
+			return (start = stop);
+		};
+	return new Promise(function(resolve, reject){
+		var doChunk = function(){
+			if (started) optimize();
+			else started = true;
+			arr.splice(0, size).forEach(process);
```
erikdesjardins: I don't see the benefit to having a pre-calculated chunk size and trying to adjust it after - why not keep processing items until you run out of time?
Something like:
```js
const end = now() + frameTime; // frameTime = 1000 / framerate * dutyCycle
for (let i = 0; i < arr.length; ++i) {
	if (now() > end) return nextChunk(arr.slice(i));
	callback(arr[i], i, arr);
}
```

This obviates the need to guess the initial chunk size, allows it to react instantly to changes in browser performance, and guarantees that each chunk will be the optimal size (+/- 1).

This also happens to be the strategy used by FastDOM (might be a FastDOM plugin now), and I don't see any notable disadvantage over the current implementation, so I have reason to believe it's at least marginally better.

We could probably even combine the two strategies and guess if the next item will take too long, i.e.

```js
const start = now();
for (let i = 0; i < arr.length; ++i) {
	if (i && (now() - start) * (1 + 1 / i) > frameTime) return nextChunk(arr.slice(i));
	callback(arr[i], i, arr);
}
```

LiterallyWTH:
The precalculated chunk size is to account for other processes.  If each one runs how long we give it, it's not aware of the other render calls callbacks.  Each one could take the time calculated for a given frame rate.

By pre-calculating chunk size based on time since the last call we naturally account for other the time needed by anything else on the render queue.

erikdesjardins:
Yeah I didn't explain quite correctly - since RAF passes in the start time of the actual frame, we would use that instead of calling `now()` immediately (my examples were incorrect, see the revision history of [this gist](https://gist.github.com/erikdesjardins/3e16d89624d165a5e139)). Which accounts for any other processes, though it'll result in the first chunk per frame taking all of the time and everything else just processing one item.

This is pretty much all solved by queuing, like you suggested - [impl](https://gist.github.com/erikdesjardins/81714dc84b1d4981e184), [with logging](https://gist.github.com/erikdesjardins/0eac883a6dfa5ea26b04).

*Now* I think this has the advantages of our two implementations without the downsides, but maybe you'll see something I don't. It'll process items from each queue (at least one per frame) until we reach `frameTime` after the actual start of the frame (i.e. the fraction of a frame we want to use at the desired framerate). As far as I can gather, RAF callbacks are the last thing to happen before the frame is rendered, so we can get an accurate picture of the framerate by looking at `now() - start`, instead of needing to wait for the next frame.

LiterallyWTH:
Ah, looks great.  The pseudo frame-start-time is returned and passes synchronously which would not take into account outside processes.  It should be set when setTimeout is called:

```javascript
const raf = (callback => { const startTime = now(); setTimeout(() => callback(startTime), 1000 / framerate)});
```

to Babel will polyfill RAF for us right, so for RES it's not as important, and it's outside of our scope of support for TB so we can probably drop it altogether.

You're right about RAF.  RAF callbacks are differed to an animation queue and I believe the API estimates when to pause the normal eventloop queue and call the animation callback queue instead.  Anything else that requires rendering executed

This might help show what's going on:
```javascript
x = y = 0, t = performance.now();

function st(){setTimeout(function(){
	if(x > 25) return
	setTimeout(function(){
		console.log('setTimeout: ',{sto:x++,raf:y,time:(performance.now()- t).toFixed(2) });
	})
	st();
	requestAnimationFrame(function(time){
		console.log('requestAnimationFrame: ',{sto:x,raf:y++,time:(time - t).toFixed(2) });
	});
})}

st();
```

Because they are differed and the frame time comes from the an independent number irrispective of RAF calls I don't think we need the `duty` coefficient anymore;  I think it'll cut off at the right time even if other processes are causing us to drop frames.

erikdesjardins:
> gotten and passed synchronously to the setTimeout callback which would not take into account outside processes

+1, I didn't put much thought into the fallback.

> I don't think we need the duty coefficient anymore

+1, makes sense to me too.

erikdesjardins:
Merged in #2852.

Thanks @LiterallyWTH!

image

LiterallyWTH:
No problem!  This guy was fun.

PR #2645, opened 2016-01-14
Fixed new comment count is still broken (#2139)

kusotool:
Fixed new comment count is still broken in Japanese interface.
function processCommentCounts(ele) was fixed by #2139, but need same fix for function updateCommentCount(mycomment) too.


andytuba:
Cherry-picked into release/4.6.1: eeca4f99919bc2fc96994b281e9386e353dc7f12

andytuba:
Thanks for catching that, @kusotool!

Issue #2637, opened 2016-01-13
Add an option to hide 'subscribe to thread' button


githue:
https://www.reddit.com/r/Enhancement/comments/40r3ku/feature_request_option_to_disable_ability_to/

jkcchan:
I've added a new option showSubscribeButton, with the description: 'Show/hide the comment subscribe button.'

In handleButton(), I check for the option's value, and returns if false.

Let me know if this is not how imagined.

PR #2647, opened 2016-01-15
Show Subscribe Button (#2637)

jkcchan:
(#2637) https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2637

I've added a new option showSubscribeButton, with the description: 'Show/hide the comment subscribe button.'

In handleButton(), I check for the option's value, and returns if false.

Let me know if this is not how imagined.


lib/modules/newCommentCount.js
```diff
@@ -336,6 +341,9 @@ addModule('newCommentCount', function(module, moduleID) {
 	}
 
 	function handleButton(threadid, action) {
+		if (module.options.showSubscribeButton.value == true){
```
13steinj: Equivalency checks to true aren't necessary, since it can only be `true` or `false`, and false wouldn't trigger it. A simple `if (module.options.showSubscribeButton.value)` would do.

Chrome/manifest.json
```diff
@@ -143,6 +143,7 @@
 				"modules/hosts/github.js",
 				"modules/hosts/onedrive.js",
 				"modules/hosts/oddshot.js",
+				"modules/hosts/truncateLongLinks.js",
```
13steinj: This seems to be leftovers from the other PR

lib/modules/newCommentCount.js
13steinj: Also, I don't remember what javascript's default return value is, but I think this would break L170-L179 since I doubt Jquery's `.append` supports it. May want to do the boolean check there (and come to think of it, the check is incorrect, it should be `!module.options.showSubscribeButton.value`, since you want it to not had the button if it's false)

jkcchan:
@13steinj thank you for the comments and patience! Will work on this tomorrow. Really appreciate the help!

lib/modules/newCommentCount.js
```diff
@@ -166,9 +171,13 @@ addModule('newCommentCount', function(module, moduleID) {
 				$thisROW.find('td:first-of-type').append(thisTrash);
 				if (isSubscribed) {
 					$thisROW.find('td:last-of-type').append(thisRenewButton).append(' ');
-					$thisROW.find('td:last-of-type').append(thisUnsubButton);
+					if (!module.options.showSubscribeButton.value){
```
andytuba: This logic seems the opposite of what you want: if the "show button" option is not turned on, add the button? I would have expected "if show button is enabled, add the button", i.e. remove `!`

jkcchan:
@andytuba Sorry for this mistake, I am making the changes now. Thank you for your help!

andytuba:
Shucks, needs rebase.

Issue #2625, opened 2016-01-12
Search Widget doesn't work


daraul:
When searching with the search widget the request goes through fine, returns OK, but nothing is added to the widget itself. If I use the url from the request the search works fine on reddit. 

![2016-01-12_13-44-42](https://cloud.githubusercontent.com/assets/2440975/12271679/b73478e6-b932-11e5-9710-9ed86fe81048.gif)

matheod:
The bug was due to #sitetable no longer here in search page. I fixed it, but now I have an issue where the first search result is not displayed. Working on it.

PR #2659, opened 2016-01-20
Fix search widget #2625

andytuba:
Cherry-pick #2649


andytuba:
@matheod Is this fix ready to ship, despite the subreddit search?

matheod:
Yes, the only problem is that the number of result will not be good, but it's better than nothing working like currently.

PR #2666, opened 2016-01-20
Fix for userTaggerToolTip appearing offscreen

corylulu:
When userTags are edited to when on the very right side of the screen, the tooltip appears partially offscreen. This fix checks for that and moves the tooltip to the left of the userTag instead.


githue:
This will conflict with #2418

corylulu:
@githue Okay, so it's fixed there as well, then problem solved. I submitted this a while back (https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2548), but it looks like you implemented it on your own afterwards when you reworked that code.

PR #2668, opened 2016-01-22
keyboardNav next/prev sibling works like it did in 4.5.4

### LINKED ISSUE #2622: Keyboard nav next sibling always uses "lock to top" scrolling
https://www.reddit.com/r/RESissues/comments/40mftf/bug_keyboard_nav_uses_lock_to_top_when_it_should/

(can also confirm this)

For reference, [keyboardNav.js@4.5.4](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/e83e5e0e01ca21e694922e95e2d6413bf32f9257/lib/modules/keyboardNav.js)

erikdesjardins:
Fixes #2622 and [this issue](https://www.reddit.com/r/RESissues/comments/41ooxs/bug_you_cant_use_the_movedownsibling_keyboard/) (from reddit).

[This](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/e83e5e0e01ca21e694922e95e2d6413bf32f9257/lib/modules/keyboardNav.js#L1602) is the behavior (from 4.5.4) that the the 'legacy' scrollStyle mimics.
It wasn't an option for j/k scrolling in 4.5.4, but I guess why not?

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -951,7 +951,7 @@ addModule('keyboardNav', {
 			}
 		} while (!target && current.length);
 
-		modules['keyboardNav']._moveToThing(target, { scrollStyle: 'top' });
+		modules['keyboardNav']._moveToThing(target, { scrollStyle: 'legacy' });
 	},
 	moveUpSibling: function() {
 		var selected = modules['selectedEntry'].selected();
@@ -964,7 +964,7 @@ addModule('keyboardNav', {
 			target = current.parent().closest('.thing')[0];
 		}
 
-		modules['keyboardNav']._moveToThing(target, { scrollStyle: 'top' });
+		modules['keyboardNav']._moveToThing(target, { scrollStyle: 'legacy' });
 	},
 	moveUpThread: function() {
 		var selected = modules['selectedEntry'].selected();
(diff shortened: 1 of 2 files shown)


andytuba:
The change in behavior is more a bug / oversight on my part than a deliberate change, so we could cheat and call it legacy or just roll it into the default behavior.

erikdesjardins:
Yeah,  the only difference between this and 4.5.4 is an extra `j/k` scrollStyle option called legacy, which behaves like `shift[-alt]-j/k` (`j/k` has never acted this way).
Everything else looks and acts like 4.5.4, as though it was never an issue.

lib/modules/keyboardNav.js
```diff
@@ -938,16 +942,20 @@ addModule('keyboardNav', {
 		if (!selected) return;
 
 		var current = $(selected.thing);
-
 		var target;
-		do {
+
+		if (current.hasClass('link')) {
```
andytuba: https://www.reddit.com/r/RESissues/comments/41ooxs/bug_you_cant_use_the_movedownsibling_keyboard/ right?

erikdesjardins: Yeah.

PR #2671, opened 2016-01-23
Fix showImages on NER pages on /u/me/saved

erikdesjardins:
Fixes [this issue](https://www.reddit.com/r/RESissues/comments/427wrx/bug_when_i_go_into_my_savedreddit_page_res_inline/) (from reddit).

### DIFF
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -448,18 +448,8 @@ RESUtils.currentSubreddit = function(check) {
 	}
 };
 RESUtils.thingsContainer = function(body) {
-	var container;
 	body = body || document.body;
-	container = body.querySelectorAll(RESUtils.thing.prototype.containerSelector);
-	// stupid sponsored links create a second div with ID of sitetable (bad reddit! you should never have 2 IDs with the same name! naughty, naughty reddit!)
-	if (container.length === 2) {
-		// console.log('skipped first sitetable, stupid reddit.');
-		container = container[1];
-	} else {
-		container = container[0];
-	}
-
-	return container;
+	return $(body).find(RESUtils.thing.prototype.containerSelector)[0];
 };
 RESUtils.$things = function(container) {
 	container = container || RESUtils.thingsContainer();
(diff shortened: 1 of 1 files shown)


lib/core/utils.js
```diff
@@ -448,18 +448,8 @@ RESUtils.currentSubreddit = function(check) {
...
-	// stupid sponsored links create a second div with ID of sitetable (bad reddit! you should never have 2 IDs with the same name! naughty, naughty reddit!)
-	if (container.length === 2) {
-		// console.log('skipped first sitetable, stupid reddit.');
-		container = container[1];
-	} else {
-		container = container[0];
-	}
-
-	return container;
+	return $(body).find(RESUtils.thing.prototype.containerSelector)[0];
```
andytuba: `body.querySelector(RESUtils.thing.prototype.containerSelector)` ?

lib/core/utils.js
```diff
@@ -448,18 +448,8 @@ RESUtils.currentSubreddit = function(check) {
 	}
 };
 RESUtils.thingsContainer = function(body) {
-	var container;
 	body = body || document.body;
-	container = body.querySelectorAll(RESUtils.thing.prototype.containerSelector);
-	// stupid sponsored links create a second div with ID of sitetable (bad reddit! you should never have 2 IDs with the same name! naughty, naughty reddit!)
```
andytuba: yay cleaning out cruft  (reddit changed it to `sitetable_organic` after whatever release added this in)

lib/core/utils.js
```diff
@@ -499,7 +489,7 @@ RESUtils.thing = function(element) {
 RESUtils.thing.prototype = {
 	thingSelector: '.thing, .search-result-link',
 	entrySelector: '.entry, .search-result-link > :not(.thumbnail)',
-	containerSelector: '.sitetable, .search-result-listing',
+	containerSelector: '.sitetable, .search-result-listing:last',
```
andytuba: may need to do `:last-of-type` to support https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2671/files#r50610107

lib/core/utils.js
erikdesjardins: `:last` is a special jQuery selector

erikdesjardins: Er, nothing else uses this does it?
I don't think we can really do this with vanilla selectors, unless we do `selected[selected.length - 1]`

lib/core/utils.js
erikdesjardins: That'll select nothing, because `.search-result-listing` isn't the last div

lib/core/utils.js
andytuba: Yeah, this'll be fine, it doesn't need to be barebones-fast.

PR #2676, opened 2016-01-24
Fix comments passing through sidebar.

githue:
Like it was in 4.5.4

Fixes https://www.reddit.com/r/RESissues/comments/42exc9/bug_comments_cover_sidebar/


githue:
@andytuba, I think this can be safely added to /r/resupdates for 4.6.0 only.

Issue #2522, opened 2015-11-16
Closing streamable expando does not pause


andytuba:
https://www.reddit.com/r/RESissues/comments/3sv7j7/bug_collapse_of_streamablecom_videos_just_hides/

tswaters:
Inside the streamable iframe, there is a listener that accepts play/pause requests - this raises an error `videojs is not defined`.  If this was changed to `player` it appears to work.  This is on streamable to fix; RES can't modify an iframe... it's already doing everything it can posting `{method: "pause"}` to the iframe.... streamable appears to recognize this but instead of pausing, it raises an error.

andytuba:
Thanks for investigating that. I'll reach out to streamable support

andytuba:
Appears to be fixed on Streamable's side.

PR #2679, opened 2016-01-25
Streamable is handled natively in reddit for submission titles.

### LINKED ISSUE #2678: Streamable videos do not display again after having collapsed expando
As mentioned in the title streamable videos will not display again after having collapsed its expando once. This is most apparent on [/r/soccer](http://reddit.com/r/soccer) where most of the video links are from streamable

ezgif com-video-to-gif 1

githue:
Fixes #2678

Copied from vimeo.js

However #2522 still applies to streamable videos in comments.

### DIFF
--- a/lib/modules/hosts/streamable.js
+++ b/lib/modules/hosts/streamable.js
@@ -1,7 +1,11 @@
 addLibrary('mediaHosts', 'streamable', {
 	domains: ['streamable.com'],
 	detect: function(href, elem) {
-		return href.indexOf('streamable.com') !== -1;
+		// Only find comments, not the titles.
+		if (href.indexOf('streamable.com') !== -1) {
+			if (elem.className.indexOf('title') === -1) return true;
+		}
+		return false;
 	},
 	handleLink: function(elem) {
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/streamable.js
```diff
@@ -1,7 +1,11 @@
 addLibrary('mediaHosts', 'streamable', {
 	domains: ['streamable.com'],
 	detect: function(href, elem) {
-		return href.indexOf('streamable.com') !== -1;
+		// Only find comments, not the titles.
+		if (href.indexOf('streamable.com') !== -1) {
+			if (elem.className.indexOf('title') === -1) return true;
```
andytuba: `if (elem.classList.has('title'))`

erikdesjardins: `if (!elem.classList.contains('title'))`

andytuba: i might have set up an autocorrect on my laptop because I keep messing that up.

githue:
Ok, updated to use classList.contains.

PR #2681, opened 2016-01-26
Created new module to modify existing links on pages

mldangelo:
Hi, 
I often click wikipedia links and am directed to the mobile site instead of the desktop site. After reading some complaints from redditors about mobile wikipedia links (some people even use things like [this](https://chrome.google.com/webstore/detail/mobile2desktop-wikipedia/hgdfjjhaahfdggnecdecjekkimepkpaf/reviews?hl=en)), I thought I would build this module. More sites can be added to options.domains in the following format:

```javascript
{
  domain: 'wikipedia.org',
  match: /(\b.m.wikipedia.)\w+/,
  find: '.m.',
  replace: '.'
},
```

The domain field is unused, but may be useful for reference purposes. Regex can be used in the find, replace, and match fields. I didn't document my code because no other modules seem to be documented. I'm happy to go back and do it, or work on other features if they would be welcomed. Thank you for your consideration.

### DIFF
--- a/Chrome/manifest.json
+++ b/Chrome/manifest.json
@@ -50,6 +50,7 @@
 				"modules/keyboardNav.js",
 				"modules/about.js",
 				"modules/commandLine.js",
+				"modules/modifyLinks.js",
 				"modules/messageMenu.js",
 				"modules/easterEgg.js",
 				"modules/pageNavigator.js",
--- a/Firefox/index.js
+++ b/Firefox/index.js
@@ -173,6 +173,7 @@ pageMod.PageMod({
 		self.data.url('modules/userTagger.js'),
 		self.data.url('modules/keyboardNav.js'),
 		self.data.url('modules/commandLine.js'),
+		self.data.url('modules/modifyLinks.js'),
 		self.data.url('modules/messageMenu.js'),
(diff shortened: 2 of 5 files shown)


erikdesjardins:
While we appreciate your contribution, we don't feel that this is something RES should be responsible for: it isn't a reddit-specific problem, for one, and it's already solved by [numerous](https://chrome.google.com/webstore/detail/wiki-redirect/oenahiaeppiepibbmfpadomieekgbmbb?hl=en-US) [other](https://chrome.google.com/webstore/detail/mobile2desktop-wikipedia/hgdfjjhaahfdggnecdecjekkimepkpaf?hl=en) [extensions](https://chrome.google.com/webstore/detail/redirector/pajiegeliagebegjdhebejdlknciafen) - the latter being just as generic as this module.

Hell, they do a better job than RES ever could, since they work everywhere on the web.

andytuba:
Since this sort of thing comes up fairly often, maybe this could pivot to from "Modify Links" to "Suggested Pairings": when user clicks a link matching a mobile URL, show a sticky notification recommending a better experience if they install such-and-such modules.

erikdesjardins:

PR #2682, opened 2016-01-26
Fix: [l=c] link opens two tabs when comment page's URL contains non-ascii character

kusotool:
Because linkURL is decoded but commentsURL is not.


andytuba:
@kusotool Thanks for raising this issue. I tweaked singleClick in a slightly different way which should sidestep the need to decodeURI.

PR #2684, opened 2016-01-27
Async overhaul

erikdesjardins:
Alt title: "all hacks must die".

Fixes #2279, fixes #2291, fixes #2497, fixes #2533, fixes #2672, closes #2536, closes #2664.
Depends on #2454, #2507, #2575.
Depends on Babel issues: ~~[T2776](https://phabricator.babeljs.io/T2776) (fixed, needs to be merged and released), [T6882](https://phabricator.babeljs.io/T6882) (I'll look into contributing a fix for this soon).~~
~~Due to these issues, it won't work with the current set of transformers - if you want to build this branch, you'll need to replace your `.babelrc` with [this](https://gist.github.com/erikdesjardins/94843e8aa51e00a427ae), which will only run in browsers with generator support.~~
~~Both fixed and released.~~
~~Not quite. [T3077](https://phabricator.babeljs.io/T3077) / [T6972](https://phabricator.babeljs.io/T6972)~~
Babel 6.6.0 is out, time is go

Contains major rewrites/refactoring for all backends, RESEnvironment, init, and media hosts.

In short: async storage (minor performance improvements), all caching moved into memory, large performance improvements for init, backends are much cleaner, safer, and easier to debug (a *ton* of hacks were removed, all browsers now present an identical messaging API with proper responses).

Many more details in the commit messages.

PR #2688, opened 2016-01-27
Fix YouTube video times

### LINKED ISSUE #376: Use limiting parameters for youtube feed api
In `getVideoJSON`, the json url currently used is `http://gdata.youtube.com/feeds/api/videos?q='+thisIDString+'&v=2&fields=entry(id,title,media:group(yt:duration,yt:videoid,yt:uploaded))&alt=json`. Because this fetches (up to 20?) videos that have ID's from the ID string anywhere in them (ID, title, description, etc/), a full feed is fetched and then iterated over after fetched to see if the ID of a video in the feed was actually an ID requested (At least this is how I interpreted it, however I could be completely incorrect).

In basic testing (logging the requested vs returned video ID's from several pages on /r/videos using both the old and new URL's), using `http://gdata.youtube.com/feeds/api/videos?q='+thisIDString+'&v=2&fields=entry(id,title,media:group(yt:duration,yt:videoid,yt:uploaded))&orderby=relevance&alt=json&max-results='+thisBatch.length` seemed to be just as accurate as the current URL in ensuring that the appropriate videos are in the feed, but this has the bonus of eliminating the number of times `modules['betteReddit'].youtubeLinkRefs` is iterated through.

If this seems like an unnecessary change, which I understand, I also noticed another oddity. In looking at this method, why is each item in the feed is processed before it checks whether it was really even a requested video (since the feed contains more videos than requested due to the nature of the 'q' parameter)? Why not move the

    ...
    for (var j = 0, lenj = modules['betteReddit'].youtubeLinkRefs.length; j < lenj; j += 1){
        if (modules['betteReddit'].youtubeLinkRefs[j][0] === thisYTID) {
    ...

 line to before any of the feed entry item processing begins?

### LINKED ISSUE #2234: YouTube video time not showing?
I think YouTube changed their API last week and broke this.  

https://www.reddit.com/r/RESissues/comments/3608v8/time_length_of_videos_no_longer_show_on_rvideos/

erikdesjardins:
Depends on #2684; [diff](https://github.com/erikdesjardins/Reddit-Enhancement-Suite/compare/promise-to-async...fix-videoTimes).

Fixes #2234, closes #376.

The quota will be about 10 million requests per day (one request = up to 50 videos, so 10 million listing pageloads or so), and it's enabled by default...not sure if that'll be enough. I really don't know how often the average user browses reddit. If in doubt, we could just migrate it to off.

Although if we leave it on, funnily enough, we'll probably get better usage statistics than the Chrome webstore gives us...

### DIFF
--- a/lib/modules/betteReddit.js
+++ b/lib/modules/betteReddit.js
@@ -262,61 +262,60 @@ addModule('betteReddit', (module, moduleID) => {
 
 				if (!match) return;
 
-				// add quotes so URL creation is doable with just a join...
-				const thisYTID = `"${match[1]}"`;
-
 				const timeMatch = getYoutubeStartTimeRegex.exec(link.href);
 				const titleMatch = titleHasTimeRegex.test(link.textContent);
 				if (timeMatch && !titleMatch) {
 					link.textContent += ` (@${timeMatch[1]})`;
 				}
 
-				const { time, title } = await getVideoInfo(thisYTID);
+				const { info, title } = await getVideoInfo(match[1]);
 
-				link.textContent += ` ${time}`;
+				link.textContent += ` - ${info}`;
 				link.setAttribute('title', `YouTube title: ${title}`);
 			});
 	}
 
-	const getVideoInfo = RESUtils.batch(async thisBatch => {
-		const data = await RESEnvironment.ajax({
-			url: 'http://gdata.youtube.com/feeds/api/videos',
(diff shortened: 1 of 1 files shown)


hillai:
So glad to see that someone fixed the issue. Good work! Hoping this will get merged.

erikdesjardins:
It will. But it won't be in the next (patch) release. It'll probably be in the next minor version release (4.7).

PR #2689, opened 2016-01-28
Use indexedDB in Firefox

### LINKED ISSUE #1893: Firefox simple-storage keeps getting trashed
also it has a small quota

erikdesjardins:
Depends on #2684; [diff](https://github.com/erikdesjardins/Reddit-Enhancement-Suite/compare/promise-to-async...native-storage).

Fixes #1893, closes #1840.

### DIFF
--- a/Chrome/background.js
+++ b/Chrome/background.js
@@ -274,6 +274,8 @@ addListener('storage', ([operation, key, value]) => {
 			return key in localStorage;
 		case 'keys':
 			return Object.keys(localStorage);
+		case 'clear':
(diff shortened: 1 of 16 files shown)


erikdesjardins:
One thing of note: this adds a new `storage` permission, so Chrome will disable RES for everyone. The message isn't *quite* as scary as it is for domain permissions, though.

Alternatively, I'd be okay with continuing to use `localStorage` in Chrome or making it an optional permission, although both would be less performant (and the latter, ugly).

PR #2690, opened 2016-01-28
Remove deprecated event dispatcher code

LordWingZero:
Event.initEvent() has been removed from the Web standards. Though some browsers may still support it, it is in the process of being dropped. Do not use it in old or new projects. Pages or Web apps using it may break at any time.

Many methods used with createEvent, such as initCustomEvent, are deprecated. Use event constructors instead.


andytuba:
merged to 4.6.1

PR #2707, opened 2016-02-04
pageNavigator: Change alt attribute to title

ljcl:
as per https://www.w3.org/TR/html5/dom.html#the-title-attribute


erikdesjardins:
Thanks!

Issue #2458, opened 2015-09-24
observe clippy option for tooltip on video embeds


andytuba:
https://www.reddit.com/r/RESissues/comments/3m7yc7/bug_clippy_is_not_disabling_tooltips_on_gifv_and/

DrLeh:
It seems like clippy isn't working at all right now.

If I'm not mistaken, I believe the `&&` should really be an `||` on this line: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/e6238bd47e9c4b656b3d563978a4fbb82657c7ee/lib/modules/showImages.js#L1781 

This would make it skip clippy text if either clippy is set to false or the title has already been applied.

erikdesjardins:
#2713

PR #2713, opened 2016-02-08
Fixed clippy to correctly be disabled.

DrLeh:
Was doing `&&` when it should've been `||`.

For example, run the following code in the console to see the difference.

```
var module={options:{clippy:{value:false}}};
elem = {title:''}
if (!module.options.clippy.value && elem.title) console.log('skipped 1');
if (!module.options.clippy.value || elem.title) console.log('skipped 2');
>> skipped 2
```

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1778,7 +1778,7 @@ addModule('showImages', function(module, moduleID) {
 	}
 
 	function setMediaClippyText(elem) {
-		if (!module.options.clippy.value && elem.title) return;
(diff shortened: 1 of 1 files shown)


andytuba:
Closes #2458 

Thanks for digging into that!

andytuba:
TODO @andytuba: cherry-pick into release/4.6.1

DrLeh:
This got left out of 4.6.1, any reason?

andytuba:
Oversight, sorry :sob: I should have made a ticket targeting 4.6.1. Let's definitely get it into master soon, at least.

PR #2730, opened 2016-02-22
External stylesheets

andytuba:
Expand loadSubredditStylesheets into also loading external urls. Deprecate r/resupdates.

### DIFF
--- a/lib/modules/stylesheet.js
+++ b/lib/modules/stylesheet.js
@@ -335,6 +335,8 @@ addModule('stylesheet', function(module, moduleID) {
 		var subreddit = RESUtils.currentSubreddit();
 		if (!subreddit) return true;
 
+		subreddit = subreddit.toLowerCase();
+		applyList = applyList.toLowerCase().split(',');
 		var all = (applyList && applyList.indexOf('all') !== -1);
 		switch (applyTo) {
 			case 'exclude':
--- a/lib/modules/stylesheet.js
+++ b/lib/modules/stylesheet.js
@@ -229,7 +229,7 @@ addModule('stylesheet', function(module, moduleID) {
 		return rows.map(function(row) {
 			var url = row[0];
 			if (subredditNameRegexp.test(url)) {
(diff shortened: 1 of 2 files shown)


lib/modules/stylesheet.js
```diff
@@ -17,14 +17,15 @@ addModule('stylesheet', function(module, moduleID) {
 		}
 	};
 
-	module.options.loadSubredditStylesheets = {
+	module.updatesStylesheet = 'https://cdn.redditenhancementsuite.com/updates.css';
```
erikdesjardins: `var updatesStylesheet`?

lib/modules/stylesheet.js
```diff
@@ -222,46 +223,49 @@ addModule('stylesheet', function(module, moduleID) {
...
-
-		var subreddits = RESUtils.options.listTypes['subreddits'].sanitizeValues(uncleanValues);
-		return subreddits;
+	var subredditNameRegexp = /^(?:\/?r\/)?([\w_]+)\/?$/;
+	var urlRegexp = /^(?:https?:\/\/[\w\.]+)?\/\w+/;
+	function sanitizeStylesheetUrls(rows) {
+		return rows.map(function(row) {
+			var url = row[0];
+			if (subredditNameRegexp.test(url)) {
+				return document.location.origin + '/r/' + subredditNameRegexp.exec(url)[1] + '/stylesheet.css';
```
erikdesjardins: `location.origin` probably isn't necessary here

lib/modules/stylesheet.js
andytuba: Public so migrate can use it.

lib/modules/stylesheet.js
andytuba: Oh, yeah, I incorrectly thought about referencing assets inside the extension -- but it's nbd because injected link

lib/modules/stylesheet.js
erikdesjardins: Ah, didn't notice that

PR #2742, opened 2016-02-28
Fix #2725 - highlightFirstCommenter doesn't when "show more comments" in children/replies

### LINKED ISSUE #2725: highlightFirstCommenter doesn't when "show more comments" in children/replies
https://www.reddit.com/r/AskReddit/comments/4693qi/what_are_some_early_warning_signs_of_trouble_that/d03a5of?context=3&limit=1

http://i.imgur.com/gLfdUDX.gifv

brandonpatrick:
The firstComments object needed to be cleared so that new comments could be checked. I added the highlightedFirstComments object to avoid adding duplicate highlight styles.

### DIFF
--- a/lib/modules/userHighlight.js
+++ b/lib/modules/userHighlight.js
@@ -228,9 +228,6 @@ addModule('userHighlight', function(module, moduleID) {
 
 			if (authorClass === undefined) return;
 
-			var authorDidReply = element.querySelector('.child .' + authorClass);
-			if (!authorDidReply) return;
-
 			var container = '.' + idClass;
 			if (module.options.dontHighlightFirstComment.value) {
 				container += ' .child';


erikdesjardins:
Cool!

It looks like this could also be fixed by removing the `authorDidReply` check on [L231/232](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2742/files#diff-57fbaa88344218f13ca331f6808032efL231). Could you try that? If it works, I'd prefer to do it that way, for simplicity.

Feel free to force-push to this branch, if it does.

brandonpatrick:
Done.

On line 201 is it still necessary to add the newComments watcher? When I remove this line, the functionality still appears to work.
`RESUtils.watchForElement('newComments', scanPageForFirstComments);`
It looks like the watcher calls 'scanPageForFirstComments' once for every new child comment loaded, but that shouldn't be necessary if the highlighting style has already been added for the thread.

erikdesjardins:
It's probably still necessary for loading new top-level comments at the bottom of the page.

brandonpatrick:
@erikdesjardins Okay, I thought there might be something like that. The branch is good to go, then

erikdesjardins:
Thanks!

PR #2744, opened 2016-02-29
Announcement notifications

andytuba:
Show a notification for new /r/RESAnnouncements post once every several days, if storage is working

Don't show ! or notification for older posts

Red ! shows all the time for "recent posts" even if storage isn't working

Regression-checked ! with pizzazz


erikdesjardins:
LGTM

PR #2753, opened 2016-03-05
Expandos lose 'collapsed' class when expanded

githue:
On the first click of an expando the 'collapsed' class doesn't get removed, which is evident if you're using Reddit's new expando CSS.

The fix is on line 172, the other changes are code style.


erikdesjardins:
This is the 4.6.1 counterpart of #2752

Issue #2716, opened 2016-02-10
strawpoll.me expandos


erikdesjardins:
https://www.reddit.com/r/Enhancement/comments/453bkw/feature_request_embed_strawpolls/

erikdesjardins:
#2754

PR #2754, opened 2016-03-05
Add strawpoll.me media host.

thybag:
Hello,

Have quickly added a stawpoll.me media host, as requested in https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2716

https://www.reddit.com/domain/strawpoll.me/
(Tested with both http/https with/without www & direct links to the embed)

Since quite a few of the strawpoll embeds are going to scroll (Can't think of a way to automatically work out a poll's height), I also added a `maxwidth` option to the standard iframe expando. (Since I can then ensure the scrollbar is directly alongside the poll). Can easily pull this bit back out if wanted though.

Thanks,
Carl


lib/modules/hosts/strawpoll.js
```diff
@@ -0,0 +1,11 @@
+addLibrary('mediaHosts', 'strawpoll', {
+	domains: ['strawpoll.me'],
+	detect: href => (/^https?:\/\/(?:www\.)?strawpoll\.me\/(embed_[0-9]\/)?([0-9]*)/i).exec(href),
+	handleLink(elem, [, , uid]) {
```
erikdesjardins: If you make `(embed_[0-9]\/)` a non-capturing group (`(?:embed_[0-9]\/)`), then this can just be `[, uid]`

lib/modules/hosts/strawpoll.js
```diff
@@ -0,0 +1,11 @@
+addLibrary('mediaHosts', 'strawpoll', {
+	domains: ['strawpoll.me'],
```
erikdesjardins: Can add `attribution: false,` here (Strawpoll adds its own "Powered by strawpoll.me" link at the bottom)

lib/modules/hosts/strawpoll.js
```diff
@@ -0,0 +1,11 @@
...
+	detect: href => (/^https?:\/\/(?:www\.)?strawpoll\.me\/(?:embed_[0-9]\/)?([0-9]*)/i).exec(href),
+	handleLink(elem, [, uid]) {
+		elem.type = 'IFRAME';
```
erikdesjardins: I think it would make sense to add `elem.expandoClass = 'selftext';` here, so it uses the text expando icon

Issue #388, opened 2013-04-29
Functionality to rotate images in imageviewer


andytuba:
http://www.reddit.com/r/Enhancement/comments/1davwc/feature_request_ability_to_rotate_images/

e.g. for images which were uploaded but not set to the correct rotation.

erikdesjardins:
#2755

Issue #834, opened 2014-03-18
imageViewer GUI controls


andytuba:
The image viewer's controls are difficult to discover.  The single discoverability item, "click and drag to resize" tooltip, is a little annoying.

In the interest of discoverability and mouse control, I'd like to see a floating toolbar that appears when hovering on a draggable image/video, with +/- and < ^ v > controls and ? to educate the user on click-drag functionality.

larsjohnsen:
Resolved by #3316, #2755

PR #2755, opened 2016-03-06
Media control bar with ability to rotate images

thybag:
Hello,

Was bored today so decided to have a crack at implementing https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/388 .

The current implementation adds a small bar (shown on hover), which gives people to ability to roate images in 90 degree's increments left/right. It also plays nice with the drag & resize stuff.

Assuming people are okay with how it works, I suspect it would be pretty straight forward to extend the bar to implement other GUI controls to, such as those suggested in https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/834 (Stuff like zoom in/out, flip, help options etc.)

~~Main limitation currently is that I've excluded the controls on gifv/videos, the actual rotation logic itself all works fine, I've just had trouble stopping them from getting cut off. (This is probably solvable  if it put a little more time in to it).~~ Should now play nice with gifv's too :)

Just wondering what peoples thoughts are? is this worth continuing or am i going in a bit of a daft direction with it.

**Examples**

Control icons show when user hovers on image
controls-hover

Image maintains size when flipped
controls-rotate


lib/modules/showImages.js
```diff
@@ -1744,6 +1749,54 @@ addModule('showImages', function(module, moduleID) {
...
+			controls.appendChild(tmp);
+		}
+
+		// Rotation controls are only for images (as videos will end up with their controls the wrong way up)
+		var media = $(mediaTag);
+		media.data('media-controls', controls);
+		$(mediaTag).parent().append(controls);
+
+		// Hook up controls
+		$(controls).find('span').click(function(e){
```
andytuba: alternatively

```javascript
$(controls).on('span[data-action]', 'click', function(e) {
    var $control = $(e.delegateTarget);
    ...

});
```

lib/modules/showImages.js
```diff
@@ -1744,6 +1749,54 @@ addModule('showImages', function(module, moduleID) {
...
+		if (!module.options.imageControls.value) return;
+
+		// Controls to show
+		var control_buttons = {
+			'rotateLeft': 'Rotate image counter-clockwise',
+			'rotateRight': 'Rotate image clockwise'
+		};
+
+		// generate controls bar
+		var controls = document.createElement('div');
```
andytuba: could also do this in template easy enough and IMO a little prettier

lib/css/modules/_showImages.scss
```diff
@@ -4,3 +4,49 @@ body:not(.res-showImages-displayImageCaptions) {
...
+	background: #fff;
+	z-index: 2;
+	bottom: 10px;
+	right: 10px;
+	border: solid 1px #eee;
+
+	span {
+		padding: 8px;
+		display: inline-block;
+		font: normal 14px/1 "Batch";
```
andytuba: How about putting `<span class="res-icon">` buttons into the control so you don't have to duplicate the icon setup css?

andytuba:
This is pretty neat and would be a great place to move the clippy text, also a visible  thumb to grab for resizing.

lib/css/modules/_showImages.scss
```diff
@@ -4,3 +4,49 @@ body:not(.res-showImages-displayImageCaptions) {
 		display: none;
 	}
 }
+
+.madeVisible {
```
githue: You probably want div.madeVisible or a.madeVisible depending on which one you're targeting.

thybag:
Thanks for all the feedback. Will have a go at getting it all cleaned up a bit when i next get a free evening.

In terms of additional buttons,  what were you guys thinking for the "clippy text" - Just something like a ? icon, which pops up a window with some instructions?

Also, is it worth me adding a zoom in & out button to the bar too? or do you think thats overkill.

Other thought i had was adding an option to settings so you could choose which corner the menu was positioned in.

Thanks,
Carl

githue:
I think most people would settle on the toolbar being in the top left so it doesn't move every time they rotate or resize. That keeps the bottom left available for a separate button for closing the expando (so you don't have to scroll up again), it's something that has been requested before.

As for the "drag to resize" hint (clippy text?), you could copy Chrome's fullscreen behavior where it overlays a message that fades away after a second (1.5s full opacity, 0.5s fade out?).

There could also be a cog icon that takes you directly to the relevant settings, but style it the same as the other icons in the name of uniformity.

lib/modules/showImages.js
thybag: Hmm, just gave this a try, but it doesn't seem like the delegated events are able to stop the click propagating to the anchor wrapping em.

andytuba:
Potential optional add-on tool idea: https://www.reddit.com/r/Enhancement/comments/49ji61/feature_request_reverse_image_search_button/

thybag:
Oo, i like the reverse image search button idea. Any idea what icon to use?

thybag:
For google looks like i can just link straight to https://images.google.com/searchbyimage?image_url=*url* so should be a pretty simple one to add :)

andytuba:
>  reverse image search button ... icon?

:mag: ? batch uniF094

I don't think batch has a "find image" or "find document" icon. The buttons might be too small for something that detailed, though.

lib/modules/showImages.js
```diff
@@ -1666,6 +1689,51 @@ addModule('showImages', (module, moduleID) => {
...
+					break;
+				case 'rotateRight':
+					module.rotateMedia(mediaTag, ($media.data('rotation-state') || 0) + 1);
+					break;
+				case 'imageLookup':			
+					let lookupUrl = ($media[0].tagName==='IMG') ? $media.attr("src") : $media.parent().attr('href');
+
+					// Google doesn't like image url's without a protacol
+					if (lookupUrl.indexOf('//') === 0){
+						lookupUrl = document.location.protocol + lookupUrl;
```
erikdesjardins: `lookupUrl = new URL(lookupUrl, location.href).href` (without the if statement), will resolve any relative URL, and is a bit more bulletproof. ([docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#Parameters))

lib/modules/showImages.js
```diff
@@ -1666,6 +1689,51 @@ addModule('showImages', (module, moduleID) => {
...
+					}
+
+					RESEnvironment.openNewTab(`https://images.google.com/searchbyimage?image_url=${lookupUrl}`, true);
```
erikdesjardins: Minor nit: this will focus the tab by default, so the `true` argument can be omitted

lib/modules/showImages.js
```diff
@@ -1902,6 +1994,38 @@ addModule('showImages', (module, moduleID) => {
 		}
 	};
 
+	module.rotateMedia = function(ele, rotationState) {
```
erikdesjardins: If this isn't used outside the module, I'd prefer to have a bare `function rotateMedia(ele, rotationState) {`.
We can add `module.rotateMedia = rotateMedia;` in the future if we need to expose it (or hopefully I'll be done converting to ES6 modules by then).

thybag:
Any blockers on this if i merge it all up to date? (will probs have to tweak UI to work with new video player)

erikdesjardins:
No, I think it'll be good to land once updated. (@andytuba?)

erikdesjardins:
Fixes #388 (for autoclose)

thybag:
All merged up, but looks like it goes a bit wrong with the new video player for gifv's etc. Gonna see if i can fix

PR #2756, opened 2016-03-06
Bugfix account for swicther "arrow" dropdown

thybag:
Tiny bug fix for the account switcher dropdown.

 When the display style on the account switcher is set to "simple arrow", the arrow icon is misaligned. This fix just switches the arrows vertical align back to middle which fixes it for me.


erikdesjardins:
LGTM

PR #2758, opened 2016-03-11
Body class for User Highlighter

githue:
Allows themes to customize User Highlighter without affecting RES users who have the module disabled. Since this doesn't change CSS specificity it shouldn't cause any visible changes; it's intended to only be discoverable by themers.

This change opens up the possibility for usernames outside of `.thing` to be affected by User Highlighter. However since it only affects specific elements with classes such as `.author.submitter` and `.author.moderator` it shouldn't be a problem.

Tested in /r/naut, modmail, inbox and vanilla reddit.

Addresses these reports:

* https://redd.it/49nwty
* https://redd.it/42dqz0

### DIFF
--- a/lib/modules/userHighlight.js
+++ b/lib/modules/userHighlight.js
@@ -201,6 +201,8 @@ addModule('userHighlight', function(module, moduleID) {
 				RESUtils.watchForElement('newComments', scanPageForFirstComments);
 				scanPageForFirstComments();
 			}
+
+			RESUtils.bodyClasses.add('res-userHighlight');
 		}
 	};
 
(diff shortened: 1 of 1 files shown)


andytuba:
Would it be overkill to extend this to every module if enabled and on match url?  Maybe provide an opt-out for certain by passing a `context=bodyClass` parameter to `isMatchUrl`?

githue:
Module.bodyclass boolean makes sense, I just assumed it was already availabe.

andytuba:
Looks like not yet, but https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/1adca0c5585505570706a589110db05038f38dfa/lib/core/utils.js#L30 would be a perfect place to add

```
if (this.bodyClass) { 
    RESUtils.bodyClasses.add(`res-${this.moduleID}`); 
}
```

PR #2664, opened 2016-01-20
Fix #2497 - Opening a link (consisting of only a hash) in a new tab opens the background page

denizdogan:
Fixes #2497 

This is a copy of [this pull request](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2526) but up-to-date with the current code base. Sorry for this, Git just wouldn't cooperate!

PR #2764, opened 2016-03-12
Fix keyboard nav regression

13steinj:
If keyboard nav settings are set to open links in the same tab, it wouldn't work because this was broken sometime down the line.


andytuba:
re: #2664

PR #2777, opened 2016-03-14
niceDateDiff: Fix edge case for day diff

mc10:
I added the Oxford comma as well; hope nobody minds.

### DIFF
--- a/lib/core/utils.js
+++ b/lib/core/utils.js
@@ -1413,32 +1413,17 @@ RESUtils.niceDateTime = function(d, usformat) {
 	return `${dateString} ${hours}:${minutes}:${seconds}`;
 };
 RESUtils.niceDateDiff = function(origdate, newdate) {
-	// Enter the month, day, and year below you want to use as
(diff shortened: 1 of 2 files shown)


lib/core/utils.js
```diff
@@ -1486,33 +1486,18 @@ RESUtils.niceDateTime = function(d, usformat) {
 	return `${dateString} ${hours}:${minutes}:${seconds}`;
 };
 RESUtils.niceDateDiff = function(origdate, newdate) {
-	// Enter the month, day, and year below you want to use as
-	// the starting point for the date calculation
 	if (!newdate) {
 		newdate = new Date();
```
erikdesjardins: Might as well make this a default param while you're at it, I guess.
`newdate` is always a Date object or undefined, so no usages should have to be changed.

erikdesjardins:
Half the length and more accurate? LGTM!

I don't feel strongly about the Oxford comma either way.

PR #2788, opened 2016-03-17
Use more generic "*" for permissions.

ZakariaRidouh:
This PR makes it so that RES can continue to work if some of the sites move to https or if someone posts a link that is http (not https) by using the `*` instead of just http or https.
Thanks!

### DIFF
--- a/chrome/manifest.json
+++ b/chrome/manifest.json
@@ -15,11 +15,9 @@
 	"content_scripts": [
 		{
 			"matches": [
-				"http://redditenhancementsuite.com/*",
-				"http://reddit.com/*",
-				"https://reddit.com/*",
-				"http://*.reddit.com/*",
-				"https://*.reddit.com/*"
+				"*://redditenhancementsuite.com/*",
+				"*://*.reddit.com/*",
+				"*://reddit.com/*",
 			],
 			"js": [
 				"vendor/polyfill.min.js",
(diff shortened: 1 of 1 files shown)


honestbleeps:
I could be wrong, but I'm pretty certain this sort of change would disable RES in Chrome for all users because of a change to permissions.

It sucks we're pigeonholed where we are rather than the smarter/better "*", but I'm pretty sure that merging this PR could result in a tech support disaster. 

I can't think off the top of my head how to test this yet, I suppose maybe with a dummy extension of some sort, but Google doesn't take too kindly to uploading bogus extensions to the webstore. Anyone got any ideas?

andytuba:
I appreciate your contribution, @ZakariaRidouh!  This might be useful in the future, but I think it's not yet necessary to merge and release with this.  I expect someone on the dev team or in the community will see a notification about any APIs that are moving to https-only in time to merge this.  (That potential situation might be a good opportunity to tweak some other permissions, too.)

Fortunately, RES's [media host modules](https://github.com/honestbleeps/Reddit-Enhancement-Suite/tree/master/lib/modules/hosts) are already pretty resilient to http vs https, and [siteModules typically use a consistent API endpoint](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/hosts/flickr.js#L7).

(That said, I do have a private extension we could test on.)

ZakariaRidouh:
@andytuba good point, but do you know is just changing the protocol (not the actual web URL) will still deactivate the extension? 

Also, it would be a great idea to test this on your dummy extension :smile: )

andytuba:
Yeah, it's worth giving it a small test.

andytuba:
@erikdesjardins found some very relevant documentation: https://developer.chrome.com/extensions/permission_warnings  (but I haven't read through it yet)

erikdesjardins:

Also, I don't expect Chrome will ever relax their restrictions on changing existing permissions, so unfortunately I doubt we'd have been able to accept this.

PR #2793, opened 2016-03-17
fix NER event name: neverEndingLoaded -> neverEndingLoad

### LINKED ISSUE #2792: Never ending load event dispatch was renamed
I don't think that this was on purpose. I'll change my script to listen for both event names...

https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/d4a9ea732ad40adc424467360f456a8e1403971e#diff-b5b236fa347169a5226fe818517af203

erikdesjardins:
Fixes #2792.

### DIFF
--- a/lib/modules/neverEndingReddit.js
+++ b/lib/modules/neverEndingReddit.js
@@ -448,7 +448,7 @@ addModule('neverEndingReddit', (module, moduleID) => {
 					NERFail(noresultsfound);
 				}
 
-				window.dispatchEvent(new Event('neverEndingLoaded', { bubbles: true, cancelable: true }));
+				window.dispatchEvent(new Event('neverEndingLoad', { bubbles: true, cancelable: true }));
 			}
 		}
 	}


mc10:
This should also probably get merged into the release branch.

caleywoods:
What's the status on this? I was about to open an issue for ``neverEndingLoad`` having stopped working but found someone in an old thread on /r/Enhancement mentioning tried neverEndingLoaded and I got that function to work.

erikdesjardins:
4.7 release should be some time in the next few months. Can't give you anything more concrete than that, sorry.

andytuba:
I guess NER should emit both events now.

caleywoods:
A replace all should fix anyone's code once this reverts back - don't see any reason to have it emitting both events.

andytuba:
neverEndingLoaded is being emitted out  in the wild, right, so somebody has probably patched their code to support the new name. If RES goes back to the old one, that fixes older broken scripts but breaks newer fixed scripts.

caleywoods:
I would be one of the people who have fixed my scripts for the time being and will have to revert those changes back once the event goes back to its original name. 

If the event is that important to someones script they will likely come looking for information in the same way I found this issue.

PR #2700, opened 2016-02-01
Add "gifv" support for giphy

empyrical:
This PR overhauls the site module for giphy.com to make the expandos act like imgur's "gifv" expandos. All of their gifvs are now available in MP4 (but not WebM yet sadly) so this is possible.

It uses most of the same code/templating from the imgur module - I noticed there's some code there for a mediacrush-based giphy player but it doesn't seem to be used anywhere.

RegExes are not my strong suit but I've gone through a bunch of gifs on the [domain listing](https://www.reddit.com/domain/giphy.com/) and it seems to work pretty thoroughly.

PR #2798, opened 2016-03-19
Unified video control interface

larsjohnsen:
Unifies the video templates and Javascript, so that every video expando has the same accompanying functionality.

What is done:
* Refactored the `convertGifToGfycat` module.
* Created video controller `videoAdvanced`, which is based on the controller for Imgur.
* Modified `generateVideoExpando` to pipe `videoAdvanced` into `generateGenericExpando` .

TODO:
- [x] Convert hosts that use custom template to `advancedVideo`.
- [x] Convert hosts with only basic video support to automatically use `advancedVideo`.


lib/modules/hosts/gfycat.js
```diff
@@ -33,7 +34,13 @@ addLibrary('mediaHosts', 'gfycat', {
 
 			// set the max width to the width of the entry area
 			element.querySelector('video').style.maxWidth = `${$(elem).closest('.entry').width()}px`;
-			window.gfyObject(element, elem.href, info.frameRate);
+
+			let url = elem.href;
```
erikdesjardins: Might as well just inline this (`url: elem.href`) in the object below

lib/modules/hosts/gifyoutube.js
```diff
@@ -29,7 +30,13 @@ addLibrary('mediaHosts', 'gifyoutube', {
 
 			// set the max width to the width of the entry area
 			element.querySelector('video').style.maxWidth = `${$(elem).closest('.entry').width()}px`;
-			window.gifyoutubeObject(element, elem.href, `https://share.gifyoutube.com/${id}.gif`);
+
+			let url = elem.href;
```
erikdesjardins: Ditto.

lib/modules/hosts/imgur.js
```diff
@@ -14,7 +14,7 @@ addLibrary('mediaHosts', 'imgur', {
 
 	// hashRe: /^https?:\/\/(?:i\.|edge\.|www\.)*imgur\.com\/(?:r\/[\w]+\/)?([\w]{5,}(?:[&,][\w]{5,})?)(\..+)?(?:#(\d*))?$/i,
 	// the modified regex below fixes detection of "edited" imgur images, but imgur's edited images are broken right now actually, falling into
-	// a redirect loop.  preserving the old one just in case.  however it also fixes detection of the extension (.jpg, for example) which
+	// a redirect loop.	preserving the old one just in case.	however it also fixes detection of the extension (.jpg, for example) which
```
erikdesjardins: Minor nit: whitespace. Not sure why there were double spaces there in the first place, you can replace them with single spaces or put them back, whatever you want.

lib/modules/showImages.js
```diff
@@ -1345,35 +1345,30 @@ addModule('showImages', (module, moduleID) => {
 		$(header).safeHtml(imageLink.imageTitle);
 		imgDiv.appendChild(header);
 
-		const template = await RESTemplates.load('GfycatUI');
```
larsjohnsen: Is this obsolete code? lib/modules/hosts/gfycat.js

erikdesjardins: This is for .gifs that get converted to gfycat videos on-the-fly, IIRC.

If you're feeling particularly adventurous, you could try to run them through `hosts/gfycat.js` after they're converted, but `showImages` is a bit of a mess - it probably won't be easy or clean to do.

lib/vendor/advancedvideo.js
```diff
@@ -0,0 +1,74 @@
+(function () {
+	var videoAdvanced;
+
+	videoAdvanced = function (element, {url, fallback, frameRate = 29.997, frameNum, source}) {
```
erikdesjardins: You unfortunately can't use destructuring (or any ES6 features) here because files in `vendor/` are not compiled.

Alternatively we may want to move this into `showImages.js` or its own file outside of `vendor/`, although we don't have a good story for the latter right now.

erikdesjardins:
:+1: LGTM after we decide what to do about my last comment (will likely get back to you in a few days, if not earlier).

lib/vendor/advancedvideo.js
larsjohnsen: `videoAdvanced` is now moved to `showImages.js`.

lib/modules/hosts/pornbot.js
```diff
@@ -20,24 +18,19 @@ addLibrary('mediaHosts', 'pornbot', {
...
 				});
 			}
-			const element = template.html(video)[0];
-			const v = element.querySelector('video');
 
-			// set the max width to the width of the entry area
-			v.style.maxWidth = `${$(elem).closest('.entry').width()}px`;
-			window.pornbotObject(element, elem.href, info.poster);
-			return element;
+			return modules['showImages'].videoAdvanced(video, {
```
erikdesjardins: Do you think this module could be converted from `type = 'GENERIC_EXPANDO';` to `type = 'VIDEO';`, because then...

lib/modules/showImages.js
```diff
@@ -1761,6 +1561,98 @@ addModule('showImages', (module, moduleID) => {
 		}
 	};
 
+	module.videoAdvanced = async function (options) {
```
erikdesjardins: ...we wouldn't need to expose this, it could just be `async function videoAdvanced(...`.

lib/modules/showImages.js
```diff
@@ -1456,20 +1259,6 @@ addModule('showImages', (module, moduleID) => {
 		mediaDiv.appendChild(element);
 		wrapperDiv.appendChild(mediaDiv);
 
-		expandoButton.addEventListener('click', () => {
-			if (options.media.blob_type === 'video' || options.media.blob_type === 'audio') {
```
larsjohnsen: Obsolete.  `blob_type` is never assigned.

lib/modules/showImages.js
```diff
@@ -1180,90 +1164,22 @@ addModule('showImages', (module, moduleID) => {
...
+	async function generateVideoExpando(expandoButton, options) {
+		// Set defaults
+		options.autoplay = typeof options.autoplay === 'undefined' ? module.options.autoplayVideo : options.autoplay;
+		options.controls = typeof options.controls === 'undefined' ? true : options.controls;
+		options.frameRate = typeof options.frameRate === 'undefined' ? 29.997 : options.frameRate;
+		options.loop = typeof options.loop === 'undefined' ? false : options.loop;
+		options.muted = typeof options.muted === 'undefined' ? true : options.muted;
+		options.playbackRate = typeof options.playbackRate === 'undefined' ? 1 : options.playbackRate;
+		options.reversable = typeof options.reversable === 'undefined' ? false : options.reversable;
+		options.time = typeof options.time === 'undefined' ? 0 : options.time;
```
larsjohnsen: There's likely some better way to do this.

lib/modules/showImages.js
```diff
@@ -775,17 +770,6 @@ addModule('showImages', (module, moduleID) => {
 			switch (imageLink.type) {
 				case 'IMAGE':
 				case 'GALLERY':
-					if (module.options.convertGifstoGfycat.value) {
-						const gif = new RegExp('^(http|https|ftp)://.*\\.gif($|/?)');
-						if (gif.test(imageLink.src)) {
```
larsjohnsen: I couldn't find any instance where this was executed.

lib/modules/showImages.js
erikdesjardins: :+1:

lib/modules/showImages.js
erikdesjardins:
```js
const {
	autoplay = module.options.autoplayVideo,
	controls = true,
	frameRate = 29.997,
	loop = false,
	muted = true,
	playbackRate = 1,
	reversable = 0,
	time = 0
} = options;

const element = await videoAdvanced({ autoplay, controls, frameRate, loop, muted, playbackRate, reversable, time });
```

It's better but not quite ideal.

empyrical: ❤ es6! That's probably the most readable solution to that problem.

lib/modules/showImages.js
```diff
@@ -1910,9 +1790,19 @@ addModule('showImages', (module, moduleID) => {
...
-			elem.src = elem.href;
+		async handleLink(elem) {
+			if (module.options.convertGifstoGfycat.value && /^(http|https|ftp):\/\/.*\.gif($|\/?)/.test(elem.href)) {
+				const { gfyName: id } = await RESEnvironment.ajax({
+					url: RESUtils.string.encode`//upload.gfycat.com/transcodeRelease?fetchUrl=${elem.href}`,
+					type: 'json',
+					cacheFor: RESUtils.DAY
+				});
+
+				await module.siteModules["gfycat"].handleLink(elem, [, id]);
```
larsjohnsen: I was mistaken about `convertGifToGfycat`, but found a way to keep it and avoid redundancy.

lib/modules/showImages.js
```diff
@@ -1910,9 +1790,19 @@ addModule('showImages', (module, moduleID) => {
...
 			elem.src = elem.href;
+
+			if (module.options.convertGifstoGfycat.value && (/^(http|https|ftp):\/\/.*\.gif($|\/?)/).test(elem.href)) {
+				const { gfyName: id } = await RESEnvironment.ajax({
+					url: RESUtils.string.encode`//upload.gfycat.com/transcodeRelease?fetchUrl=${elem.href}`,
+					type: 'json',
+					cacheFor: RESUtils.DAY
+				});
+
+				if (id) await module.siteModules['gfycat'].handleLink(elem, [undefined, id]);
```
larsjohnsen: I was mistaken about `convertGifToGfycat`, but found a way to keep it and avoid redundancy.

larsjohnsen:
I added a progress bar, which I hope is not too distracting.
image

honestbleeps:
does the progress bar only show on hover / go away after a couple seconds of not-hover? that'd probably be ideal.

Ajedi32:
So correct me if I'm wrong, but does this replace the default HTML 5 video controls with the UI shown in the screenshot above? If so, I may have a few UX concerns...

larsjohnsen:
@honestbleeps The current implementation is always showing the grey bar with the progress indicator. I do however want the indictor to be minimally obtrusive, so I believe it is a good idea to either hide it or lower its contrast to be almost imperceptible.

@Ajedi32 
My hope is that it can supplement the browser's HTML5 video controls, mostly by allowing the user to change the playback rate. To some degree would this replace the default native controls, as they are disabled on most GIF media hosts (by design in RES).  Would not showing the duplicate controls when the native controller are enabled remove some of your concerns?

Ajedi32:
Ah okay, so you're saying this would be in addition to the current video playback controls, and it would also display for GIF-like videos?

In that case I don't really have any objections. As long as I'm here now though, I might as well add my 2 cents:

1. It seems kinda strange how the pause button appears on the right side of the video in the screenshot, not the left like just about every other video player in existence.
2. I want to make sure the ability to resize the videos by dragging isn't broken. That's one of the features I really like about the built-in player vs site-specific players like YouTube, etc.
3. Let's avoid adding redundant controls to videos which already have the built-in browser playback control. Either don't display these controls on those videos, or let's replace the browser controls entirely with something better than what browsers already have.

empyrical:
should my pull request in #2700 be shuttered and redone for this?

larsjohnsen:
Thanks, @Ajedi32 
Resize by tagging stays intact.

Updated looks:
Non-hover, browser controls disabled:
image
Hover, browser controls disabled:
image
Hover, browser controls enabled:
image

larsjohnsen:
Added a expandable panel for playback rate options which can toggled by clicking on the current speed text:
image

lib/modules/showImages.js
```diff
@@ -1761,6 +1576,132 @@ addModule('showImages', (module, moduleID) => {
...
+			if (!optionsSpeed.initialized) {
+				optionsSpeed.initialized = true;
+
+				optionsSpeed.addEventListener('click', e => {
+					switch (e.srcElement.dataset['option']) {
+						case 'speed':
+							vid.playbackRate = parseFloat(e.srcElement.dataset['speed'], 10) || 1;
+							break;
+						case 'save':
+							module.options.defaultPlaybackRate.value = vid.playbackRate;
```
larsjohnsen: This value is not actually saved to the back-end, so any speed saved is only remembered on the current page. Any thoughts on how to fix this, or is this a bad pattern?

Ajedi32: To me, "Set as default" strongly implies that this is persistent behavior. So my take is that it should either be made persistent, or the text should be re-worded. (Or as a third option, the feature could just be removed. I'm not actually sure how useful a "default" playback rate that only affects the current page would be.)

erikdesjardins: :+1: on removing it.

I'm not sure that we need more than one way to change the speed, either - having a dropdown seems redundant when you can accomplish the same thing with a few more clicks on the +/- buttons.

larsjohnsen: I agree that the speed selectors are mostly redundant. The idea was to have a way to easily reset the speed, but on second thought might that not be all that necessary.

I am however still interested in finding a way to have the playback rate remembered, either temporarily (till next page load) or persistent (by saving it somewhere). I think the dropdown can be avoided by having an option in the console for whether it should be remembered automatically, and if we find no way to do that, have the user manually set the playback rate manually.

lib/modules/showImages.js
```diff
@@ -1636,7 +1446,7 @@ addModule('showImages', (module, moduleID) => {
 		}
 		// we need to use a different onload event for videos...
 		if (mediaTag.tagName === 'VIDEO') {
-			mediaTag.addEventListener('loadeddata', handleMediaLoad);
+			mediaTag.addEventListener('loadedmetadata', handleMediaLoad);
```
larsjohnsen: Updated so that expando size is adjusted earlier, as the video element is sized at `loadedmetadata`.

lib/modules/hosts/gifyoutube.js
```diff
@@ -6,42 +6,20 @@ addLibrary('mediaHosts', 'gifyoutube', {
 		(/^https?:\/\/share\.gifyoutube\.com\/(\w+)\.gif/i).exec(href)
 	),
 	async handleLink(elem, [, id]) {
```
erikdesjardins: Minor nit: this `async` can be removed now that there's nothing `await`ed.

lib/modules/showImages.js
```diff
@@ -771,21 +774,22 @@ addModule('showImages', (module, moduleID) => {
 				}
 			}
 		} else if (showHide) {
+			if (module.options.convertGifstoGfycat.value &&
+					imageLink.type === 'IMAGE' &&
+				(/^(http|https|ftp):\/\/.*\.gif($|\/?)/).test(imageLink.src)) {
+				const { gfyName: id } = await RESEnvironment.ajax({
+					type: 'json',
+					url: RESUtils.string.encode`//upload.gfycat.com/transcodeRelease?fetchUrl=${imageLink.src}`,
```
erikdesjardins: This can be:
```js
url: '//upload.gfycat.com/transcodeRelease',
data: { fetchUrl: imageLink.src },
```
(it'll still be encoded)

lib/modules/showImages.js
```diff
@@ -1181,90 +1185,26 @@ addModule('showImages', (module, moduleID) => {
...
+	async function generateVideoExpando(expandoButton, options) {
+		const filledOptions = {
+			autoplay: module.options.autoplayVideo,
+			controls: true,
+			frameRate: 24,
+			loop: false,
+			muted: true,
+			playbackRate: 1,
+			reversable: false,
+			time: 0
```
erikdesjardins: Object spread can be used here instead of `Object.assign`, although it might be a bit less obvious what's going on:
```js
const filledOptions = {
  /* ... */
  time: 0,
  ...options
};
```

erikdesjardins:
LGTM % those nits.

Anyone have any final comments before I test and merge? (hopefully some time this week)

andytuba:
Final comments: it's great to finally unify these video players, especially with several recent new media hosts that were mostly copypaste jobs. Big thanks, @larsjohnsen!

erikdesjardins:
Thanks @larsjohnsen!

image

PR #2799, opened 2016-03-19
Add in npm install into prestart script

ZakariaRidouh:
In this PR, it removes another command that needs to get run to start up the development environment. Now, `npm install` is run during the pre start script before `gulp clean`. Hope this helps! 

:smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,10 +86,10 @@ First time installation:
 
 1. Install [node.js](http://nodejs.org) (version 4+).
 1. Install [Python 2](https://www.python.org/downloads/) (*not* version 3).
-1. Navigate to your RES folder.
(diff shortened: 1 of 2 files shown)


erikdesjardins:
:-1: This is ostensibly a good idea, but in the long run, I think the overhead would end up costing us more time than we save - it's not usually necessary to run `npm install`, and when it is, it's not unreasonable to do manually.

On slower machines `npm install` can take nearly a minute to traverse our dependency tree, which is too long when the initial build takes ~30 seconds.

mc10:
^ I'm going to decline for the above reasons. I think it's reasonable to expect users to run `npm install` before running any further scripts.

PR #2809, opened 2016-03-20
keyboardNav: fix moveDownSibling for lock to top

VirtualSatai:
The option "lock to top" doesn't work as described when using the "moveDownSibling" keybind in the
current master. Instead it follows the behavior of legacy.

This small patch fixes that.

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -922,7 +922,11 @@ addModule('keyboardNav', (module, moduleID) => {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


beck:
Why keep the `legacy` branching logic at all? What would happen if it always uses the configured pref?

I had a similar issue with the hard-coded scroll style, using the config fixed it as well:
https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2830/files

VirtualSatai:
@beck Didn't want to risk changing the behaviour of the other modes, as I havn't tried them out.

PR #2812, opened 2016-03-22
Use chrome.storage in Chrome

erikdesjardins:
Should have been part of #2689, but I didn't do enough research.

Turns out the `storage` and `unlimitedStorage` permissions [don't cause warnings](https://developer.chrome.com/extensions/permission_warnings#nowarning).

Edit: tested this with another published extension, it definitely doesn't seem to cause warnings

### DIFF
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -214,62 +214,25 @@ addListener('permissions', async ({ operation, permissions, origins }, { id: tab
 	}
 });
 
-// Circular references can't exist in storage, so we don't need to consider that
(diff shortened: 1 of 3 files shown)


erikdesjardins:
Merging this now cause it's critical that we find any bugs.

Looks like we'll still have Edge compatibility too, since the [mouse gestures extension](https://dev.windows.com/en-us/microsoft-edge/extensions/#installing-extensions) uses `chrome.storage`.

Issue #2634, opened 2016-01-13
XKCD Expandos


Ajedi32:
It'd be great if links to XKCD comics were expanded by RES.

Note that I'm currently working on this feature myself.

Ajedi32:
So here's the deal, [XKCD's API](https://xkcd.com/json.html) doesn't send any CORS headers. I'm currently looking at the code that gets around that for Twitter and OneDrive, but those are both the `GENERIC_EXPANDO` type. Any ideas on how to do this with the `IMAGE` expando type?

As I understand it, that whole method for working around a lack of CORS is actually just a big hack right now. Maybe it's time to refactor?

erikdesjardins:
> time to refactor

Coming to a master branch near you...eventually:

![image](https://cloud.githubusercontent.com/assets/7673145/12285078/5959daf2-b984-11e5-982f-f338287aad43.png)

Ajedi32:
:+1: Nice.

Oooh, and we get to use await too? Sweet! :wink:

Ajedi32:
FYI: http://forums.xkcd.com/viewtopic.php?f=10&t=113315

andytuba:
> await

Yep, master has transpilation now :)

Ajedi32:
Are we also transitioning to native Promises instead of the jQuery-based ones I've seen scattered around the codebase? The example media host seems to suggest that we need to return a jQuery promise object from the `handleLink` and `handleInfo` methods, which AFAIK is impossible if we declare them as `async`. (They'll always return a native promise then.)

erikdesjardins:
Yeah, I'm converting everything to native promises. I'm also getting rid of `handleInfo` since it adds pointless boilerplate when you can `await` as many things as you need in `handleLink`.

I'm done almost everything, just need to do a few more things (repenting for my sin of caching to permanent storage, etc.).

honestbleeps:
FWIW, I believe converting to native promises may be the final nail in the coffin for Opera 12 support - and I'm 100% ok with that.  I understand its users are passionate, but there are not enough of them to justify the continued work/maintenance required. It has an IE6-ish level of odd quirks etc.

erikdesjardins:
[He's dead, Jim](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2565)

erikdesjardins:
#2815

PR #2815, opened 2016-03-22
Add XKCD media host (via oembed)

thybag:
Captions currently don't work, although this looks to be fixed by https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2534/files

### DIFF
--- a/lib/modules/hosts/xkcd.js
+++ b/lib/modules/hosts/xkcd.js
@@ -6,7 +6,7 @@ addLibrary('mediaHosts', 'xkcd', {
 		// Go get data
 		const data = await RESEnvironment.ajax({
 			url: 'https://noembed.com/embed',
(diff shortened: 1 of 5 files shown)


lib/modules/hosts/xkcd.js
```diff
@@ -0,0 +1,24 @@
+addLibrary('mediaHosts', 'xkcd', {
+	domains: ['xkcd.com'],
+	attribution: false,
```
erikdesjardins: You can use something like `logo: 'https://xkcd.com/s/919f27.ico'` instead of `attribution: false` to show the favicon below the image, e.g.

erikdesjardins:
:+1: LGTM
(that's a much better favicon url than mine)

Ajedi32:
:+1: I tried implementing this earlier using the official XKCD API (#2634), but got stuck because they didn't support CORS. This is much nicer...

PR #2818, opened 2016-03-25
Mediahost - load on demand

thybag:
Hello,

This is essentially just a slightly altered version of https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2815 . The relevant tweaks are in https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/f23bc78eadb265f32ef8e8e5ce2375ef4302a835

Effectively, the aim of this tweak is to try and avoid making unneeded API calls, as for example when hitting a page like https://www.reddit.com/r/xkcd the previous implementation would be kicking off 15 or so requests no noembed, while chances are I may only actually ever want to look at one or two of the comics.

The current implementation just adds an additional call in the `generateImageExpando` function to the `elem`'s onFirstLoad method, which then calls off and generates everything needs for the standard expando process to take place as normal.

If the approach sounds sane/useful - I'd be happy to add the functionality to all the Expando types. Additionally I'd be tempted to make it a "built in", but optional part of the mediahost's (so people could just add a method like the `initExpando` (unsure if thats a good name)) whenever they wanted to make use of it. (In my view whenever significant processing is needed, or interactions with external API's.

Anyways, just wanted to get a feel for what other people thought of the idea?

### DIFF
--- a/lib/modules/hosts/xkcd.js
+++ b/lib/modules/hosts/xkcd.js
@@ -3,10 +3,16 @@ addLibrary('mediaHosts', 'xkcd', {
 	logo: 'https://xkcd.com/favicon.ico',
 	detect: href => (/^https?:\/\/(?:www\.|m\.)?xkcd\.com\/([0-9]+)/i).exec(href),
 	async handleLink(elem, [, id]) {
-		// Go get data
+		elem.type = 'IMAGE';
+		elem.imageId = id;
+		// On first expand, init data
+		elem.onFirstExpand = this.initExpando;
+	},
+	async initExpando(elem){
+
 		const data = await RESEnvironment.ajax({
 			url: 'https://noembed.com/embed',
-			data: { url: `http://xkcd.com/${id}/` }, // noembed doesn't believe in the HTTPS urls, so these need to be "http"
+			data: { url: `http://xkcd.com/${elem.imageId}/` }, // noembed doesn't believe in the HTTPS urls, so these need to be "http"
 			type: 'json'
 		});
 
@@ -16,7 +22,6 @@ addLibrary('mediaHosts', 'xkcd', {
 		const img = temp.querySelector('img');
 
 		// Set attributes
-		elem.type = 'IMAGE';
 		elem.imageTitle = img.getAttribute('alt');
 		elem.caption = img.getAttribute('title');
 		elem.src = img.getAttribute('src').replace('https://noembed.com/i/', ''); // remove noembed image proxy & link direct
--- a/lib/modules/hosts/xkcd.js
+++ b/lib/modules/hosts/xkcd.js
@@ -6,7 +6,7 @@ addLibrary('mediaHosts', 'xkcd', {
 		// Go get data
 		const data = await RESEnvironment.ajax({
 			url: 'https://noembed.com/embed',
-			data: { url: `http://xkcd.com/${id}/` }, // noembed doesn't believe in the HTTPS urls, so these need to be "http" 
+			data: { url: `http://xkcd.com/${id}/` }, // noembed doesn't believe in the HTTPS urls, so these need to be "http"
 			type: 'json'
 		});
 
--- a/lib/modules/hosts/xkcd.js
+++ b/lib/modules/hosts/xkcd.js
@@ -1,0 +1,24 @@
+addLibrary('mediaHosts', 'xkcd', {
+	domains: ['xkcd.com'],
+	attribution: false,
+	detect: href => (/^https?:\/\/(?:www\.|m\.)?xkcd\.com\/([0-9]+)/i).exec(href),
+	async handleLink(elem, [, id]) {
+		// Go get data
+		const data = await RESEnvironment.ajax({
+			url: 'https://noembed.com/embed',
+			data: { url: `http://xkcd.com/${id}/` }, // noembed doesn't believe in the HTTPS urls, so these need to be "http" 
+			type: 'json'
+		});
+
+		// documentFragment doesn't support innerHTML, so use div instead
+		const temp = document.createElement("div");
+		temp.innerHTML = data.html;
+		const img = temp.querySelector("img");
+
+		// Set attributes
+		elem.type = 'IMAGE';
+		elem.imageTitle = img.getAttribute("alt");
+		elem.caption = img.getAttribute("title");
+		elem.src = img.getAttribute("src").replace("https://noembed.com/i/", ""); // remove noembed image proxy & link direct
+	}
+});
--- a/lib/modules/hosts/xkcd.js
+++ b/lib/modules/hosts/xkcd.js
@@ -1,6 +1,6 @@
 addLibrary('mediaHosts', 'xkcd', {
 	domains: ['xkcd.com'],
-	attribution: false,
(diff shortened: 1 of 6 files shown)


Ajedi32:
I don't really know. While this is certainly a better use of network resources, it seems to be that the best way to improve performance from the user's perspective is to have all this data already pre-fetched before they even click the expand button.

One thing I do really like about this though is that it will make the expand buttons appear without waiting on a network request. (Right?) That seems likely to be a pretty significant improvement to perceived performance.

thybag:
> One thing I do really like about this though is that it will make the expand buttons appear without waiting on a network request. 

Indeed :) Hard to really say what people would notice "/  My assumption would be that in most cases, loading just one resource would generally be fast enough (in response to a click) that its not really an issue.

Not really sure whether "more instant buttons", or "more instant expanding" would be the more noticeable change. Personally, it just feels a bit wrong to see all the xhr requests firming away in the background when in a lot of cases the data may never be used - although admittedly thats not something 99.999% or RES's user base are likely to care about :p

Ajedi32:
I think the best approach when possible is to show buttons immediately, then start sending AJAX requests in the background afterwards to get ready for when the user actually clicks them.

andytuba:
For some cases, it's necessary to send the network request for metadata which affects the button, such as button media type (static image, gifv/gfy/giphy, album, video with sound) or number of images in an album (shown in title top).

For expandos like that, maybe show a generic expando button with a throbber overlaid?

lib/modules/showImages.js
```diff
@@ -817,8 +817,14 @@ addModule('showImages', (module, moduleID) => {
 		updateParentHeight(expandoButton);
 	}
 
-	function generateImageExpando(expandoButton) {
+	async function generateImageExpando(expandoButton) {
 		const imageLink = expandoButton.imageLink;
+
+		// Handle init if needed
+		if (typeof imageLink.onFirstExpand === 'function') {
+			await imageLink.onFirstExpand(imageLink);
```
andytuba: +async/await: tidy!

lib/modules/hosts/xkcd.js
```diff
@@ -0,0 +1,28 @@
+addLibrary('mediaHosts', 'xkcd', {
+	domains: ['xkcd.com'],
+	logo: 'https://xkcd.com/favicon.ico',
+	detect: href => (/^https?:\/\/(?:www\.|m\.)?xkcd\.com\/([0-9]+)/i).exec(href),
```
andytuba: Unrelated to this PR, but it would be neat if [showImages `detect` call](https://github.com/thybag/Reddit-Enhancement-Suite/blob/feature/xkcd-alt/lib/modules/showImages.js#L613) would `detect.test ? detect.test(href) : detect.call ? mod.detect(elem.href, elem) : false`

lib/modules/showImages.js
```diff
@@ -817,8 +817,14 @@ addModule('showImages', (module, moduleID) => {
 		updateParentHeight(expandoButton);
 	}
 
-	function generateImageExpando(expandoButton) {
+	async function generateImageExpando(expandoButton) {
 		const imageLink = expandoButton.imageLink;
+
+		// Handle init if needed
+		if (typeof imageLink.onFirstExpand === 'function') {
```
andytuba: Re [my comment about preloading metadata](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2818#issuecomment-201909688), I'd vote for a function name more like "fetchLinkMetadata".

andytuba:
I'm also in favor of deferring XHR requests to "as late as possible / only as early as necessary", especially for capped usage like Google's YouTube API or to avoid stressing smaller hosts. IIRC @honestbleeps wanted to build that into BetterZoom as well.

It could include a concession to the user in the form of an option like `preloadMetadata` (just before `preloadAlbum`) which follows the current legacy behavior of "API all the things".

lib/modules/showImages.js
thybag: Renamed :+1:

lib/modules/hosts/xkcd.js
thybag: Ooo, I like this idea. Means media hosts can just provide a regex directly if they ain't doing anything special :)

thybag:
Have updated with a couple of the suggestions applied. The new version should now work for all expando types & doesn't require just auto run `fetchLinkMetadata` if its defined when an expando is activated.

Also included the suggested tweak so the detect can be provided straight as a regex to exec if wanted.

andytuba:
Sweet! Now which site modules need preload?  Imgur for sure.  I think giphy supports albums.  

I just remembered the YouTube video length api call is part of betteReddit, not showImages, but that kinda makes more sense because it applies to mostly reddit/embedly-supplied expandos.

thybag:
Well, pixpit, graphiq & wikipedia would probably benefit from the change, but it'd probably make sense for me to update them in to their own pull requests once this is merged.

From a quick look, i suspect the following would also be easy to port to use the mechanism in this pull.

* gfycat
* fitbamob
* flickr
* futurism
* gifyoutube
* github
* iloopit
* min.us
* photobucket
* pornbot
* soundcloud
* tumblr
* possibly twitter

Basically anything where the TYPE is already known, for the rest I'd have to look in to getting a solution like the one you suggest in https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2818#issuecomment-201909688 working.

thybag:
Just wondering if people are interested in me progressing this further?

Is it worth me porting over a load of other media hosts to work in this way as part the pull? or is it better to have this pull more stand alone-y?

Ajedi32:
I really like the idea. Current enhancements I'd like to see in this area (obviously just my personal preferences, feel free to push back):

- [ ] Display a generic expando icon (possibly with some kind of subtle animated loading indicator) immediately after `detect` returns true (to improve UI responsiveness)
- [ ] When `handleLink` returns, update the generic icon to the correct media type if available (or remove the icon if `handleLink` rejects)
- [ ] Add a new "auto" option to the `preloadLinkMetadata` config
  - [ ] Preloads most media hosts (saves latency from an extra network request on expand)
  - [ ] Doesn't preload hosts with low API limits
  - [ ] Doesn't preload smaller hosts that might struggle under the additional traffic
  - [ ] Doesn't preload when on a capped data connection (can we detect this?)
  - [ ] Make it the default

thybag:
I'm a little unsure about the value of the `auto` option, in most cases the latency of a single ajax request is going to be somewhat negligible in comparison to actually loading the iframe/image/video anyway. 

I guess another option if is latency is a concern could be to have a smart-prefetch, for example triggering the `fetchLinkMetadata` when a user first hovers over the expando icon - which in most cases would give it a small head start vs the user actually clicking. (Although this wouldn't help for people using keyboard shortcuts).

re: the generic icon (for cases where the final type is unknown), i like the idea, although would probably want to avoid trying to design an icon if possible :p For the majority of cases (at least that i can think of) the type is known anyway, so its probably not an immediate concern. (although an intermediary "loading"expando icon  to show when a user triggers an expando could be a nice touch anyway for if one is being slow for whatever reason)

larsjohnsen:
@thybag A compromise could be to prefetch when the expando-button becomes visible.

PR #2829, opened 2016-03-27
Remove npm-shrinkwrap.json

mc10:
Was added in #2603. Shrinkwrap seems unnecessary for devDependencies, and is [excluded by default](https://docs.npmjs.com/cli/shrinkwrap) for that reason.


erikdesjardins:
I originally added `npm-shrinkwrap` because it's the only way to force npm to install consistent versions of every dependency and sub-dependency. Without it, nobody (in IRC) could get master to work without nuking their `node_modules` (from 4.6.1) first.

I guess it's probably overkill now; I'd be fine with removing it if we pin our dependencies to exact versions instead of caret ranges. Sub-dependencies will still be a free-for-all, but...they'll probably be fine.

erikdesjardins:
[Hah, the build even fails on npm 2 when the deps aren't locked down,](https://travis-ci.org/honestbleeps/Reddit-Enhancement-Suite/jobs/118784519#L179) cause someone added a peerDependency on a minor/patch release (tsk tsk, whichever package that was).

That's a pretty good case for npm-shrinkwrap if I've ever seen one ;)

mc10:
Yeah, I'm alright with pinning down versions.

Issue #2628, opened 2016-01-13
mediaBrowseMode is always "lock to top" despite selected scrollStyle


beck:
This is especially problematic when using keyboard navigation to hide posts. When you stop on a post, and the previous requests to hide are being executed, the post in view will pop out of frame when the hide requests finish.

Example:
- hide three posts & stop
- post pops out of view

Video: https://dl.dropboxusercontent.com/u/1728414/2015%20RES%20hiding.mov

Before the recent release this was not an issue because `mediaBrowseMode` seemed to abide `scrollStyle` (or at the very least, would not lock to top)

rel: https://www.reddit.com/r/RESissues/comments/40mftf/bug_keyboard_nav_uses_lock_to_top_when_it_should/

Ajedi32:
Possibly related to #2622?

andytuba:
thanks for report!

@Ajedi32 related but slightly different i think.  i messed a lot with scrolling in 4.6.0.

erikdesjardins:
Are you certain that it used to act differently in 4.5.4? In a fresh install of 4.5.4, mediaBrowseMode acts the same way it does now - ignoring scrollStyle and always locking to top when you move down.

It is entirely possible that there's a third setting affecting this, though.

beck:
Installed master, seems to be fixed.  Assuming it was corrected with https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2668

Thanks @erikdesjardins

beck:
There has been some cleanup & refactoring, issue seems to have been re-introduced.

Removing the hardcoded scroll style fixes for me:
https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2830

larsjohnsen:
I don't believe this is an issue anymore: `scrollStyle` is only set to `top` if `scrollOnExpando` is enabled https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/218c1bdae2b2595793d209c5ce50933bc5644b53/lib/modules/keyboardNav.js#L806 (the `scrollOnExpando` description should be updated to reflect this).

erikdesjardins:
And #3347 made onHideMoveDown not scroll

beck:
This issue should be reopened as a regression.
Here is a pull to fix: https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2628

PR #2830, opened 2016-03-27
Remove hardcoded scroll style

beck:
Fixes https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2628

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -815,6 +815,7 @@ addModule('keyboardNav', (module, moduleID) => {
 		const current = fromThing;
 		const things = modules['selectedEntry'].selectableThings();
 		const index = things.index(current);
+		const scrollStyle = module.options.scrollStyle.value;
 
 		let target;
 		let targetIndex = Math.max(1, index);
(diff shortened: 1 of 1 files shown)


thybag:
Think the linter on travis is complaining because it wants you to use the new ES6 magic, ie. `_moveToThing(target, { scrollStyle: scrollStyle });` can be just `_moveToThing(target, { scrollStyle });`

erikdesjardins:
:-1:  I'm still not convinced that #2628 is valid.

`mediaBrowseMode` has always locked to top and ignored the selected scroll style, by design. Maybe there was a bug in dev where it didn't, but not in any release, AFAIK. [The post](https://www.reddit.com/r/RESissues/comments/40mftf/bug_keyboard_nav_uses_lock_to_top_when_it_should/) linked in #2628 is a different issue.

[Try out a 4.5.4 build and see for yourself.](https://github.com/honestbleeps/Reddit-Enhancement-Suite/files/191030/res-4.5.4.zip)

This also ruins the UX for anyone that uses j/k to scroll, because you won't be able to see the expandos once you get the bottom of the page.

A better solution to your problem would be to (have an option to? or just default behaviour with fixHideLinks) make hidden posts transparent instead of removing them.

beck:
> mediaBrowseMode has always locked to top and ignored the selected scroll style, by design

Is this good design?

With this change, it now uses my pref `directional` and the expandos step down the list of posts instead of the entire site jerking downward.

> you won't be able to see the expandos once you get to the bottom of the page.

Is it fair to consider this a separate bug that should be addressed?

What I see happening locally:

If the image loads quickly `directional` scrolling will consider the image height and scroll just enough to keep the expando in view.  If the image is slow to load, the scroll finishes and the image falls off the bottom.

This change definitely makes this race bug more painful—IMO best to expose it vs hide it by overriding user pref with hardcoded behavior.

I see another pull of similar concerns:
https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2809

erikdesjardins:
Fixed in a slightly different way by some other PRs, see #2628

PR #2833, opened 2016-03-28
Fixes #2832.  Test failed depending on the system time zone.

### LINKED ISSUE #2832: niceDateDiff test fails when system timezone is BST
This may not be very important, it's more of a development issue, but it's a test that fails, so, here it goes:

- I created a fresh Ubuntu VM to have a look at the RES code.  
- The location is set to London, and today (march 28th 2016) it means it's BST (British summer time)
- git clone https://github.com/honestbleeps/Reddit-Enhancement-Suite.git
- npm install 
- npm test => 1 test fails:

t.is(d('Mar 01 2015', 'Apr 01 2015'), '1 month')
       |                                          
       "30 days"                                  
  
  AssertionError: '30 days' === '1 month'
    Test.fn (niceDateDiff.js:20:4)

I did a couple of checks (node version doesn't seem to matter) and after changing the VM's timezone to UTC with 

sudo dpkg-reconfigure tzdata

the test passes OK.  

What's going on here is that April 1st goes into the date (at 00:00!), somehow an hour is subtracted off of it for what I'm sure are very good reasons, leaving March 31st at 23:00, and then both dates are in the same month, so it's not a month of difference any more. 
If you're wondering why an hour doesn't get subtracted from March 1st too, it's because daylight savings time happens March 27th, so Date doesn't see any reason to do anything to it.

Also, if your system's time zone west of UTC, like anywhere in America, the test also passes (there may still be an hour difference in the dates, but "into" the day, not back to "yesterday").

Complexity is fractal.

I don't think the code under test is doing anything wrong, just that the test doesn't specify the inputs to the function well enough.

So.  I would understand if this issue was closed as "won't fix".  Dates are hard.  Timezones are really, really hard.  I'm going to have a look and attempt a solution involving making the test independent of the system's timezone. 

And maybe you get more contributions from people around here if all tests pass without needing to fiddle with timezones!

MrP:
Tested with several different system time zones (BST, UTC, a bunch in America, Moscow).
Please let me know if I should do anything different in the code, the commit, PR or whatever.  I tried to follow the style guidelines, but I may have missed something.

And do feel free to close the issue, ignore the pull request, fix it yourself in some other way, copypaste the code or whatever makes your life easier. 

Kind regards,
Victor

### DIFF
--- a/lib/core/__tests__/niceDateDiff.js
+++ b/lib/core/__tests__/niceDateDiff.js
@@ -5,8 +5,11 @@ import test from 'ava';
 import { RESUtils } from '../utils';
 
 test('Various date ranges', t => {
+	function dateWithoutTimezone(dateString) {
+		return new Date(Date.parse(dateString) - (new Date(dateString).getTimezoneOffset() * 60000));
+	}
 	function d(from, to) {
-		return RESUtils.niceDateDiff(new Date(from), new Date(to));
+		return RESUtils.niceDateDiff(dateWithoutTimezone(from), dateWithoutTimezone(to));
 	}
 
 	t.is(d('Jan 01 2015', 'Jan 01 2015'), '0 days');


thybag:
Looks sane to me :)

erikdesjardins:
LGTM

erikdesjardins:
Thanks @MrP!

MrP:
Thanks to you for the quick response time!

PR #2836, opened 2016-03-29
Force `bottom: auto` on .res-navTop

empyrical:
This stops the glitch that causes `#header-bottom-right` to become as high as the header on a lot of subreddits. (this is caused by `top: 0` and `bottom: 0` being active at the same time)

It didn't used to do that until a couple of weeks ago, so the order of `<style>` tags must've moved around on reddit's end

### DIFF
--- a/lib/css/modules/_styleTweaks.scss
+++ b/lib/css/modules/_styleTweaks.scss
@@ -67,7 +67,7 @@ body:not(.res-styleTweaks-visitedStyle) {
 	#header-bottom-right {
 		top: 19px;
 		border-radius: 0 0 0 3px;
-		bottom: auto;
+		bottom: auto !important;
 	}
 
 	.beta-notice {


erikdesjardins:
In 4.6.x there's a race between RES and DOM parsing, causing the order of style tags to be indeterminate.

On master I wouldn't expect this to be necessary, since we no longer use `addCSS` for navTop, i.e. we don't inject style tags (unless you can repro?).

andytuba:
@empyrical How's it feel on master?

erikdesjardins:
Closing as I haven't been able to repro this in master. (but I can fairly often in 4.6.1)

Feel free to bump if it starts to show up again.

empyrical:
Oops, forgot about this, sorry!

Just tried it myself and it indeed seems okay on `master`, no need for this PR

PR #2841, opened 2016-03-30
Remove old Selenium tests

mc10:
We don't even support Opera anymore.


andytuba:
:+1:  I hope to bring these back .. eventually .. but hey, that's what git log is for.

PR #2847, opened 2016-04-01
readme: remove potentially misleading line about Gulp

erikdesjardins:
It seems to imply that Gulp is one of the things you need to install manually, which is no longer the case.


andytuba:
Yeah, no need to expose what is now an internal implementation detail.

PR #2857, opened 2016-04-04
add a delay on sortable widget

educcjr:
add a delay on sortable widget to prevent unwanted drags


erikdesjardins:
Thanks!

For those following along at home: specifically, this adds a delay between when you start dragging and when it replaces the element with a placeholder, so stuff doesn't get shifted around if you accidentally drag.

Issue #2867, opened 2016-04-07
New Comment Count shows up repeatedly with bad data


andytuba:
https://i.imgur.com/LiHZgp8.png

happens sporadically to my work laptop running v4.7.0 commit c2f6efcff4bbce1da703635949987bb07abad2f7

allthefoxes:
To reproduce

1. Visit a comment thread
2. Leave and come back when there are new comments highlighted
3. Expand a "show more comments"
4. Profit
5. Repeat as desired until you reach infinity

---

Images:

https://www.allthefoxes.me/res/bugdir/infinite/step1.png - something with new comments
https://www.allthefoxes.me/res/bugdir/infinite/step2.png - Click a "load more comments"
https://www.allthefoxes.me/res/bugdir/infinite/step3.png - Results
https://www.allthefoxes.me/res/bugdir/infinite/step4.png - Repeated a few more times
https://www.allthefoxes.me/res/bugdir/infinite/infinite.png - Infinity

As the official bug namer, I declare this bug "4.7 Infinity"

thybag:
Going to have a dig at this one

PR #2872, opened 2016-04-09
Fix duplicating new comment counts

thybag:
Fixes https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2867

It appears a couple of issues were at play with this one.

* The `updateCommentCountFromMyComment` didn't check that the new comments added were actually by the user (so expanding a collapsed comment would run the `updateCommentCount` logic multiple times)
* The `updateCommentCount` was appending a new "new comments" section each time it got run
* Also, it was counting the numbers within the new comment section (since .textContent returned the child text to), which is why the numbers went so crazy.

Have refactored it a bit (mostly to get my head around it), and it all appears to be working correctly for me.

* `updateCommentCountFromMyComment` - new checks the user is logged in & made the triggering post, before calling in to `updateCommentCount`
* `updateCommentCount` will check if `currentCommentID` is already defined, if so it will only allow the `mycomment` logic to run, thus ensures the rest can't be rerun accidentally.
* "mycomment" mode just increments the counter and calls save.
* save count logic and show count logic now live in there own functions.

Hope this all looks sane :)

### DIFF
--- a/lib/modules/newCommentCount.js
+++ b/lib/modules/newCommentCount.js
@@ -269,47 +269,90 @@ addModule('newCommentCount', (module, moduleID) => {
 		});
 	}
 
-	function updateCommentCountFromMyComment() {
-		updateCommentCount(true);
+	function updateCommentCountFromMyComment(entry) {
+		const user = RESUtils.loggedInUser();
+		// don't update for user comment count, if detected new comment wasn't from the user
+		if (user && user === $(entry).find(".author").text()) {
+			updateCommentCount(true);
+		}
 	}
 
 	let currentCommentID;
 
-	function updateCommentCount(mycomment) {
-		const IDre = /\/r\/[\w]+\/comments\/([\w]+)\//i;
-		const matches = IDre.exec(location.href);
-		let currentCommentCount = 0;
-		if (matches) {
-			currentCommentID = matches[1];
-			const thisCount = document.querySelector('#siteTable a.comments');
-			if (thisCount) {
-				// split number from the word comments
-				currentCommentCount = thisCount.textContent.replace(/\D/g, '');
-				if (commentCounts[currentCommentID]) {
-					const prevCommentCount = commentCounts[currentCommentID].count;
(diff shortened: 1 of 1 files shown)


lib/modules/newCommentCount.js
```diff
@@ -269,47 +269,90 @@ addModule('newCommentCount', (module, moduleID) => {
...
+	/**
+	 * Show new comments count to user
+	 *
+	 * @param {int} currentCommentCount - number of comments found on this page
+	 * @returns {void}
+	 */
+	function showNewComments(currentCommentCount) {
+		const prevCommentCount = commentCounts[currentCommentID].count;
+		const diff = currentCommentCount - prevCommentCount;
+		const $newString = $(`<span class="newComments">&nbsp;(${diff} new)</span>`);
```
erikdesjardins: Nit: might as well put the string straight into ```.append(`<span...```

erikdesjardins:
LGTM % that one nit

lib/modules/newCommentCount.js
```diff
@@ -269,47 +269,89 @@ addModule('newCommentCount', (module, moduleID) => {
 		});
 	}
 
-	function updateCommentCountFromMyComment() {
-		updateCommentCount(true);
+	function updateCommentCountFromMyComment(entry) {
+		const user = RESUtils.loggedInUser();
+		// don't update for user comment count, if detected new comment wasn't from the user
+		if (user && user === $(entry).find('.author').text()) {
```
andytuba: You could probably use `RESUtils.thing(entry).getAuthor()` for this.  (might have a slightly different name than getAuthor)

lib/modules/newCommentCount.js
```diff
@@ -269,47 +269,89 @@ addModule('newCommentCount', (module, moduleID) => {
...
+			updateCommentCount(true);
+		}
 	}
 
 	let currentCommentID;
 
+	/**
+	 * save an updated CommentCount
```
andytuba: woo docs

PR #2874, opened 2016-04-09
Video advanced controls only visible on hover

larsjohnsen:
New look when native controls are not active:
image


andytuba:
seems legit

erikdesjardins:
Tested; LGTM

PR #2875, opened 2016-04-09
Sort imageList before showing all

larsjohnsen:
Reduces the time before the first image is shown after clicking "Show images", since the `imageList` array may be unordered due to async media-hosts.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -495,11 +495,13 @@ addModule('showImages', (module, moduleID) => {
 	}
 
 	function updateRevealedImages(type) {
-		for (const image of imageList) {
-			if ($(image).hasClass(type) || image.imageLink.expandOnViewAll) {
-				revealImage(image, findImageFilter(image.imageLink));
-			}
-		}
+		const imagesReveal = imageList
+			.filter(v => v.classList.contains(type) || v.imageLink.expandOnViewAll)
+			.map(v => [v.getBoundingClientRect().top, v])
+			.sort((a, b) => a[0] - b[0])
+			.map(v => v[1]);
+
+		RESUtils.forEachChunked(imagesReveal, image => { revealImage(image, findImageFilter(image.imageLink)); });
 	}
 
 	function findImageFilter(image) {


lib/modules/showImages.js
```diff
@@ -472,11 +472,15 @@ addModule('showImages', (module, moduleID) => {
...
 
 	function updateRevealedImages(type) {
-		for (const image of imageList) {
-			if ($(image).hasClass(type) || image.imageLink.expandOnViewAll) {
-				revealImage(image, findImageFilter(image.imageLink));
-			}
+		let images = imageList.filter(v => $(v).hasClass(type) || v.imageLink.expandOnViewAll);
+
+		if (RESUtils.pageType() === 'linklist') {
+			images = images.map(v => [$(v).closest('.thing').data('rank'), v])
```
larsjohnsen: `Rank` only exists in `linklist`-pages, so an alternative could be to save the element number in the DOM when calling `checkElementForImage` and use that as basis for the sort.

Alternatively, change the way `imageList` is created and filled.

andytuba: It would be nice to move `$(v).closest('.thing).data('rank')` to `RESutils.thing::rank` to abstract this for future re-use.

lib/modules/showImages.js
```diff
@@ -472,11 +472,15 @@ addModule('showImages', (module, moduleID) => {
 	}
 
 	function updateRevealedImages(type) {
-		for (const image of imageList) {
-			if ($(image).hasClass(type) || image.imageLink.expandOnViewAll) {
-				revealImage(image, findImageFilter(image.imageLink));
-			}
+		let images = imageList.filter(v => $(v).hasClass(type) || v.imageLink.expandOnViewAll);
```
andytuba: (potentially pre-optimizing but) vanilla JS equivalent of `$(v).hasClass(type)` would be `v.classList.contains(type)`

lib/modules/showImages.js
```diff
@@ -472,11 +472,15 @@ addModule('showImages', (module, moduleID) => {
...
+
+		if (RESUtils.pageType() === 'linklist') {
+			images = images.map(v => [$(v).closest('.thing').data('rank'), v])
+				.sort((a, b) => a[0] - b[0])
+				.map(v => v[1]);
 		}
```
andytuba: nice lispifying and precomputing the expensive part :+1:

lib/modules/showImages.js
andytuba: > change the way imagelist is created and filled

That's a neat idea. How about something like .. in `findAllImages`, populate `imageList` with `ImagePromise` objects sorted by vertical position? This is easy to pull from `allElements` on pageload, but how how to splice in new ImagePromise when expandoing selftext or loading more comments? Should `imageList` be a tree structure mimicking the DOM tree?

larsjohnsen: I think a binary search on `offsetTop` would suffice.

erikdesjardins: Or even just change this line to
```js
images = images.map(v => [$(v).offset().top, v])
```

N is going to be small (<1000) in almost all cases, so I don't think there'd be a noticeable gain from sorting them up front. At least not enough to justify the complexity, IMO.

andytuba: > `images = images.map(v => [v.getBoundingClientRect().top, v])`

That seems best, especially since any lag in sorting the list will be greatly overshadowed by the strain of loading all those expandos :D

andytuba:
Maybe we'll optimize imageList construction later, but let's not let the perfect block the good.

PR #2876, opened 2016-04-10
Enhanced memory conservation

thybag:
A few changes to try and improve the effectiveness of `conserveMemory`.

* Attempt to speed up code (use single bool `image._resExpandoReady` to denote media lazyUnload has to care about. Avoids unnecessary calls to `getBoundingClientRect` & `reloadRevealedImage`)
* Collapsed media outside the `bufferScreens` will be tidied up and set to its un-initialized state (this seems to makes a big difference to memory on pages with lots of iframes etc)
* Fix `reloadRevealedImage` & `unloadRevealedImage` being called on collapsed images (due to `getBoundingClientRect` just returning 0 for all values in this case)


andytuba:
Hmm, how much do we care about a little cruft on ._resExpandoReady?

lib/modules/showImages.js
```diff
@@ -298,8 +307,21 @@ addModule('showImages', (module, moduleID) => {
 		const maximumTop = viewportHeight * (bufferScreens + 1);
 		const minimumBottom = viewportHeight * bufferScreens * -1;
 
-		imageList.forEach(image => {
-			if (image.imageLink && image.imageLink.media) {
+		activeImageList.forEach(image => {
+			// If image has no expandoBox, remove from this list
+			if (!image.expandoBox) {
+				activeImageList.splice(activeImageList.indexOf(image), 1);
```
erikdesjardins: This will end up skipping over the next element - you probably want to use a set.

lib/modules/showImages.js
```diff
@@ -298,8 +307,21 @@ addModule('showImages', (module, moduleID) => {
 		const maximumTop = viewportHeight * (bufferScreens + 1);
 		const minimumBottom = viewportHeight * bufferScreens * -1;
 
-		imageList.forEach(image => {
-			if (image.imageLink && image.imageLink.media) {
+		activeImageList.forEach(image => {
```
erikdesjardins: Nit: use `for (const image of activeImageList) {` when it's not chained with anything else. Also works with sets.

lib/modules/showImages.js
thybag: Ah, still not fully got my head around all the new ES6 stuff :p Set's api makes a lot more sense with what i'm doing :+1:

PR #2878, opened 2016-04-11
Multireddit quicklinks menu

andytuba:
screen shot 2016-04-10 at 8 08 33 pm

### DIFF
--- a/lib/modules/multiredditNavbar.js
+++ b/lib/modules/multiredditNavbar.js
@@ -31,7 +31,7 @@ addModule('multiredditNavbar', (module, moduleID) => {
 				['top', './top'],
 				['top this month', './top?t=month'],
 				['gilded', './gilded'],
-				['promoted', './ads'],
+				['promoted', './ads']
 			]
 		},
 		hoverDelay: {
@@ -79,7 +79,7 @@ addModule('multiredditNavbar', (module, moduleID) => {
 				pin: modules['hover'].pin.right,
 				offsetWidth: -10,
 				offsetHeight: 1,
-				bottomPadding: 0,
+				bottomPadding: 0
 			})
 			.populateWith(() => populateSectionMenu(link.href))
 			.begin();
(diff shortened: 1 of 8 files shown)


lib/modules/multiredditNavbar.js
```diff
@@ -0,0 +1,117 @@
...
+			.reduce((prev, curr) => (curr ? prev.add(curr) : prev), $())
+			.add(populateSectionItem('.', [
+				`<i>${module.moduleName}</i>`,
+				modules['settingsNavigation'].makeUrlHash(moduleID, 'sectionMenu')
+			]))
+	];
+
+	function populateSectionItem(baseUrl, link) {
+		if (!(link && link.length >= 2)) {
+			return;
```
erikdesjardins: Errors from the `consistent-return` lint can be a bit misleading; it wants all or none of the return statements to return a value (`undefined` is allowed, as long as it's explicit).

lib/modules/multiredditNavbar.js
```diff
@@ -0,0 +1,117 @@
+addModule('multiredditNavbar', (module, moduleID) => {
+	module.moduleName = 'Multireddit Navigation';
+	module.description = 'Enhance the navigation bar shown on the left side of the frontpage.';
+	module.category = 'Subreddits';
+
+	module.includes = [
+		RESUtils.regexes.frontpage
+	];
+
+	Object.assign(module.options, {
```
erikdesjardins: This should be identical to `module.options = {` ?

lib/modules/multiredditNavbar.js
andytuba: `return false` seems like a reasonable thing for the linter to ask for.

lib/modules/multiredditNavbar.js
```diff
@@ -0,0 +1,117 @@
...
+				offsetHeight: 1,
+				bottomPadding: 0,
+			})
+			.populateWith(() => populateSectionMenu(link.href))
+			.begin();
+	}
+
+	const populateSectionMenu = (baseUrl) => [
+		module.options.sectionLinks.value
+			.map((...args) => populateSectionItem(baseUrl, ...args))
```
erikdesjardins: `.map(link => populateSectionItem(baseUrl, link))` ?

andytuba: what, being explicit and performant? that's crazy talk.

PR #2882, opened 2016-04-13
Fix keynav followLink on comments pages

### LINKED ISSUE #2879: Under certain configuration, "enter" opens links twice on comment pages
Problem: 
---

Under specific configurations, pressing 'enter' from the comments page will open it both in the current tab and in a new tab.

Version:
---

RES 4.6.1
Firefox 45.0.2

Repro steps:
---

1. Open RES keyboard navigation settings.
2. Set "followLink" to "shift+enter".
3. Set "followLinkNewTab" to "enter".
4. Set "toggleChildren" to "enter". (this is the default)
5. Visit any non-self post comments page (e.g. [this one](https://www.reddit.com/r/CFB/comments/4ee14v/hamfisted_satellite_camp_ban_creates_unintended/))
6. With the link selected as the active keyboard navigation element, hit "enter".

**Expected result**: Link opens either in new tab (preferable) or same tab, but not both
**Actual result**: Link opens in both original tab and new tab

Cause
---

There are two places where the code special-cases links on the comments page.

In [followLinkNewTab](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/keyboardNav.js#L194):

	if (!RESUtils.isPageType('comments') || selected.thing.classList.contains('link')) {`

In [toggleChildren](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/keyboardNav.js#L1010)

	if (selected.thing.classList.contains('link')) {
		// Ahh, we're not in a comment, but in the main story... that key should follow the link.

The toggleChildren case appears to be the less correct of the two; in my opinion it should do nothing if the active thing is a link and defer to other key bindings that can respect user configuration.

erikdesjardins:
Fixes #2879

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -179,21 +179,16 @@ addModule('keyboardNav', (module, moduleID) => {
 			description: 'Display parent comments.'
 		},
 		followLink: {
-			include: ['linklist', 'profile', 'search'],
+			include: ['linklist', 'profile', 'comments', 'search'],
 			value: [13, false, false, false], // enter
-			description: 'Follow link (hold shift to open it in a new tab) (link pages only)'
+			description: 'Follow link (link pages only)'
 		},
 		followLinkNewTab: {
 			include: ['linklist', 'profile', 'comments', 'search'],
 			value: [13, false, false, true], // shift-enter
 			description: 'Follow link in new tab (link pages only)',
 			callback() {
-				const selected = modules['selectedEntry'].selected();
-				if (!selected) return;
-
-				if (!RESUtils.isPageType('comments') || selected.thing.classList.contains('link')) {
-					module.followLink(true);
-				}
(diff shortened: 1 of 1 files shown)


lib/modules/keyboardNav.js
```diff
@@ -1006,28 +1001,24 @@ addModule('keyboardNav', (module, moduleID) => {
...
 		if (!selected) return;
 
-		if (selected.thing.classList.contains('link')) {
-			// Ahh, we're not in a comment, but in the main story... that key should follow the link.
-			module.followLink();
-			return;
-		} else {
-			// find out if this is a collapsed or uncollapsed view...
-			let thisToggle = selected.entry.querySelector('a.expand');
+		if (selected.thing.classList.contains('link')) return;
```
erikdesjardins: This is really the only line that was changed in this function, the rest was reindented

PR #2901, opened 2016-04-16
Imgur uses 'description' instead of 'caption' for single images

larsjohnsen:


erikdesjardins:
Huh, I'm surprised that nobody noticed this before.

For those following along at home: `caption` isn't in any of Imgur's response models.

Issue #1833, opened 2014-12-01
Use gifv inside imgur albums


andytuba:
https://www.reddit.com/r/RESissues/comments/2nytwl/bug_imgur_albums_with_gifs_between_20m_and_50m_do/

fireattack:
Just found a more serious example about this issue: https://www.reddit.com/r/oddlysatisfying/comments/4b304b/incredibly_satisfying_power_washing_videos/

You can see here, for the second images of this album, the GIF link doesn't even work (not animated) at all.

PR #2903, opened 2016-04-16
Separate gallery from image

### LINKED ISSUE #2828: Don't show album next/prev controls if album only contains 1 item

### LINKED ISSUE #2883: image rotate controls shouldn't show on failed images
@thybag 
screen shot 2016-04-13 at 4 39 17 am

`.RESImageError` is added to failed images

### LINKED ISSUE #2918: Expandos don't have the z-index they used to
So they don't overlap the sidebar or extend outside a text post body (i.e. on https://www.reddit.com/r/restests/comments/46ymr0/test_all_the_expandos/).

### LINKED ISSUE #2923: Media controls appear when hovering beside an image...
...instead of just when hovering on top of it.

larsjohnsen:
- Separated gallery from `generateImageExpando`
- Gallery requires an array `src`, which pieces it pipes into their respective generation functions
- Gallery preloads by default 2 pieces
- `buildExpando` is abstracted from `generate*Expando`, which creates a container it can put the media which is generated from `generate*` into
 - All `generate*Expando` has been partially refactored to `generate$0`, and now only requires an options-object, and return a element
 - `generateImage`, `generateAudio`, and `generateText` is now template based
 - Methods `collapse` on `expand` allows the media to determine itself how to process these events
- The placeholder is replaced with setting the parent height
 - When a media is resized, it will emit a `resize` event
 - `rotateMedia` only hooks into the event when utilized
- `conserveMemory` is refactored to work for both expandos and galleries, and by calling the media's `restore` and `unload` functions let itself determine how to go into a low-memory state
- Some simplication of `makeMediaZoomable`, `makeMediaMovable`, and `setMediaControls`

Fixes #2918, fixes #2828, fixes #2923, fixes #2883 
Lays groundwork for #1833

Tested on Firefox 46 and Chrome 50.

### DIFF
--- a/lib/core/init.js
+++ b/lib/core/init.js
@@ -1,4 +1,3 @@
-import _ from 'lodash';
 import { $ } from '../vendor';
 import * as Metadata from './metadata';
 import * as Modules from './modules';
@@ -90,8 +89,8 @@ export const loadOptions = loadDynamicOptions
 		_loadModulePrefs(),
 	]));
 
-export const addOptionsBodyClasses = loadOptions
-	.then(() => _addOptionsBodyClasses());
+export const addModuleBodyClasses = loadOptions
+	.then(() => _addModuleBodyClasses());
 
 export const always = Promise.all([loadOptions, headReady])
 	.then(() => allModules('always'));
@@ -126,15 +125,17 @@ function allModules(key, checkShouldRun = false) {
 	);
 }
 
-// Adds body classes for enabled options that have `bodyClass: true`
+// Adds body classes for modules or enabled options that have `bodyClass: true`
 // In the form `res-moduleId-optionKey` for boolean options
 // and `res-moduleId-optionKey-optionValue` for enum options
 // spaces in enum option values will be replaced with underscores
-function _addOptionsBodyClasses() {
+function _addModuleBodyClasses() {
 	for (const module of Modules.all()) {
-		if (!Modules.isRunning(module) || _.isEmpty(module.options)) continue;
+		if (!Modules.isRunning(module)) continue;
 
-		for (const [optId, opt] of Object.entries(module.options)) {
+		if (module.bodyClass) BodyClasses.add(`res-${module.moduleID}`);
+
+		for (const [optId, opt] of Object.entries(module.options || {})) {
 			if (!(opt.bodyClass && opt.value)) continue;
 
 			if (opt.type !== 'enum' && opt.type !== 'boolean') {
@@ -219,6 +220,6 @@ function setupProfiling() {
 		`beforeLoad stalled for ${diff('headReady', 'loadOptions')}`,
 		`go stalled for ${diff('bodyReady', 'beforeLoad')}`,
 		`afterLoad stalled for ${diff('contentLoaded', 'go')}`,
-		`addOptionsBodyClasses was late by ${diff('bodyStart', 'loadOptions')}`,
+		`addModuleBodyClasses was late by ${diff('bodyStart', 'loadOptions')}`,
 	].reduce((acc, line) => `${acc}<div>${line}</div>`, '')));
 }
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -217,6 +217,12 @@ img {
 		display: none !important;
 	}
 
+	.res-title,
+	.res-caption,
+	.res-credits {
+		max-width: 700px;
+	}
+
 	.res-title {
 		font-size: 13px;
 	}
@@ -232,18 +238,12 @@ img {
 }
 
 .res-gallery {
-	.res-gallery-title,
-	.res-gallery-caption {
-		max-width: 700px;
-	}
-
 	.res-gallery-title {
 		margin: 3px;
 		font-size: 15px;
 	}
 
 	.res-gallery-caption {
-		max-width: 700px;
 		margin-bottom: 3px;
 		font-size: 12px;
 	}
@@ -289,11 +289,6 @@ img {
 .res-image {
 	a {
 		display: inline-block;
-	}
-
-	.res-image-title,
-	.res-image-caption {
-		max-width: 700px;
 	}
 
 	.res-image-media {
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -72,24 +72,10 @@ img {
 	&.video-muted.expanded:hover { background-position: 0 _(11); }
 }
 
-.madeVisible {
-	clear: left;
-	display: block;
-	overflow: hidden;
-	user-select: none;
-
-	&.usertext-body {
-		user-select: text; // Allow selection if is a text expando
-	}
-
-	a.madeVisible {
-		display: inline-block;
-		overflow: hidden;
-	}
-}
-
 .res-expando-siteAttribution {
-	.madeVisible > & {
+	display: block;
+
+	.res-expando > & {
 		clear: left;
 		margin: .5em;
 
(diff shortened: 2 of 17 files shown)


lib/modules/showImages.js
```diff
@@ -863,266 +856,186 @@ addModule('showImages', (module, moduleID) => {
...
-		}
+	async function generateGallery(options) {
+		const pieces = options.src.map(src => ({ src, container: null, media: null }));
 
-		if ('caption' in imageLink) {
-			const captions = document.createElement('div');
-			captions.className = 'imgCaptions';
-			$(captions).safeHtml(imageLink.caption);
-			imgDiv.appendChild(captions);
+		// No need for gallery functionality if length === 1
```
andytuba: We may still need to show title/description for gallery/album, but nav controls should be hidden.

lib/modules/showImages.js
```diff
@@ -828,33 +829,25 @@ addModule('showImages', (module, moduleID) => {
...
-					break;
-				case 'GENERIC_EXPANDO':
-					generateGenericExpando(expandoButton, imageLink.expandoOptions);
-					break;
-				default:
-					throw new Error(`Invalid expando type: ${imageLink.type}`);
-			}
+			// Let's not bring the expando / imageLink element any further
+			const options = {
+				caption: imageLink.caption,
```
andytuba: `module.options.displayImageCaptions.value ? imageLink.caption : null` here maybe?

lib/modules/showImages.js
```diff
@@ -863,266 +856,186 @@ addModule('showImages', (module, moduleID) => {
...
-				}
-				/* falls through */
-			case 'IMAGE':
-				addImage(imgDiv, which, this);
-				break;
-			default:
-				throw new Error(`Invalid image expando type: ${imageLink.type}`);
+			// In case something appends it before it is ready
+			piece.media = document.createElement('div');
+			let element = await generateMedia(pieceOptions);
```
andytuba: yussss finally recursiveing this - gifv's in albums hooray!

erikdesjardins:
Likely fixes (or will fix) #1833 and #2828.

erikdesjardins:
[`onExpand`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/a0ca0868d40462185bf190174971a1e93b91dbf8/lib/modules/hosts/twitter.js#L22) might be getting called too early (the expando element has to be in the page for it to work).

larsjohnsen:
Squashed and rebased on the webpack branch.

andytuba:
Can we wrap [`.expando` { overrides}](https://github.com/erikdesjardins/Reddit-Enhancement-Suite/compare/webpack...larsjohnsen:seperate-gallery-image#diff-89584aec799bdede6bcda2b284599a60R226) in a "module is enabled" body class?  If showImages is disabled, I'd like to avoid interfering with related default reddit styles.

lib/css/modules/_showImages.scss
```diff
@@ -285,22 +170,152 @@ div.madeVisible {
...
+		font-size: 12px;
+	}
+
+	.res-gallery-individual-controls {
+		display: flex;
+		align-items: center;
+		margin: 4px 0 4px 10px;
+
+		&[last-piece=true] .res-gallery-next::after,
+		&[first-piece=true] .res-gallery-prev::after {
```
erikdesjardins: This class is `res-gallery-prev` in some places and `res-gallery-previous` in others

erikdesjardins:
The album size placeholder seems not to shrink when resizing:
image

erikdesjardins:
I'm not a big fan of the "Show in film strip mode" button -- I don't think it would be used very often, and it seems a bit out of place, but we'll see what others say.

erikdesjardins:
Of course, I can't forget to mention that refactoring showImages is something that we've been wanting to do for a while, thanks!

erikdesjardins:
showImages.js:987

    options.pieces = await Promise.all(options.src.map(async piece => {

`async piece =>` doesn't need to be async, you can return directly and `Promise.all` will take care of it

erikdesjardins:
showImages.js:1267

    console.error('Could not post "play" command to iframe', options.embed);

Might as well log the error `e` itself here, it could be useful (and below for "pause")

erikdesjardins:
showImages.js:1519

    const rc = e.target.getBoundingClientRect();

Nit: destructure this `const { top, left } ...`

erikdesjardins:
@larsjohnsen There are a few TODOs left, are there any more that you want to fix before this is merged?

Otherwise, this will probably be good to merge after another round of review and testing.

larsjohnsen:
Thanks for the feedback, @erikdesjardins 

The now remaining TODO is concerning rewriting of the media hosts files (perhaps by pushing all options into `expandoOptions`), but I didn't plan on doing that at this stage.

lib/modules/showImages.js
```diff
@@ -56,8 +57,8 @@ module.options = {
 	},
 	preloadImages: {
 		type: 'boolean',
-		value: false,
-		description: 'Preload gallery images for faster browsing. Beware: this is at the expense of lots of bandwidth usage.',
+		value: true,
+		description: 'Preload 2 gallery pieces for faster browsing.',
```
larsjohnsen: Modified default here since new gallery behavior limits number of preloaded pieces.

larsjohnsen: Modified default here since new gallery behavior limits number of preloaded pieces.

lib/core/init.js
```diff
@@ -126,14 +126,16 @@ function allModules(key, checkShouldRun = false) {
...
+// Adds body classes for modules or enabled options that have `bodyClass: true`
 // In the form `res-moduleId-optionKey` for boolean options
 // and `res-moduleId-optionKey-optionValue` for enum options
 // spaces in enum option values will be replaced with underscores
-function _addOptionsBodyClasses() {
+function _addModuleBodyClasses() {
 	for (const module of Modules.all()) {
 		if (!Modules.isRunning(module) || _.isEmpty(module.options)) continue;
 
+		if (module.bodyClass) BodyClasses.add(`res-${module.moduleID}`);
```
erikdesjardins: This should happen before the `_.isEmpty(module.options)` check, so it works with modules with no options

lib/core/init.js
```diff
@@ -126,15 +126,17 @@ function allModules(key, checkShouldRun = false) {
...
-function _addOptionsBodyClasses() {
+function _addModuleBodyClasses() {
 	for (const module of Modules.all()) {
-		if (!Modules.isRunning(module) || _.isEmpty(module.options)) continue;
+		if (!Modules.isRunning(module)) continue;
 
-		for (const [optId, opt] of Object.entries(module.options)) {
+		if (module.bodyClass) BodyClasses.add(`res-${module.moduleID}`);
+
+		for (const [optId, opt] of Object.entries(module.options || {})) {
```
larsjohnsen: @erikdesjardins Fixed

erikdesjardins:
showImages.js:320 Nit: extra newline

erikdesjardins:
showImages.js:376 `.filter(v => v.matches('.collapsed'))`
(as well as any other usage of Element::matches)

Edge doesn't support that so we need to use `$(v).is('.collapsed')` :(

erikdesjardins:
showImages.js:389 `this.forEach(({ expando, data }) => {`
(and below in lazyUnload)

Nit: prefer for-of loops instead of `.forEach` when it's not at the end of a `.map`, `.filter`, etc. chain

erikdesjardins:
showImages.js:946 `options.credits = html ? html.html() : null;`
(and a few times below)

`html` will always be a jQuery object here, so you don't need the ternary (I think?)

erikdesjardins:
showImages.js:957 `options.sources = $(imageLink).data('sources');`

It looks like only Onedrive and defaultAudio use this.
If it's trivial to convert them to using `expandoOptions` as well, I'd love to get rid of it; if not, don't worry about it.

erikdesjardins:
showImages:1136 `console.log(index, index >= newIndex - preloadCount && index <= newIndex + preloadCount);`

Missed debugging?

erikdesjardins:
showImages:1485 `const dragSize = p(p(e.clientX - left, 2) + p(e.clientY - top, 2), 0.5);`

Use `Math.hypot`

erikdesjardins:
showImages.js:1658 `.then(_.once(setAdvancedControls));`

`_.once` is redundant here, promise callbacks are only called once.

i.e. Promise.all will resolve (once) after both `mouseenter` and `loadedmetadata` have fired, not once for each.

erikdesjardins:
showImages.js:1663 `element.parentElement.replaceChild(generateImage({`

Nit: maybe this would be a bit more concise with `$(element).replaceWith(generateImage(...))`

larsjohnsen:
> html will always be a jQuery object here, so you don't need the ternary (I think?)

For some reason is an empty string returned instead when there's no content.

erikdesjardins:
Looks good aside from those nits; I'm testing as we speak

erikdesjardins:
> For some reason is an empty string returned instead when there's no content.

Ah! Yes, because the internal `$(this).html(...)` call looks like `$(this).html()`.

We should probably fix this on the `safeHtml` side: in `vendor/index.js`, adding `|| ''` after the `Pasteurizer.safeParseHTML(string).wrapAll('<div></div>').parent().html()` will probably do it.

lib/templates/image.mustache
```diff
@@ -0,0 +1,14 @@
+<div class="res-image">
+	{{#title}}
+	<h4 class="res-title">{{title}}</h4>
+	{{/title}}
+	{{#caption}}
+	<div class="res-image-caption">{{caption}}</div>
```
erikdesjardins: I think these classes (and those for credits, and in galleries, etc) should be `res-caption res-image-caption` instead of just `res-image-caption`; otherwise the `.res-caption` styles are never used (AFAICT)

lib/css/modules/_showImages.scss
```diff
@@ -285,22 +176,150 @@ div.madeVisible {
...
-// video-advanced
+.res-expando {
+	display: block;
+
+	.usertext-body {
+		user-select: text; // Allow selection if is a text expando
+	}
+
+	.res-expando-media {
+		display: inline-block;
```
erikdesjardins: Can you add
```scss
> * {
    display: inline-block;
}
```
immediately below this line?

(The `.res-image` and `.video-advanced` containers still end up with `block` sizing, so they fill the full width, which can interfere with middle-click scrolling -- I know, that's a *very* niche use-case)

If you think that selector is too broad and would cause problems, feel free to refine it, but I didn't notice any.

erikdesjardins:
showImage.js:61 `description: 'Preload 2 gallery pieces for faster browsing.',`

Can you make the number of pieces that get preloaded into an option, default 2?
That seems like a good default, but I'd like to have the option to load more.

erikdesjardins:
showImages.js:399 `if (testKeepLoaded(data) && media.state === mediaStates.UNLOADED) {`

`media.state` is tested, but is never assigned.
`state` is only held in a local on line 1165 (`let state = mediaStates.NONE;`)

larsjohnsen:
> Can you make the number of pieces that get preloaded into an option, default 2?

Replaced it with a new option (in lieu of migrating it), `galleryPreloadCount`.

erikdesjardins:
Woo thanks @larsjohnsen!

70966eae-e546-11e5-805b-286ef43cba17

PR #2916, opened 2016-04-25
ES6 modules overhaul

erikdesjardins:
Closes #2541.

No more globals, proper precompiled Mustache templates, even more performance (down to ~~50ms~~ 10ms on beforeLoad latency, from 200ms), reduced background page duplication, removed a lot of Babel transforms for the modern browsers (i.e. not Safari), etc. RES is also much more statically analyzable now; `eslint-plugin-import` is doing _work_.

`utils/index.js`, `environment/index.js`, and everything in `core` are good things to look at to get an idea of the changes.

~~I'll rebase and possibly squash some commits soon.~~ ~~Also, there are unnecessary blocks in most modules to avoid creating a million rebase conflicts, I'll remove them before merging.~~

Currently:
Node: good
Chrome: good
Firefox: good
Safari: almost good
Webpack: totally awesome  [ -- @andytuba ]


Ajedi32:
> +3,546 −7,120

:+1: It's pretty amazing how much ES6 can simplify things.

PR #2924, opened 2016-04-28
Fix media controls showing when not hovered

thybag:
Fixes https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2923

Selector was previously targeting `div.madeVisible` & `a.madeVisible`, rather than the a (which wraps the dragged-able element) itself..


erikdesjardins:
Already fixed in #2903 (four of us have fixed this now), maybe I shouldn't have opened an issue :sweat_smile:

erikdesjardins:
Closing in favour of #2903

PR #2925, opened 2016-04-28
Default to no max image height

thybag:
In the case of images such as https://www.reddit.com/r/TopGear/comments/4gt5r3/the_rebention/ the maxHeight restriction ends up causing them to display as extremely thin. In most cases (that I can think of at least), it seems like a better default behavior to just let the image be really long, vs limiting it and unintentionally causing the image to be unreadable.

* Change default maxHeight to 0
* Migrate any existing default settings to 0
* Only apply maxWidth/maxHeight if not set to 0.
* Update description to mention 0=unlimited


lib/modules/showImages.js
```diff
@@ -1114,8 +1114,8 @@ addModule('showImages', (module, moduleID) => {
 			image.classList.add('RESImage');
 			image.id = `RESImage-${RESUtils.randomHash()}`;
 			image.src = sourceImage.src;
-			image.style.maxWidth = `${module.options.maxWidth.value}px`;
```
andytuba: While we're in here, let's upgrade this to sanitize and validate:

```
const maxWidth = parseInt(module.options.maxWidth.values, 10);
if (maxWidth > 0) image.style.maxWidth = `${maxWidth}px`;
```

and same for maxHeight

cs462-unikent: Updated :)

thybag:
Should all be up to date again with the latest version of RES :+1:

PR #2946, opened 2016-05-07
Automated beta releases (for Chrome)

### LINKED ISSUE #2728: Deploy to Chrome beta from travis on tagged releases
Publish https://developer.chrome.com/webstore/using_webstore_api#trustedtesters

Put API key in travis settings: https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings

and while we're at it, https://docs.travis-ci.com/user/deployment/releases#Uploading-Multiple-Files

erikdesjardins:
Resolves #2728.

This does two things:

- For tags with an ~~odd minor version number~~<sup>1</sup> (to be used as beta releases), automatically builds and publishes the Chrome version to a beta listing
- ~~For *all* version number tags, builds all targets and uploads the artifacts to a new GitHub release~~ #3181

<sup>1: Actually, we might as well release all versions to the beta channel, so beta users don't have to switch back to the main extension for normal releases.</sup>

**Todo**

- [x] [Create beta app listing in Chrome webstore](https://chrome.google.com/webstore/developer/dashboard) (use [the package from 4.7.1](https://github.com/honestbleeps/Reddit-Enhancement-Suite/releases/tag/v4.7.1) and this [beta icon](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/images/beta128.png)) and private visiblity (select the google group in the dropdown and make sure you do this before first publish)
- [x] [Get Chrome webstore API OAuth tokens](https://erikdesjardins.github.io/chrome-deploy-key/) (client id, client secret, refresh token) and add as environment variables with [these names](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2946/files#diff-7bce112a2436361628559f89e56eae6bR23) to [Travis CI](https://travis-ci.org/honestbleeps/Reddit-Enhancement-Suite/settings)
- [x] add beta extension id to [deploy.js](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2946/files#diff-7bce112a2436361628559f89e56eae6bR26)
- [x] Write beta instructions / disclaimers for store listing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,9 @@ deploy:
     on:
       tags: true
       node: 6
+  - provider: script
+    script: node build/deploy.js
+    skip_cleanup: true
+    on:
+      tags: true
+      node: 6
--- a/build/deploy.js
+++ b/build/deploy.js
@@ -1,0 +1,53 @@
+/* eslint-disable import/no-commonjs, import/no-nodejs-modules */
+
+const fs = require('fs');
+const path = require('path');
+const deploy = require('chrome-extension-deploy');
+const { version } = require('../package.json');
+const isBetaVersion = require('./isBetaVersion');
+
+if (isBetaVersion(version)) {
+	console.log(`Deploying ${version} beta release...`);
+
+	deployChromeBeta();
+} else {
+	console.log(`Deploying ${version} stable release...`);
+
+	deployChromeBeta();
+	deployChromeStable();
+}
+
+function deployChromeBeta() {
+	console.log('Deploying Chrome beta...');
+
+	deploy({
+		clientId: process.env.CHROME_CLIENT_ID,
+		clientSecret: process.env.CHROME_CLIENT_SECRET,
+		refreshToken: process.env.CHROME_REFRESH_TOKEN,
+		id: 'flhpapomijliefifkkeepedibpmibbpo',
+		zip: fs.readFileSync(path.join(__dirname, '../dist/zip/chrome.zip')),
+		to: deploy.TRUSTED_TESTERS,
+	}).then(() => {
+		console.log('Chrome beta deployment complete!');
+	}, err => {
+		console.error('Chrome beta failed:', err);
+		process.exitCode = 1;
+	});
+}
+
+function deployChromeStable() {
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -12,3 +12,28 @@ script:
   - node dist/node/loader.entry.js
 after_success:
   - npm run report-coverage
+branches:
+  only:
+    - master
+    - /^v[0-9]/
```
Ajedi32: Pretty sure this is actually restricting all builds/tests to only those branches, not just deploys, right?

erikdesjardins: Yeah, but it won't effectively change anything because it only applies to branches and tags in this repo (PRs are always built).

Mainly, it's the simplest way to avoid building (and then trying to deploy) tags that don't look like version numbers.

Ajedi32:
Seems like a reasonable idea. I've already been doing something similar to this by just using a version of RES that I built myself on my computer.

Maybe this would give more non-technical users a chance to test upcoming changes though, which could help avoid situations like in the previous release where few people were aware of the upcoming changes to night mode until after the release. Getting early feedback from users is a good thing.

> For tags with an odd minor version number (to be used as beta releases)

Is there a reason you're suggesting using odd version numbers here as beta releases instead of the SemVer convention of appending `-beta.0`, `-beta.1`, etc to the version number? http://semver.org/#spec-item-9

andytuba:
> Is there a reason you're suggesting using odd version numbers here as beta releases instead of the SemVer convention of appending -beta.0, -beta.1, etc to the version number?

Yes, browsers only allow numbers and periods in version numbers.

.travis.yml
```diff
@@ -10,3 +10,28 @@ script:
   - node dist/node/loader.entry.js
 after_success:
   - npm run report-coverage
+branches:
+  only:
+    - master
```
andytuba: Deploy from master too?

andytuba:
Store listing / disclaimer / instructions text: https://docs.google.com/document/d/1uEyNRkB4Vg8NSgXt_tGDJ-Seth8rAsI3zC26QC7rk18/edit?usp=sharing

PR #2975, opened 2016-05-15
Adds Youtube parameters more dynamically

corylulu:
A modification to the previous merge to adds Youtube parameters more dynamically and elegantly, making it simpler to make additions or changes down the line, could also be used as a template for other modules.

### DIFF
--- a/lib/modules/hosts/youtube.js
+++ b/lib/modules/hosts/youtube.js
@@ -13,45 +13,33 @@ export default {
 	},
 	handleLink(elem, [, hash]) {
 		let src = `https://www.youtube.com/embed/${hash}?enablejsapi=1&enablecastapi=1`;
-		const paramResults = getUrlParams(elem.href);
-		for (const paramName in paramResults) {
-			const paramValue = paramResults[paramName];
-			// Add accepted parameters to the embeded url.
-			switch (paramName) {
-				// Embeded player doesn't support segmented timestamps (e.g '&t=8m14s'), needs conversion.
-				case 't':
-					let starttime = 0;
-					const timeBlocks = { h: 3600, m: 60, s: 1 };
-					const timeRe = /[0-9]+[hms]/ig;
-					// Get each segment e.g. 8m and calculate its value in seconds
-					const timeMatch = paramValue.match(timeRe);
+		const params = getUrlParams(elem.href);
+		if (params.t) {
+			let starttime = 0;
+			const timeBlocks = { h: 3600, m: 60, s: 1 };
+			const timeRe = /[0-9]+[hms]/ig;
+			// Get each segment e.g. 8m and calculate its value in seconds
+			const timeMatch = params.t.match(timeRe);
 
-					if (timeMatch) {
-						timeMatch.forEach(ts => {
-							const unit = timeBlocks[ts.charAt(ts.length - 1)];
-							const amount = parseInt(ts.slice(0, -1), 10);
-							// Add each unit to starttime
-							starttime += unit * amount;
-						});
-					} else {
-						// support direct timestamp e.g. t=200
-						starttime = parseInt(paramValue, 10);
-						if (isNaN(starttime)) starttime = 0;
-					}
-					src = `${src}&start='${starttime}`;
-					break;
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/youtube.js
```diff
@@ -11,47 +11,50 @@ export default {
...
-					// Add each unit to starttime
-					starttime += unit * amount;
-				});
-			} else {
-				// support direct timestamp e.g. t=200
-				starttime = parseInt(starttimeResult[1], 10);
-				if (isNaN(starttime)) starttime = 0;
+		const findParamsRe = /[\?&]([\w-]+)=([\w-]+)/g;
+		let paramResults;
+		while ((paramResults = findParamsRe.exec(elem.href)) !== null) {
```
erikdesjardins: `getUrlParams` can be used to do this; you can import it as `import { getUrlParams } from '../../utils';`

corylulu:
@erikdesjardins Okay, I've made that change.

lib/modules/hosts/youtube.js
```diff
@@ -11,49 +12,48 @@ export default {
...
-		const starttimeResult = starttimeRe.exec(elem.href);
-		// End time is always in integer format, no need to parse
-		const endtimeRe = /[&\?]end=([0-9]+)/i;
-		const endtimeResult = endtimeRe.exec(elem.href);
-		// Finds playlist id if included
-		const playlistRe = /[&\?]list=([\w\-_]+)/i;
-		const playlistResult = playlistRe.exec(elem.href);
+		let src = `https://www.youtube.com/embed/${hash}?enablejsapi=1&enablecastapi=1`;
+		const paramResults = getUrlParams(elem.href);
+		for (const paramName in paramResults) {
```
erikdesjardins: Seems like `if (params.t)`, `if (params.end)`, etc. would be simpler than looping with a switch

corylulu: @erikdesjardins Oh, didn't think to do that. I'm only used to working in strong typed languages. Not a fan of doing things weakly typed and find switch cases similar than if cases in cases like this. Curious to know if there is actually any performance benefit, though. But I'll make the change.

corylulu:
@erikdesjardins Does this look good to you now?

lib/modules/hosts/youtube.js
erikdesjardins: I bet the loop is actually slower :stuck_out_tongue_winking_eye:  since V8 can optimize direct property loads like `params.t` to loads by an offset, as if it were a struct (well, ideally -- depending on object shape, inline cache state, etc.), and collecting all of an object's keys for for..in is (relatively) expensive. But performance doesn't matter here regardless.

Testing for property existence/truthiness with `if (param.t)` is very common in JS; tbh I don't think I've ever seen a for..in loop used like that before. :sweat_smile:

I agree with the duplication but we'll worry about it when more than two cases are duplicated.

lib/modules/hosts/youtube.js
```diff
@@ -37,21 +30,16 @@ export default {
...
-
-		if (endtimeResult) {
-			// `&version=3` seems to be required for end time to work
-			src += `&end=${endtimeResult[1]}&version=3`;
-		}
-
-		if (playlistResult) {
-			// appends playlist id if found
-			src += `&list=${playlistResult[1]}`;
+			src = `${src}&start='${starttime}`;
```
erikdesjardins: `+=` was fine here; since the RHS is a string you're guaranteed to get a string (and IMO it's clear enough to read; if we were appending two vars alone then it would be clearer to do ``` `${foo}${bar}` ```)

lib/modules/hosts/youtube.js
```diff
@@ -37,21 +30,16 @@ export default {
...
+		} else if (params.end) {
```
erikdesjardins: These should be separate `if` statements since they could all potentially be true

lib/modules/hosts/youtube.js
corylulu: @erikdesjardins didn't pass the checks, which is why I changed it.

PR #2988, opened 2016-05-18
ShowImages resizer: Reduce Firefox lag

larsjohnsen:
Limits `dragMedia` to at most one execution per frame.


larsjohnsen:
Updated to use `frameDebounce` from #2989

erikdesjardins:
Thanks!

PR #2990, opened 2016-05-18
Update #2906: Captions position

erikdesjardins:
Closes #2906.

Implemented with flexbox order, since there isn't a simple way to reorder elements in a moustache template (AFAIK).

~~Changed the default to `allAbove`, since the status quo is king (people hate change), and the problem it solves is very niche.~~
[after discussion in IRC] Keeps `titleAbove`; moves the option down next to `mediaControls` for discoverability. Hopefully that'll be good enough.

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -205,10 +205,6 @@ img {
 		.selftext.expanded ~ * & {
 			max-width: 100%;
 		}
-
-		> * {
-			display: inline-block;
-		}
 	}
 
 	.res-expando-independent {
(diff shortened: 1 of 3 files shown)


andytuba:
I'm leaning towards defaulting to `titleAbove` since that's how Imgur does it and, apparently, Tumblr.

lib/css/modules/_showImages.scss
```diff
@@ -486,3 +484,41 @@ img {
 		background: none;
 	}
 }
+
+.res-showImages-captionsPosition {
+	@at-root {
+		.res-image,
+		.res-gallery {
+			display: inline-flex;
+			flex-direction: column;
```
larsjohnsen: Appears that this breaks gallery filmstrip-mode.

PR #2991, opened 2016-05-18
ShowImages: Hide MediaControls when resizing

larsjohnsen:
MediaControls can interfere with resizing (by triggering a `mouseout` event on hover).

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -186,6 +186,7 @@ img {
 		transition: opacity 300ms;
 	}
 
+	body.res-media-resizing &,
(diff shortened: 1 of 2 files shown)


lib/css/modules/_showImages.scss
```diff
@@ -186,6 +186,7 @@ img {
 		transition: opacity 300ms;
 	}
 
+	@at-root body.res-media-resizing &,
```
erikdesjardins: I *think* `@at-root` is redundant here, since the selector includes `&`.

larsjohnsen: You're right. Updated.

PR #2993, opened 2016-05-18
add settings link for media site module to site attribution

andytuba:
screen shot 2016-05-18 at 2 37 59 pm
site settings link


erikdesjardins:
Spacing is a little tight for me, strangely (in both FF and Chrome on Windows):

image

thybag:
Just wondering if its worth also removing the settings icon from the "media control bar", given its moving to the site attribution etc?

andytuba:
I'd rather keep both since the attribution link goes directly to the particular site module settings while the control bar links to its particular setting.. and because users are myopic.

lib/core/migrate.js
```diff
@@ -262,6 +262,54 @@ const migrations = [
 		async go() {
 			await migrators.generic.updateOption('singleClick', 'hideLEC', false, true);
 		},
+	}, {
+		versionNumber: '4.7.0-showImages-siteModuleIDs',
+		async go() {
+			await migrators.generic.moveOption('styleTweaks', 'display uploadly', 'styleTweaks', 'display_uploadly');
```
erikdesjardins: This should be `showImages` right?

lib/templates/searchResultOption.mustache
```diff
@@ -1,10 +1,11 @@
 <div class="SearchRES-result-header">
     <span class="SearchRES-result-title">{{title}}</span>
-				<span class="SearchRES-breadcrumb">RES settings console
-					&gt; {{category}}
-                    &gt; {{moduleName}} ({{moduleID}})
-					{{#optionKey}} &gt; {{optionKey}}{{/optionKey}}
-				</span>
+	<span class="SearchRES-breadcrumb">RES settings console
+		&gt; {{category}}
+        &gt; {{moduleName}} ({{moduleID}})
```
erikdesjardins: Spacing is a little off here

lib/core/migrate.js
andytuba: yyyyyyyep

PR #2994, opened 2016-05-19
showImages tweaks

erikdesjardins:


larsjohnsen:
Removing `max-width` from `.res-captions` exposes an issue in the gallery where the captions may horizontally overflow when the piece is wider than the background. This expando width is set in `.res-expando-media` by `textMaxWidth`, but I am unsure on how to propagate it to the gallery pieces.

larsjohnsen:
Solution to issue mentioned above:
`               .res-expando-media > *,
               .res-gallery-pieces,
               .res-gallery-pieces > * {
                       width: inherit;
               }
`

PR #2996, opened 2016-05-21
ShowImages: Style fixes

larsjohnsen:
Propagates the width of the expando-parent into the expandobox, so that text won't horizontally overflow.


erikdesjardins:
You're basically a CSS wizard. Thanks!

PR #2997, opened 2016-05-21
Filter visited links

### LINKED ISSUE #2456: Ability to toggle "hide visited links"
Simple enough to implement (hopefully). Add a toggle in the settings in reddit to hide all links with the ':visited' pseudo-selector.

With this, we can easily see only content we have not viewed yet in subs we frequent often. For browsers that have history sync, this will also automatically carry over to all other browsers you have RES installed on.

larsjohnsen:
Useful for browsing /top.

Creates a button for removing links already visited, similar to "view images"

Closes #2456

### DIFF
--- a/chrome/background.entry.js
+++ b/chrome/background.entry.js
@@ -148,6 +148,12 @@ addListener('addURLToHistory', url => {
 	chrome.history.addUrl({ url });
 });
 
+addListener('isURLVisited', url =>
+	new Promise(async resolve => {
+		resolve((await apiToPromise(chrome.history.getVisits)({ url })).length > 0);
+	})
+);
+
 chrome.pageAction.onClicked.addListener(({ id: tabId }) =>
 	sendMessage('pageActionClick', undefined, { tabId })
 );
--- a/chrome/background.entry.js
+++ b/chrome/background.entry.js
@@ -148,10 +148,8 @@ addListener('addURLToHistory', url => {
 	chrome.history.addUrl({ url });
 });
 
-addListener('isURLVisited', url =>
-	new Promise(async resolve => {
-		resolve((await apiToPromise(chrome.history.getVisits)({ url })).length > 0);
-	})
+addListener('isURLVisited', async url =>
+	(await apiToPromise(chrome.history.getVisits)({ url })).length > 0
 );
 
 chrome.pageAction.onClicked.addListener(({ id: tabId }) =>
--- a/firefox/background.entry.js
+++ b/firefox/background.entry.js
@@ -314,6 +314,15 @@ addListener('addURLToHistory', url => {
 	});
 });
 
+addListener('isURLVisited', url =>
+	new Promise(resolve => {
+		/* eslint no-unused-vars: ["error", { "args": "none" }] */
+		historyService.isURIVisited(ioService.newURI(url, undefined, undefined), (aURI, visited) => {
+			resolve(visited);
+		});
+	})
+);
+
 addListener('multicast', (request, worker) => {
 	const isPrivate = priv.isPrivate(worker);
 	return Promise.all(
--- a/firefox/background.entry.js
+++ b/firefox/background.entry.js
@@ -316,7 +316,6 @@ addListener('addURLToHistory', url => {
 
 addListener('isURLVisited', url =>
 	new Promise(resolve => {
-		/* eslint no-unused-vars: ["error", { "args": "none" }] */
 		historyService.isURIVisited(ioService.newURI(url, undefined, undefined), (aURI, visited) => {
 			resolve(visited);
 		});
--- a/lib/environment/history.js
+++ b/lib/environment/history.js
@@ -11,3 +11,11 @@ export async function addURLToHistory(url) {
 
 	await sendMessage('addURLToHistory', url);
 }
+
+/**
+ * @param {...string} url
+ * @returns {Promise<boolean, *>}
+ */
+export function isURLVisited(url) {
+	return sendMessage('isURLVisited', url);
+}
--- a/lib/modules/filteReddit.js
+++ b/lib/modules/filteReddit.js
@@ -7,6 +7,7 @@ import {
 	currentMultireddit,
 	currentSubreddit,
 	currentUserProfile,
+	CreateElement,
 	indexOptionTable,
 	isCurrentMultireddit,
 	isCurrentSubreddit,
@@ -15,6 +16,7 @@ import {
 	string,
 	watchForElement,
 } from '../utils';
+import { isURLVisited } from '../environment';
 import * as CommandLine from './commandLine';
 import * as CustomToggles from './customToggles';
 import * as Menu from './menu';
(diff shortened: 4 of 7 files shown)


lib/modules/showImages.js
```diff
@@ -448,26 +455,35 @@ function createImageButtons() {
 		$mainMenuUL.css('display', 'inline-block'); // Override dashboard's subreddit style.
 	}
 
+	function addButton(options, func) {
```
larsjohnsen: Perhaps move to ../utils/createElement as 'createTabMenuButton'

lib/modules/filteReddit.js
```diff
@@ -746,6 +755,17 @@ module.go = function() {
 			toggleNsfwFilter(toggle, true);
 		}
 	);
+
+	if (isPageType('linklist')) {
+		if (module.options.showRemoveVisitedTab.value && process.env.BUILD_TARGET !== 'safari') {
+			CreateElement.tabMenuItem({
```
andytuba: How do you feel about putting this in the floating island with the "back to top" button instead of as a tab menu item? It's been a longstanding regret to keep the "view images" button in the tab bar, especially for power-users who end up with lots of other tabs.

larsjohnsen: I don't think it's well placed there, as one would then have to scroll before initiating the option. 

Perhaps as an alternative could be to place a `filter by` line at top of the linklist, similar to 'navigate by'.
image

andytuba: I like the idea of a line specifically for filter toggles. Add-on feature: opting to show custom toggles on that line.

andytuba: > one would then have to scroll before initiating the option.

Maybe that "visibleAfterScroll" floating island should start below the header, rather than hidden, then pin to top when the header is out of view.

chrome/background.entry.js
```diff
@@ -162,6 +162,12 @@ addListener('addURLToHistory', url => {
 	chrome.history.addUrl({ url });
 });
 
+addListener('getVisitedURLs', async urls =>
+	(await Promise.all(
+		urls.map(async url => [url, await apiToPromise(chrome.history.getVisits)({ url })])
+	)).filter(v => v[1].length).map(v => v[0])
```
erikdesjardins: If this is returning multiple results, I'd prefer a map from url -> visited state instead of an array of visited urls.

Although, since it's only ever used to check one URL at a time (AFAICT), you could just accept a single URL and return a boolean, which would simplify the implementation a bit. We can always expand it to accept multiple URLs in the future if necessary.

lib/environment/history.js
```diff
@@ -11,3 +11,11 @@ export async function addURLToHistory(url) {
 
 	await sendMessage('addURLToHistory', url);
 }
+
+/**
+ * @param {...string} url
+ * @returns {Promise<string[], *>}
+ */
+export async function getVisitedURLs(...urls) {
+	return await sendMessage('getVisitedURLs', urls);
```
erikdesjardins: `async`/`await` here are redundant, we can just return the promise returned by sendMessage directly.

lib/modules/filteReddit.js
```diff
@@ -64,6 +67,11 @@ module.options = {
 		value: false,
 		description: 'Don\'t filter anything on users\' profile pages',
 	},
+	showRemoveVisitedTab: {
+		type: 'boolean',
+		value: true,
+		description: `${process.env.BUILD_TARGET === 'safari' ? '(Not available on Safari) ' : ''}Show a \'remove visited\' tab at the top of each subreddit, to easily toggle whether to remove visited links.`,
```
erikdesjardins: Use `value: process.env.BUILD_TARGET !== 'safari'` and `noconfig: process.env.BUILD_TARGET === 'safari'`

safari/environment.js
```diff
@@ -48,5 +48,8 @@ addInterceptor('deleteCookies', cookies => {
 // The iframe hack doesn't work anymore, so Safari has no way to add urls to history
 addInterceptor('addURLToHistory', () => {});
 
+// XXX No point in implementing this when addURLToHistory is not implemented
+addInterceptor('getVisitedURLs', () => {});
```
erikdesjardins: Should return a sensible default instead of undefined.

lib/modules/filteReddit.js
```diff
@@ -787,30 +808,31 @@ function scanEntries(ele) {
...
 			const postFlair = thing.getPostFlairText();
+			const postLink = thing.getPostLink().href;
 
 			let filtered = executeCustomFilters(thing);
 
 			currSub = currentSubreddit() ? currentSubreddit().toLowerCase() : null;
 			postSubreddit = thing.getSubreddit() || currSub;
 
 			if (!filtered) filtered = filterTitle(postTitle, postSubreddit);
+			if (!filtered) filtered = removeVisitedActive && await filterVisited(postLink);
```
erikdesjardins: I am a bit concerned about having a potentially slow async operation in the middle of filtering, since right now filters can run before first paint, which I doubt could happen with `filterVisited` enabled.

Maybe do `filterVisited` in a second `things::forEachChunked` pass directly after this one, so the fast filters are always fast?

andytuba:
How about renaming this "Filtering visited links" to avoid conflating with reddit's "hide post" action?

firefox/background.entry.js
```diff
@@ -314,6 +314,15 @@ addListener('addURLToHistory', url => {
 	});
 });
 
+addListener('isURLVisited', url =>
+	new Promise(resolve => {
+		/* eslint no-unused-vars: ["error", { "args": "none" }] */
```
erikdesjardins: Are you sure this is needed? ESLint should allow unused args as long as they're followed by used args

chrome/background.entry.js
```diff
@@ -146,6 +146,12 @@ addListener('addURLToHistory', url => {
 	chrome.history.addUrl({ url });
 });
 
+addListener('isURLVisited', url =>
```
erikdesjardins:
```js
addListener('isURLVisited', async url =>
	(await apiToPromise(chrome.history.getVisits)({ url })).length > 0
);
```

or maybe

```js
addListener('isURLVisited', async url => {
	const visits = await apiToPromise(chrome.history.getVisits)({ url });
	return visits.length > 0;
});
```

safari/environment.js
```diff
@@ -48,5 +48,8 @@ addInterceptor('deleteCookies', cookies => {
 // The iframe hack doesn't work anymore, so Safari has no way to add urls to history
 addInterceptor('addURLToHistory', () => {});
 
+// XXX No point in implementing this when addURLToHistory is not implemented
+addInterceptor('isURLVisited', () => undefined);
```
erikdesjardins: `false` instead of `undefined` (so minor, I know, sorry)

I guess `undefined` would represent "indeterminate", but I don't really want to allow the possibility of `if (visited === false) ... else if (visited === undefined)` (prefer to throw if we need to handle an unimplemented API, probably).

PR #2998, opened 2016-05-22
ShowImages: Rewrite expando

### LINKED ISSUE #1244: "Duplicate" image after edit comment containing image
If you post a comment which contains an image, then come back later and edit it, then the image expando is marked as "duplicate, can not expand".

### LINKED ISSUE #1835: Keyboard nav link # should ignore duplicate links
https://www.reddit.com/r/RESissues/comments/2o3a1g/bug_reddit_enhancement_suite_will_not_ignore/

### LINKED ISSUE #2724: mediaBrowseMode breaks when Never Ending Reddit loads next page
This may be related to #2302, but I don't tend to use the "View all" button. To reproduce this bug:

1. Ensure Keyboard Navigation > `mediaBrowseMode` and Never Ending Reddit > `autoLoad` are both enabled.
2. Load up any list of posts with [lots of images](https://www.reddit.com/r/images) (this happens with any expandable media)
3. Give focus to the top list item
4. Press X to expand the image/media attached
5. Press J once to verify that `mediaBrowseMode` expands the image for the next post
6. Hold down the J key until Never Ending Reddit loads the next page of posts

If the timing works out, the next page will load but the images will cease expanding on the next page.

This may be due to a race condition, because some scenarios carry the media expansion into the next page. I see this with normal usage (i.e. not when holding the J key down), which may come down to images that affect RES detecting a bottom edge for scrolling, loading that next page of posts, and immediately giving focus to the next list item before the attached media can be detected. By the time the media is detected, the focus was already granted to the list item so the `mediaBrowseMode` check has already occurred.

### LINKED ISSUE #2783: Clicking media expander before the page loads results in duplicates
Basically, if you go into a comments thread and click the media expander button on the OP for a post that has some kind of media, the media will load and then once the page finishes loading, the expander turns back to the non-expanded state and can be expanded again, creating a duplicate

browser: chrome

res version: 4.6.0

### LINKED ISSUE #3055: Expando in newly posted/edited comment opens and autoplays

larsjohnsen:
- New class `Expando` containing much of the functionality of `expandoButton` that was scattered around.
- Swaps the original expando with that of the duplicate instead of moving viewport.
- Replaces duplicate-expando text with a sepia colorored icon.
- Adjusts expando icons in text, so they're aligned with the text.
- Quicker generation of the expando button, and a new load icon is used if the media type cannot be instantaneously determined.
- Removes customImageFilters (as defined by subreddits; may be reintroduced in filterline)
- Misc fixes (memory leak, lazyUnload, keyboard toggling, …)

Fixes #1244 
Fixes #1835 (since duplicates behaves as independent expandos)
Fixes #2783 (does not create expando if native is already expanded; also, native expando is removed earlier)
Mostly fixes #2724 (Remembers expandState even if expando is not initialized, but requires that `findAllImages` has evaluated it)
Edit (erikdesjardins): Fixes #3055


andytuba:
Sounds like a much more useful design.

lib/modules/showImages.js
```diff
@@ -644,82 +648,118 @@ function domainToModuleName(hostname) {
...
+		'defaultVideo',
+		'defaultAudio',
+		domainToModuleName(imageLink.hostname),
+	]
+		.map(modId => domainModuleMap[modId])
+		.find(mod => mod && (detectResult = mod.detect(imageLink)));
+
+	if (!detectResult || !siteModule) return;
+
+	if (module.options.keepNativeExpando.value) {
```
erikdesjardins: No longer exists (also used below)

lib/modules/showImages.js
```diff
@@ -644,82 +648,118 @@ function domainToModuleName(hostname) {
...
-			}
+		try {
+			const options = await siteModule.handleLink(imageLink.href, detectResult);
+			createExpandoButton(expandoButton, options);
+			updateViewImagesCount();
+		} catch (e) {
+			console.error(`showImages: error detecting image expando for ${imageLink.href}`);
+			console.error(e);
+
+			imageList.splice(imageList.indexOf(expandoButton));
```
erikdesjardins: `_.remove(imageList, expandoButton)`

lib/modules/showImages.js
larsjohnsen: Most of this code is replaced in the latest commit.

erikdesjardins: Maaybe I should have read all of the commits before commenting :sweat_smile:

lib/modules/showImages.js
```diff
@@ -644,286 +636,366 @@ function domainToModuleName(hostname) {
...
+			IMAGE: 'image',
+			GALLERY: 'image gallery',
+			TEXT: 'selftext',
+			VIDEO: 'video',
+			IFRAME: 'video',
+			AUDIO: 'video', // yes, still class "video", that's what reddit uses.
+			NOEMBED: 'video',
+			GENERIC_EXPANDO: 'selftext',
+		}[options.type];
+		this.button.className = `\
```
erikdesjardins: Just have one string per line and concat or `[].join()` them

lib/modules/showImages.js
```diff
@@ -644,286 +636,366 @@ function domainToModuleName(hostname) {
...
+			type: 'json',
+			url: '//upload.gfycat.com/transcodeRelease',
+			data: { fetchUrl: options.src },
+			cacheFor: DAY,
+		});
+
+		if (!info.gfyName) throw new Error();
+
+		return {
+			options: await gfycat.handleLink('', [undefined, null, info]),
```
erikdesjardins: Maybe pass `info` as a third argument; I don't think it makes much sense for it to show up in the "regex results array"

lib/modules/showImages.js
```diff
@@ -644,286 +636,366 @@ function domainToModuleName(hostname) {
...
-	// don't reveal images for invisible buttons (offsetParent is a cheaper way of checking
-	// visibility than jquery's .is(':visible'))
-	if ((!expandoButton) || (!expandoButton.offsetParent)) {
-		return false;
-	}
-	// showhide = false means hide, true means show!
+		let reveal = false;
+		if (module.options.autoExpandSelfText.value && this.thing.isSelfPost() && this.inText) {
+			// Expand the all autoExpand-expandos in additon to the first non-autoExpand expando
+			reveal = reveal || this.options.autoExpand ||
```
erikdesjardins: `reveal = reveal || this.options...` -> `reveal = this.options...`

(it's always false here)

lib/modules/showImages.js
```diff
@@ -644,286 +636,366 @@ function domainToModuleName(hostname) {
...
+	}
 
-		if (!siteModule) return;
+	expando.$button.insertAfter(!inText && thing.getTitleElement() ?
+		mediaLink.parentElement : mediaLink);
 
-		elem.site = siteModule.moduleID;
+	try {
+		if (module.options.ignoreDuplicates.value && isDuplicate) {
+			expando.buildDuplicate(await promise);
```
erikdesjardins: High-level comment: ideally, this promise would resolve to something like `{ options, siteModule }` so it can always be reused by a duplicate, regardless of whether or not we're ignoring duplicates.

I believe that would also simplify a lot of the logic in the second half of this function, and we then wouldn't have to expose `resolve`/`reject` from a bare promise (which is pretty ugly).

larsjohnsen: I made an attempt at cleaning it up now. I also removed the `ignoreDuplicates` option, since the UX issues associated with the previous solution are gone.

lib/modules/keyboardNav.js
```diff
@@ -1078,16 +1073,29 @@ module.toggleChildren = function() {
 	click(thisToggle);
 };
 
-module.toggleExpando = function() {
-	const selected = SelectedEntry.selectedThing();
-	if (!selected) return;
+module.toggleExpandos = function(scrollOnExpanding) {
```
erikdesjardins: I think this could be a local function instead of a property on `module`; I probably missed it while converting the modules

lib/modules/keyboardNav.js
```diff
@@ -1078,16 +1073,29 @@ module.toggleChildren = function() {
...
-			scrollToElement(selected.thing, { scrollStyle: 'top' });
+	// If any open expandos exists within thing, collapse all
+	// Else, expand all
+	const openExpandos = expandos.filter(v => v.open);
+
+	if (openExpandos.length) {
+		for (const expando of openExpandos) expando.collapse();
+	} else {
+		for (const expando of expandos) {
+			if (expando.constructor.name !== 'Expando' || expando::ShowImages.isExpandWanted(thing.thing)) {
```
erikdesjardins: I'd prefer importing `Expando` and using `instanceof`, or checking for `expando.expand` (if that's equivalent)

lib/modules/showImages.js
```diff
@@ -638,292 +569,244 @@ function domainToModuleName(hostname) {
...
 
-	const options = {
-		loadHeight: 20, // gives an indication that the expando is opened
-	};
+	updateAutoExpandCount();
+}
 
-	const expando = expandoButton.expandoBox = $(expandoTemplate(options))[0];
-	const mediaContainer = expando.querySelector('.res-expando-media');
+function retrieveExpandoOptions(thing, { siteModule, detectResult, mediaLink, href }) {
```
erikdesjardins:
```js
const retrieveExpandoOptions = _.memoize(
    async (...) => {...},
    (thing, { href }) => href
);
```

And then you can remove `expandoPromises`

lib/modules/showImages.js
```diff
@@ -638,292 +569,244 @@ function domainToModuleName(hostname) {
...
+	return new Promise(async (resolve, reject) => {
```
erikdesjardins: This promise is probably not necessary if you make `retrieveExpandoOptions` async...

lib/modules/showImages.js
```diff
@@ -638,292 +569,244 @@ function domainToModuleName(hostname) {
...
+				href, // Since mediaOptions.href may be overwritten
+				generateMedia: () => generateMedia(mediaOptions, siteModule),
+				mediaOptions,
+				onMediaAttach() {
+					trackLoad();
+					if (mediaOptions.onAttach) mediaOptions.onAttach();
+				},
+			};
+
+			resolve(expandoOptions);
```
erikdesjardins: ...because you can return this...

lib/modules/showImages.js
```diff
@@ -638,292 +569,244 @@ function domainToModuleName(hostname) {
...
+		} catch (e) {
+			reject(e);
```
erikdesjardins: ...and remove the try-catch, which will behave identically to this.

lib/modules/showImages.js
```diff
@@ -1146,22 +1035,24 @@ function generateIframe(options) {
...
+				console.error('Could not post "pause" command to iframe', options.embed, e);
 			}
 		}
 		return removeInstead;
 	};
 
 	keepMediaVisible(iframeNode);
 
-	return iframeNode;
+	element.ready = new Promise(resolve => { resolve(); });
```
erikdesjardins: `new Promise(resolve => { resolve(); })` -> `Promise.resolve()`

larsjohnsen:
Thanks, @erikdesjardins. Much appreciated.

lib/modules/keyboardNav.js
```diff
@@ -1078,19 +1074,36 @@ module.toggleChildren = function() {
 	click(thisToggle);
 };
 
-module.toggleExpando = function() {
-	const selected = SelectedEntry.selectedThing();
-	if (!selected) return;
+function toggleExpandos(scrollOnExpanding) {
```
erikdesjardins: This function might make more sense in showImages.js, perhaps with the signature `toggleExpandos(thing, scrollOnExpanding)` or `toggleExpandos(expandos, scrollOnExpanding)`. But I don't feel strongly about this.

lib/modules/showImages.js
```diff
@@ -638,292 +572,220 @@ function domainToModuleName(hostname) {
...
+		const info = await ajax({
+			type: 'json',
+			url: '//upload.gfycat.com/transcodeRelease',
+			data: { fetchUrl: options.src },
+			cacheFor: DAY,
+		});
 
-		if (detectResult && !elem.NSFW) {
-			imagesRevealed[href] = dupeAnchors;
+		if (!info.gfyName) throw new Error();
```
erikdesjardins: Might as well give this error a message

lib/modules/showImages.js
```diff
@@ -638,292 +572,220 @@ function domainToModuleName(hostname) {
...
-		textFrag.setAttribute('class', 'RESdupeimg');
-		removeRedditExpandoButton(elem);
-		$(textFrag).html(`<a class="noKeyNav" href="#img${parseInt(imagesRevealed[href], 10)}" title="click to scroll to original">[RES ignored duplicate link]</a>`);
-		$(elem).after(textFrag);
+		return {
+			options: await gfycat.handleLink('', [], info),
+			siteModule: gfycat,
+		};
+	} catch (e) {
+		throw Error('Could not convert gif to video', options.src);
```
erikdesjardins: I don't think it's standard for the `Error` constructor to accept multiple arguments, so concat `options.src` with the message.

Also minor nit: `s/throw Error/throw new Error/`

lib/utils/Thing.js
```diff
@@ -202,12 +204,36 @@ export default class Thing {
...
+			};
+		}
+
+		return expando;
+	}
+
+	getTextExpandos() {
+		let textExpandos = Array.from(this.entry.querySelectorAll(Thing.expandoSelector))
+			.map(v => expandos.get(v));
+		textExpandos = _.remove(textExpandos, this.getEntryExpando());
```
erikdesjardins: `_.remove` operates in-place and returns an array of removed elements (i.e. just `this.getEntryExpando()`, if present); I'm not sure that that's what's intended here -- I would assume that this is intended to be:

```js
const entryExpando = this.getEntryExpando();
return Array.from(this.entry.querySelectorAll(Thing.expandoSelector))
    .map(v => expandos.get(v))
    .filter(v => v && v !== entryExpando);
```

lib/utils/Thing.js
```diff
@@ -202,12 +204,36 @@ export default class Thing {
...
+		return textExpandos.filter(v => v);
 	}
 
-	getExpandoButtons() {
-		return this.entry.querySelectorAll('.expando-button, .search-expando-button');
+	getExpandos() {
+		return this.getTextExpandos().concat(this.getEntryExpando()).filter(v => v);
```
erikdesjardins: I'd prefer:
```js
return [...this.getTextExpandos(), this.getEntryExpando()].filter(v => v);
```

lib/utils/expando.js
```diff
@@ -0,0 +1,221 @@
+import { $ } from '../vendor';
+import expandoTemplate from '../templates/expando.mustache';
+
+// @type {WeakMap.<expandoButton, Expando>}
+export const expandos = new WeakMap();
+
+// Used for specifying primary expando for any given link
+// @type {Object.<string, Expando>} - key: href
+export const primaryExpandos = {};
```
erikdesjardins: I somewhat want this to be a `Map` for symmetry with `expandos`, but I don't feel strongly about that. Feel free to leave it as-is.

lib/modules/showImages.js
```diff
@@ -376,38 +397,22 @@ const resizeSources = {
...
-		const bufferScreens = module.options.bufferScreens.value || 2;
-		const viewportHeight = window.innerHeight;
-		const maximumTop = viewportHeight * (bufferScreens + 1);
-		const minimumBottom = viewportHeight * bufferScreens * -1;
-
-		const isWithinBuffer = ele => {
-			const { bottom, top } = ele.getBoundingClientRect();
-			return top <= maximumTop && bottom >= minimumBottom;
-		};
+		const activeExpandos = Array.from(primaryExpandos).map(([, expando]) => expando);
```
erikdesjardins: This (and other places where `Object.values()` was replaced) could be changed to one of two things:

```js
const activeExpandos = Array.from(primaryExpandos.values())
```

or keep using generators all the way down (with  `map`, `filter`, etc. from utils/generator.js):

```js
// Empty collapsed when beyond buffer
for (const expando of primaryExpandos.values()) {
/* ... */
primaryExpandos.values()
    ::filter(v => v.open)
    ::map(v => ({ media: v.media, data: v.media }))
    ::lazyUnload(isWithinBuffer);
```

lib/modules/showImages.js
```diff
@@ -638,292 +602,220 @@ function domainToModuleName(hostname) {
...
+function getMediaInfo(mediaLink) {
+	let detectResult;
+	const siteModule = [
+		'default',
+		'defaultVideo',
+		'defaultAudio',
+		domainToModuleName(mediaLink.hostname),
+	]
+		.map(modId => domainModuleMap[modId])
+		.find(mod => mod && (detectResult = mod.detect(mediaLink)));
```
erikdesjardins: While we're at it with the generators, we could probably make this a bit cleaner:

```js
const [siteModule, detectResult] = [
    'default',
    'defaultVideo',
    'defaultAudio',
    domainToModuleName(mediaLink.hostname),
]
    ::map(modId => domainModuleMap[modId])
    ::filter(mod => mod)
    ::map(mod => [mod, mod.detect(mediaLink)])
    ::find(([, result]) => result) || [];
```

...actually that might be worse :confused:

lib/modules/showImages.js
```diff
@@ -1032,11 +928,16 @@ function generateGallery(options) {
...
+			const first = newIndex - preloadCount;
+			const last = newIndex + preloadCount;
+
 			pieces.filter(piece => piece.media)
 				.map(piece => ({ media: piece.media, data: piece }))
 				::lazyUnload(piece => {
 					const index = pieces.indexOf(piece);
-					return index >= newIndex - preloadCount && index <= newIndex + preloadCount;
+					if (last > pieces.length && last % pieces.length >= index) return true;
+					if (first < 0 && ((first % pieces.length) + pieces.length) % pieces.length <= index) return true;
```
erikdesjardins: `((first % pieces.length) + pieces.length) % pieces.length`

This might deserve its own function in utils/math.js. Maybe `positiveModulo`?

lib/modules/showImages.js
```diff
@@ -443,141 +461,97 @@ function createImageButtons() {
...
+			else if (!onlyOpen) expando.collapse();
+		});
+}, 100, { leading: true });
 
-function updateRevealedImages(type = 'expando-button') {
-	imageList
-		.filter(v => v.classList.contains(type))
-		.map(v => [v.getBoundingClientRect().top, v])
+function sortExpandosVertically() {
+	return this
```
erikdesjardins: Needs to be `Array.from(this)` to accept arrays or generators

erikdesjardins:
One last thing: right now `autoExpandSelfText` expands everything in any self post, that is, in both self post expandos and on a self post's comments page.

I think we probably don't want it to expand in the latter case (à la 4.6.1).

PR #3002, opened 2016-05-22
ShowImages: mediaHost does not require imageLink

larsjohnsen:


larsjohnsen:
@erikdesjardins Completed this now and removed `rewriteMediaHostOptions`.

lib/modules/hosts/photobucket.js
```diff
@@ -5,22 +5,23 @@ export default {
...
+		let src = href.replace('.html', '');
 
 		// user linked direct image so no need to hit API
-		if (prefix === 'i') {
-			elem.src = href;
-		} else {
+		if (prefix !== 'i') {
 			const { imageUrl } = await ajax({
 				url: 'https://api.photobucket.com/v2/media/fromurl',
 				data: { url: href },
```
erikdesjardins: `const href` was changed to `let src` above; this needs to be changed too.
image

lib/modules/hosts/defaultAudio.js
```diff
@@ -10,7 +10,7 @@ export default {
 		// important that acceptRegex is last (the returned value)
 		return !rejectRegex.test(href) && acceptRegex.exec(href);
 	},
-	handleLink(elem, [, extension]) {
+	handleLink(href, [, extension]) {
```
erikdesjardins: :+1: for getting rid of `elem` here

lib/modules/hosts/imgur.js
```diff
@@ -98,18 +93,18 @@ const imgur = {
 		let getExpandoInfo;
 
 		if (info.images && info.images.length) {
-			getExpandoInfo = this._handleAlbum(elem, info);
+			getExpandoInfo = this._handleAlbum(href, info);
```
erikdesjardins: Might as well return these directly (no `await`) and get rid of `getExpandoInfo`

lib/modules/hosts/memegen.js
```diff
@@ -12,11 +12,11 @@ export default {
...
-	handleLink(elem, [, isAnimated, id]) {
-		elem.type = 'IMAGE';
-		elem.src = isAnimated ?
-			`http://a.memegen.com/${id}.gif` :
-			`http://m.memegen.com/${id}.jpg`;
+	detect: ({ href }) => (/^https?:\/\/(?:www|ar|ru|id|el|pt|tr)\.memegen\.(?:com|de|nl|fr|it|es|se|pl)(\/a)?\/(?:meme|mem|mim)\/([A-Za-z0-9]+)/i).exec(href),
+	handleLink(href, [, isAnimated, id]) {
+		return {
+			type: 'IMAGE',
+			src: `http://a.memegen.com/${id}.${isAnimated ? 'gif' : 'jpg'}`,
```
erikdesjardins: We might need to keep the `a.memegen`/`m.memegen` distinction for `isAnimated`

lib/modules/hosts/twitter.js
```diff
@@ -18,12 +18,15 @@ export default {
...
-			media: {},
+		// Script requires element to be attached to document when starting
+		const $dummy = $('<div>');
+
+		return {
+			type: 'GENERIC_EXPANDO',
+			muted: true,
+			expandoClass: 'twitter',
+			generate: () => $dummy[0],
+			onAttach: () => $dummy.replaceWith($(html)),
```
erikdesjardins: replaceWith can accept the html string directly, `$(html)` is mildly redundant

lib/modules/keyboardNav.js
```diff
@@ -1181,9 +1181,7 @@ module.nextGalleryImage = function() {
 	}
 };
 
-module.toggleViewImages = function() {
-	ShowImages.setShowImages();
-};
+module.toggleViewImages = ShowImages.toggleViewImages;
```
erikdesjardins: I believe this is only used in one place (keyboard shortcut), can you move it there and get rid of `module.toggleViewImages`?

larsjohnsen: I'll do it in #2998

lib/modules/showImages.js
```diff
@@ -260,6 +260,11 @@ module.options = {
 		value: true,
 		description: 'Autoplay inline videos',
 	},
+	keepNativeExpando: {
```
erikdesjardins: I'm not sure that we need an option for this, since we've been doing it for a while without one (unless you think you'd use it)

larsjohnsen: Some of the mediaHosts contains something akin to `!elem.classList.contains('title')` in `detect`, but perhaps that can be replaced with an option `isIdenticalToNative`

lib/modules/hosts/imgur.js
```diff
@@ -43,73 +42,75 @@ const imgur = {
...
 
-	async handleLink(elem) {
-		const href = elem.href.split('?')[0];
-		let groups, hash, getInfo;
+		let groups, hash;
 
 		if ((groups = this.galleryHashRe.exec(href))) {
 			hash = groups[1];
-			getInfo = this._api(string.encode`gallery/${hash}`);
+			return ['api', string.encode`gallery/${hash}`];
```
erikdesjardins: I'd prefer ```return [this._api, string.encode`gallery/${hash}`];``` or ```return () => this._api(string.encode`gallery/${hash}`);``` (probably the latter)

lib/modules/showImages.js
erikdesjardins: Yeah I originally added those, but I was wrong, they're not actually identical: for example, reddit's embeds don't get added to history when you expand them (if you have that option enabled).

So I think it's fine to just override everything

erikdesjardins:
Thanks! Will merge as soon as the build passes.

Sorry about the delay.

lib/modules/keyboardNav.js
larsjohnsen: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3004/commits/4b7552a3eccb216d50b69ebbd81d868f92be601d

PR #3004, opened 2016-05-24
ShowImages: Rewrite custom image filter

larsjohnsen:
This uses the `li.selected` class to indicate that a tab / filter is active; another class may be more appropriate.

### DIFF
--- a/lib/templates/tabMenuLink.mustache
+++ b/lib/templates/tabMenuLink.mustache
@@ -1,0 +1,8 @@
+<li class="res-tabmenu-button">
+	<a class="{{ className }}" {{#title}}title="{{ title }}"{{/title}}>
+	<label>
+		<input type="checkbox" {{checked}}>
+		{{text}}
+	</label>
+	</a>
+</li>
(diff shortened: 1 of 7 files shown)


lib/utils/createElement.js
```diff
@@ -80,3 +81,30 @@ export function table(items, callback) {
...
+		return {
+			label: document.createElement('div'),
+			checkbox: document.createElement('input'),
+		};
+	}
+
+	const checkbox = element.querySelector('input');
+	checkbox.addEventListener('change', e => {
+		e.stopPropagation();
+		element.classList[e.target.checked ? 'add' : 'remove']('selected');
```
erikdesjardins: `element.classList.toggle('selected', e.target.checked)`

lib/templates/tabMenuLink.mustache
```diff
@@ -0,0 +1,8 @@
+<li class="res-tabmenu-button">
+	<a class="{{ className }}" {{#title}}title="{{ title }}"{{/title}}>
+	<label>
+		<input type="checkbox" {{checked}}>
+		{{text}}
```
erikdesjardins: The newlines here add extra space, might as well put everything inside the `<label>` on one line (could use CSS of course but that's probably overkill)

image

larsjohnsen: Fixed. `.tabmenu li` uses `display: inline`, which for some reason caused that extra margin.

PR #3010, opened 2016-05-25
Shortened repetitive descriptions for ease of read

saizai:
Current help menu is hard to read at a glance because it is tl;dr. I've trimmed descriptions of things that are just variants of their preceding item.

Another thing that needs to be done is to make the help menu obey night mode, but I'm not sure how to do that.

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -157,7 +157,7 @@ const commands = {
 	moveDown: {
 		include: ['linklist', 'modqueue', 'profile', 'search'],
 		value: [74, false, false, false], // j
-		description: 'Move down to the next link or comment in flat lists',
+		description: '… to next link',
 	},
 	moveUpComment: {
 		include: ['comments', 'inbox'],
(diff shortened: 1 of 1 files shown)


andytuba:
> Another thing that needs to be done is to make the help menu obey night mode, but I'm not sure how to do that.

https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3319

andytuba:
This is a good idea, but the actual phrasing doesn't seem to help readability. It also hurts searching for options.  Maybe change the descriptions into something like `... short version. <p>Longer description</p>`?

saizai:
@andytuba I did it in context - they're all shown together. So it's meant to look like a list of variants on the previous command.

andytuba:
However, they're not always shown together, like when the user searches options. This change also makes it more difficult to search for the options.

It would be better to do something like add a `shortDescription` property to each item and change [`keyboardNav.drawHelp`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/35fd8aaf83c9a6cff419f312ff8b94d60e9ace69/lib/modules/keyboardNav.js#L810) to pull from `opt.menuDescription || opt.description`

@erikdesjardins architectural opinion on "short description"?

saizai:
Ah, didn't realize that. +1 to your short vs long description idea, then.

erikdesjardins:
@andytuba It's not great but I don't think there's anything better

PR #3013, opened 2016-05-27
ShowImages: Restore expando classes

larsjohnsen:
May as well restore them.

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -10,28 +10,28 @@ img {
 	image-orientation: from-image;
 }
 
-.commentarea .thing .usertext-body .expando-button,
-.listing-page .thing .expando-button {
-	padding: 0;
-}
-
-.profile-page .comment .parent .expando-button {
-	display: inline-block;
-	float: none;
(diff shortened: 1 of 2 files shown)


lib/css/modules/_showImages.scss
```diff
@@ -43,39 +43,42 @@ img {
...
+
+		&.image.collapsedExpando { background-position: 0 _(0); }
+		&.image.collapsedExpando:hover { background-position: 0 _(1); }
+		&.image.expanded { background-position: 0 _(2); }
+		&.image.expanded:hover { background-position: 0 _(3); }
+		&.gallery.collapsedExpando { background-position: 0 _(4); }
+		&.gallery.collapsedExpando:hover { background-position: 0 _(5); }
+		&.gallery.expanded { background-position: 0 _(6); }
+		&.gallery.expanded:hover { background-position: 0 _(7); }
+		&.video-muted.collapsed { background-position: 0 _(8); }
```
erikdesjardins: If `.video.res-media-muted` is used above then it should be used here too, or vice versa.

PR #3015, opened 2016-05-28
ShowImages MediaControls: Style clean-up

larsjohnsen:
Also fixes gearIcon

image

### DIFF
--- a/lib/templates/mediaControls.mustache
+++ b/lib/templates/mediaControls.mustache
@@ -1,10 +1,10 @@
 <div class="res-media-rotatable">
-	<div class="RESMediaControls {{position}}">
-		<span class="res-icon gearIcon" title="Settings" data-action="showImageSettings"></span>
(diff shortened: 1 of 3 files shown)


lib/templates/mediaControls.mustache
```diff
@@ -1,10 +1,10 @@
 <div class="res-media-rotatable">
-	<div class="RESMediaControls {{position}}">
-		<span class="res-icon gearIcon" title="Settings" data-action="showImageSettings"></span>
-		<span class="rotateLeft res-icon" title="Rotate image counter-clockwise" data-action="rotateLeft"></span>
-		<span class="rotateRight res-icon" title="Rotate image clockwise" data-action="rotateRight"></span>
+	<div class="res-media-controls res-media-controls-{{x}} res-media-controls-{{y}}">
+		<div class="res-icon gearIcon" title="Settings" data-action="showImageSettings"></div>
```
andytuba: it would be nice to make these `<button>` but would resetting that style be too much trouble?

PR #3027, opened 2016-05-31
showImages: Allow iframe overflow

larsjohnsen:
Removes the width restriction imposed on it by `.entry`, similar to videos and images.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1118,13 +1118,16 @@ function generateIframe(options) {
 	const iframeNode = document.createElement('iframe');
 	iframeNode.src = (module.options.autoplayVideo.value && options.embedAutoplay) ?
 		options.embedAutoplay : options.embed;
-	iframeNode.width = options.width || '640';
-	iframeNode.height = options.height || '360';
 	iframeNode.allowFullscreen = true;
-	iframeNode.setAttribute('frameborder', '0');
-	iframeNode.style.maxWidth = options.maxwidth || '100%';
+	iframeNode.style.border = '0';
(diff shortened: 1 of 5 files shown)


lib/modules/showImages.js
```diff
@@ -1118,13 +1118,17 @@ function generateIframe(options) {
...
-	iframeNode.height = options.height || '360';
+	iframeNode.style.minWidth = options.width || '640px';
+	iframeNode.style.height = options.height || '360px';
 	iframeNode.allowFullscreen = true;
 	iframeNode.setAttribute('frameborder', '0');
-	iframeNode.style.maxWidth = options.maxwidth || '100%';
 
 	iframeNode.expand = () => {
+		iframeNode.dispatchEvent(new CustomEvent('resize', { bubbles: true }));
+		if (iframeNode.style.minWidth.match(/%$/) && iframeNode.clientWidth < 480) {
```
erikdesjardins: `iframeNode.style.minWidth.endsWith('%')`

erikdesjardins: Actually, could we accomplish the same thing by doing:
```js
iframeNode.style.minWidth = '480px';
iframeNode.style.width = options.width || '640px';
```
above?

Or is there a reason that `options.width` needs to be the `minWidth`?

larsjohnsen: `iframe` behaves weirdly (at least in Chrome) when parent has a width-limit. Fixed it by wrapping it in a non width limited `div` instead, in latest commit.

larsjohnsen: Never mind, that did not actually work. Reverted it, and added a comment instead.

PR #3034, opened 2016-06-01
Switch soundcloud to less bloated player, no ajax

corylulu:
This uses a different player that is similar, smaller, and does the same stuff. It also allows for a method to create the embeded player using a far simpler method.

### DIFF
--- a/lib/modules/hosts/soundcloud.js
+++ b/lib/modules/hosts/soundcloud.js
@@ -6,12 +6,11 @@ export default {
 	domains: ['soundcloud.com'],
 	logo: 'https://a-v2.sndcdn.com/assets/images/sc-icons/favicon-2cadd14b.ico',
 	detect: () => true,
-	async handleLink(href) {
-		const src = string.encode`https://w.soundcloud.com/player/?url=${href}`;
+	handleLink(href) {
 		return {
 			type: 'IFRAME',
-			embed: src,
-			height: '124px',
+			embed: string.encode`https://w.soundcloud.com/player/?url=${href}`,
+			height: '166px',
 			pause: '{"method":"pause"}',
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/soundcloud.js
```diff
@@ -8,21 +7,11 @@ export default {
 	logo: 'https://a-v2.sndcdn.com/assets/images/sc-icons/favicon-2cadd14b.ico',
 	detect: () => true,
 	async handleLink(href) {
```
erikdesjardins: `async` can be removed

lib/modules/hosts/soundcloud.js
```diff
@@ -8,21 +7,11 @@ export default {
...
-				url: href,
-				format: 'json',
-				iframe: 'true',
-			},
-			type: 'json',
-		});
-
-		// Get src from iframe html returned
-		const src = $(html).attr('src');
+		const src = string.encode`https://w.soundcloud.com/player/?url=${href}`;
```
erikdesjardins: Could move this into the object, since this function is so simple now

lib/modules/hosts/soundcloud.js
```diff
@@ -8,21 +7,11 @@ export default {
...
 		return {
 			type: 'IFRAME',
 			embed: src,
+			height: '124px',
```
erikdesjardins: I think 166px is better (it's also used [here](https://soundcloud.com/pages/embed)). And `width: '100%'`

Examples:

200px
image

166px
image

124px
image

corylulu: Yeah, was debating between the two, thought 166px looks better for posts, but worse for comments... but sure...

PR #3041, opened 2016-06-03
Imgur cleanup

erikdesjardins:


lib/modules/hosts/imgur.js
```diff
@@ -77,6 +76,7 @@ const imgur = {
 			headers: {
 				Authorization: `Client-ID ${this.apiID}`,
 			},
+			cacheFor: DAY,
```
erikdesjardins: Not sure if this is too long. Maybe `endpoint.startsWith('album/') ? HOUR : DAY` in case album captions are updated?

andytuba:
These aren't cached by browser?

erikdesjardins:
...of course they are, duh :sweat_smile: 
I think the delay from the batching fooled me

andytuba:
Hah yeah the batching isn't useful yet. If it ever gets implemented I can take advantage of it then.

andytuba:
It=batch api endpoint

Issue #2902, opened 2016-04-16
customFilter condition for "has expando"


andytuba:
That'll be interesting to implement. Make sure to check for both reddit expandos and showImages expandos.

https://www.reddit.com/r/Enhancement/comments/4f3frb/is_it_possible_to_filter_your_front_page_to_only/

Issue #2896, opened 2016-04-14
Option for filteReddit to also hide those posts


andytuba:
https://www.reddit.com/r/Enhancement/comments/4ejk01/feature_request_autohide_reddits_builtin_hide/

andytuba:
I would recommend editing `filteReddit.js`:

1. add a boolean option `hideFilteredPosts`
1. add a line to [where "this post is filtered" class is applied](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/01f634d3a61904617d418d828dc4ff597b8c8524/lib/modules/filteReddit.js#L791) that checks `if (module.options.hideFilteredPosts.value) { add post ID to a list to send to the hide-many-posts api endpoint}` 
1. before [notifying filtered](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/01f634d3a61904617d418d828dc4ff597b8c8524/lib/modules/filteReddit.js#L808) send a request to the "hide many posts" api with that list

Issue #2723, opened 2016-02-15
Any interest in feature for filtering reddit saved page by subreddit?


chrisshroba:
I'm proposing a feature where a user can filter his/her `https://www.reddit.com/user/<username>/saved/` page by subreddit, to be able to see only posts saved from a particular subreddit, such as /r/gifrecipes or /r/funny.  If there's interest, I'll look into implementing!

andytuba:
Since this is already a feature of reddit gold, it's inappropriate for RES.

However, it would be useful to add "when browsing saved page" or "when browsing saved page for /u/[optional]"  as a criteron to filteReddit customFilters.

andytuba:
I think @larsjohnsen has handled this in filterline?

Issue #2296, opened 2015-06-16
"hide all" posts button 


andytuba:
Now that reddit has an API endpoint for hiding 50 posts at once, we may as well add a button to RES.

I propose adding it to the "Page #" NER banner and below the nextprev buttons when NER is disabled, and it can remove the posts preceding the button.  It could have a yes/no confirmation (optional, default on). After clicking the button, it should hide all posts on front-end and scroll to keep the user looking at the button/banner.  This might should be offloaded to the background page in case of throttling.

crosma:
This would be great for Reddit load wise because of people who browse like I do. I view everything on a page then hide it all. I do this because I browse in random bits when I'm not busy and I hate seeing 50%+ repeats every time I load up my front page. Currently I just scroll back to the first post and then hold H which generates 25 requests.

aperson:
If this ever gets implemented, please don't limit it to NER.  I don't use NER but would love to have a 'hide all' button.

andytuba:
For non-NER how about an option to change the "next page" button to "hide all and view next page", and/or a button next to "next" that's "hide all and refresh" if you're still on page 1?

aperson:
Hide all and refresh would be my preferred option.

andytuba:
https://www.reddit.com/r/Enhancement/comments/4tafpb/hide_threads/

andytuba:
https://www.reddit.com/r/Enhancement/comments/5frnyo/feature_request_a_mark_all_as_read_button_that/

PR #3042, opened 2016-06-04
filteReddit: Filterline [RFC]

### LINKED ISSUE #1011: reveal which filteReddit rule filtered posts
andytuba            it'd be kinda interesting to add a toggle to filteReddit that flagged posts instead of hiding them          3:29
andytuba            esp with a little "why was this filtered?" info thing revealing the specific rule           3:30
HandicapperGeneral          yea         3:31
HandicapperGeneral          or oh           3:31
HandicapperGeneral          there's the notification that says how many were filtered           3:31
HandicapperGeneral          that should have an info box

### LINKED ISSUE #1467: nsfw filter only for /r/all
http://www.reddit.com/r/Enhancement/comments/2b98it/my_list_of_subreddits_i_filter_from_rall_what_are/cj3gzal

this is easy to implement but harder to design

i want to do something a little more expansive with this anyway.

### LINKED ISSUE #1508: High-pass score filter for frontpage posts per subreddit
http://www.reddit.com/r/Enhancement/comments/2divdx/feature_request_filter_posts_that_show_up_on/

Although it might result in a bad-ish user experience, it still seems like a reasonable enhancement.  people can take it upon themselves to create a bad experince.

larsjohnsen:
peek 2016-09-08 13-05
- Filters has 3 states: inactive, active and active inverse.
- Some filters can have an additional criterion (such as age, expando type, score)
  - Criterions are entered via the console
  - These filters are hidden till criterion are chosen
- Duplicate filters are allowed, such as `age < 10h`, `age > 5h`, and `expando` `not expando native`
- Filter states and criterion are automatically remembered for the given subreddit
- Display filter reason for the individual `Thing`
- Rewrite of the decision tree for what shall be filtered
  - The original filters (`ExternalFilter`) is integrated with the filterline
  - Possibility to remove the `ExternalFilter` entry which caused something to be filtered
- Export to `customFilters`
- Create filter from selected `Thing` attribute 

Ideas:
- Save set of filters as a group, which also can be restored in another subreddit
- Option to place filterline in the sidebar
- Comments opened: Integrate with newComments to only show a post when exceeding x number of new comments

Resolves #1467 — implemented toggle for nsfw
Resolves #1508 — via the command line
Resolves #1011 — using `show filter-reason`
Lays groundwork for #2902 — a has expando case is added, but is not included in customFilters as it doesn't support async
Lays grundwork for #2296 (utils/thingsHide.js)
Implements #2896 — a checkbox to toggle whether those currently filtered shall be hidden, is placed in the dropdown
Lays groundwork for #2723, #2201


BenMcGarry:
What's the current status with this PR, would be nice to get it into 4.7.

lib/modules/filteReddit.js
```diff
@@ -746,9 +753,17 @@ module.always = function() {
 	Menu.addMenuItem(nsfwSwitch, onClickSwitch);
 };
 
+let filterline;
+export let forceExpandoCompletion = false;
```
erikdesjardins: I don't really think this belongs in filteReddit. I'd prefer to have showImages export a function allowing the caller to enable/disable forced expando completion.

lib/modules/filteReddit.js
```diff
@@ -746,9 +753,17 @@ module.always = function() {
...
+export let expandoCompletionCallback;
```
erikdesjardins: Similar thing here, we shouldn't put the burden of null-checking on the caller nor should showImages know about the internals of this module. Instead, have showImages export a function to register (and unregister, if necessary) callbacks.

lib/modules/filteReddit.js
```diff
@@ -786,18 +801,224 @@ module.go = function() {
...
 
-	if (isPageType('linklist')) {
-		if (module.options.showFilterVisitedLinks.value) {
-			CreateElement.tabMenuItem({
-				text: 'filter visited links',
-			}).checkbox.addEventListener('change', e => {
-				removeVisitedLinksActive = e.target.checked;
-				scanEntries();
-			});
+function getSubredditDefinedFilters() {
```
erikdesjardins: TBH I'm not sure if we need to keep this -- I've never, even seen any subreddit using it. Unless you've seen it used, I'd remove it.

lib/modules/filteReddit.js
```diff
@@ -786,18 +801,224 @@ module.go = function() {
...
+				const score = parseInt(this.criterion, 10);
+				if (isNaN(score)) return null;
+				return module.options.customFilters.cases.score.evaluate(thing, { op: '>', val: score });
+			},
+			visible: false,
+			visibleOnCriterion: true,
+		},
+	];
+
+	filters.forEach(::filterline.addFilter);
```
erikdesjardins: We don't need to bind these; I'd prefer `filters.forEach(filter => filterline.addFilter(filter))` or a for-of loop

larsjohnsen:
Some new features I added now:
- Ability to show reason for filtering
- Shortcuts to and superficial control of the other filter engines (`keywords`, `customFilters`, ...)

@BenMcGarry I don't think it's ready yet.

larsjohnsen:
@erikdesjardins Given this any more thought? It's been fairly stable for a while now (for my use cases, anyhow), so it'd be great to get more feedback.

larsjohnsen:
Preliminary instructions. Any thoughts?

----

To activate the filterline, click on the '∀' icon in the tab menu.

- Click once on a filter (e.g. 'has expando') to only show things which matches it.
- Click again to only show things which does not match it.
- To clear a filter (delete or make inactive), right click on it.

Some filters support a criterion. This is entered by going through the "more" dropdown and clicking on a filter. E.g. does 'has expando' support 'image', which makes it filter only things which has image expandos.

You can add multiply filters of the same type. To do this, click on the '+' button left on the name.

You can use the currently selected thing as basis for a filter. To do this, click on '=' to the right of the 

'Save to customFilters' extracts the current filters and inserts them into a new customFilter, which can further specify when the filters shall be active.

You can also use the command line to manipulate the filterline. Pressing the key 'f' will bring it up.

andytuba:
I'd like to add a "quickly disable all filters" button somewhere. Do you think I should tack it onto the end of filterLine?

larsjohnsen:
@andytuba That would be a nice addition to filterline. I envision it as a small toggle (off - on) floating to the right.

andytuba:
I like this idea. I'll base a branch off this and work on it this week.

My vote's on getting this caught up to date, sticking "BETA" in the description as a CYA, and getting it merged so we can start playing with it.

lib/modules/filteReddit.js
```diff
@@ -12,21 +14,609 @@ import {
...
+		falseText: '¬ visited',
+		defaultTemplate() {
+			return { type: 'isVisited' };
+		},
+		fields: [
+			'post link has not been visited',
+		],
+		disabled: process.env.BUILD_TARGET === 'safari' || process.env.BUILD_TARGET === 'edge',
+		get alwaysShow() { return !this.disabled; },
+		async evaluate(thing) {
```
erikdesjardins: Either remove `async` (which is unnecessary currently) or `await` `isURLVisited` (don't care which)

lib/modules/filteReddit.js
```diff
@@ -12,21 +14,609 @@ import {
...
+		falseText: '¬ comments opened',
+		defaultTemplate() {
+			return { type: 'commentsOpened' };
+		},
+		fields: [
+			'comment link has not been visited',
+		],
+		disabled: process.env.BUILD_TARGET === 'safari' || process.env.BUILD_TARGET === 'edge',
+		get alwaysShow() { return !this.disabled; },
+		async evaluate(thing) {
```
erikdesjardins: Ditto with `async`

lib/modules/filteReddit.js
```diff
@@ -704,26 +904,41 @@ module.include = [
...
+	restorePage = browseContexts.currentLocation.getCurrent();
+
+	filterlineKey = `RESmodules.filteReddit.${restorePage.toLowerCase()}`;
+
+	filterlineState = await Storage.get(filterlineKey);
+	initialFilterlineVisibility = module.options.showFilterline.value || (filterlineState && filterlineState.active);
+
+	if (initialFilterlineVisibility) {
+		// Pad the space where the filterline will be shown, so that it won't push `siteTable` down when initialized
+		BodyClasses.toggle(true, 'res-filteReddit-filterline-pad-until-ready');
```
erikdesjardins: `BodyClasses.add('res-filteReddit-filterline-pad-until-ready');`

lib/modules/filteReddit.js
```diff
@@ -704,26 +904,41 @@ module.include = [
...
-	filterNSFW(module.options.NSFWfilter.value);
+let filterlineKey;
+let filterlineState;
+let initialFilterlineVisibility;
+let restorePage;
+
+module.beforeLoad = async () => {
+	updateNsfwBodyClass(module.options.NSFWfilter.value);
+
+	restorePage = browseContexts.currentLocation.getCurrent();
```
erikdesjardins: Should be possible to make this `const` and initialize it immediately

lib/modules/filteReddit.js
```diff
@@ -704,26 +904,41 @@ module.include = [
...
+
+	filterlineKey = `RESmodules.filteReddit.${restorePage.toLowerCase()}`;
```
erikdesjardins: and then this too

lib/modules/filteReddit.js
```diff
@@ -704,26 +904,41 @@ module.include = [
...
+
+	filterlineState = await Storage.get(filterlineKey);
```
erikdesjardins: Async storage ops in `beforeLoad()` aren't to be taken lightly since they block `go()` for all other modules.

Can we move the active status out of `filterlineState` and into a module option, so an extra storage fetch isn't necessary? (like [nightmodeOn](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/ac47977fe300c606757025bcbd2225539614ba29/lib/modules/nightMode.js#L24))

Then we can do the fetch just before `createFilterline` (maybe pass it down to where it's used so we don't need the global).

lib/modules/filteReddit.js
```diff
@@ -732,338 +947,817 @@ module.go = () => {
...
-				className: 'res-filter-visited-links',
-			}).addEventListener('change', e => {
-				removeVisitedLinksActive = e.detail;
-				scanEntries();
+	update(state, criterion, { save = true } = {}) {
+		this.state = state;
+
+		if (criterion !== undefined && criterion !== this.criterion) {
+			if (this.parse.length) this.criterion = criterion;
+			else if (criterion) console.log('Filter does not support criterion. Ignoring criterion');
```
erikdesjardins: Might as well log the value of `criterion` too

lib/modules/filteReddit.js
```diff
@@ -732,338 +947,817 @@ module.go = () => {
...
+
+	if (initialFilterlineVisibility) {
+		filterlineTab.click();
+		BodyClasses.toggle(false, 'res-filteReddit-filterline-pad-until-ready');
+	}
+
+	return filterline;
+}
+
+class Filter {
```
erikdesjardins: I think it would make sense to move these classes, the big object definitions at the top (`browseContexts`, etc.) to their own files in `lib/modules/filteReddit` and import them here; should make things easier to manage

lib/modules/filteReddit.js
```diff
@@ -732,338 +947,817 @@ module.go = () => {
...
-		} else if (fullmatch) {
-			// simple full string match
-			return stringToSearch === filter.searchString;
+const regexRegex = /^\/(.*)\/([gim]+)?$/;
+
+const getStringMatchFilters = _.memoize(type => {
+	const sources = module.options[type].value;
+
+	return sources.map(source => {
+		const [matchString, applyTo = 'everyhere', applyList = '', except = ''] = Array.isArray(source) ? source : [source];
```
erikdesjardins: `everyhere` -> `everywhere`

larsjohnsen:
Thanks, @erikdesjardins 

I couldn't figure out how to integrate filterlineState.active to options. Any ideas?

erikdesjardins:
Good for rebase on #3531 now

lib/modules/filteReddit/Filter.js
```diff
@@ -0,0 +1,195 @@
...
+import { Storage } from '../../environment';
+import filterlineFilterTemplate from '../../templates/filterlineFilter.mustache';
+import * as Notifications from '../notifications';
+import * as SelectedEntry from '../selectedEntry';
+import * as FilteReddit from '../filteReddit';
+
+export default class Filter {
+	disabled = false;
+	matchesCount = 0;
+	state = null; // FilterState
```
erikdesjardins: :heart: :heart: :heart: types

erikdesjardins:
Awesome!

PR #3045, opened 2016-06-05
Tell user to reload when score has been revealed

nyuszika7h:
Instead of giving them a negative amount of time with incorrect pluralization.

### DIFF
--- a/lib/modules/betteReddit.js
+++ b/lib/modules/betteReddit.js
@@ -175,7 +175,11 @@ module.go = function() {
 					const scoreHiddenDuration = parseInt(this.getAttribute('title').match(/[0-9]+/)[0], 10);
 					const postTime = new Date(timeNode.getAttribute('datetime')).getTime();
 					const minutesLeft = Math.ceil((postTime + scoreHiddenDuration * 60000 - new Date().getTime()) / 60000);
-					this.setAttribute('title', `score will be revealed in ${minutesLeft} minute${minutesLeft > 1 ? 's' : ''}`);
+					if (minutesLeft >= 1) {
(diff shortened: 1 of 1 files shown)


lib/modules/betteReddit.js
```diff
@@ -175,7 +175,11 @@ module.go = function() {
 					const scoreHiddenDuration = parseInt(this.getAttribute('title').match(/[0-9]+/)[0], 10);
 					const postTime = new Date(timeNode.getAttribute('datetime')).getTime();
 					const minutesLeft = Math.ceil((postTime + scoreHiddenDuration * 60000 - new Date().getTime()) / 60000);
-					this.setAttribute('title', `score will be revealed in ${minutesLeft} minute${minutesLeft > 1 ? 's' : ''}`);
+					if (minutesLeft >= 1) {
+						this.setAttribute('title', `score will be revealed in ${minutesLeft} minute${minutesLeft === 1 ? 's' : ''}`);
+					} else {
+						this.setAttribute('title', 'reload page to reveal score');
+					}
```
13steinj: +1, but I fell as if this can be done better. Reloading is not very user friendly in some cases

Just as some psuedocode:

    } else {
        this.setAttribute('title', 'click to reveal score');
        $(this).click(function(e) {
            var thing = $(this).parent().parent().parent();
            var things = [thing];
            $.merge(things, thing.find('thing'));
            var ids = l.map(i => l[i].id.replace('thing_t1_', '').replace('thing_t3_', ''));
            var mappedthings = {};
            $.each(ids, function(index, val){mappedthings[val] = $(things[index]).find('>.entry .score-hidden')});
            var newscoredata = //request data from thing's permalink .json
        });
    }

And using that data update all possible hidden score spans. I'm on mobile so that was a tad lazy and possible full of mistakes.

PR #3053, opened 2016-06-06
Fix oddshot.tv expando to work with new URLs

pmlocek:
Oddshot changed URL scheme to something like "shot/streamer/hash". Changed regex to match new and old URLs.


erikdesjardins:
Thanks!

PR #3054, opened 2016-06-07
Crushed PNGs using OptiPNG

robotortoise:
Crushed PNGs using OptiPNG

They're small images so it didn't save much space.


andytuba:
Neat! Thanks for putting these in.

Unfortunately, most of these assets won't be going into the next release. We're switching out all the 👽++ with the new UFO branding, and I'd like to replace the CSS buttons with a UFO+CSS.

Can I ... png you when those ready?

robotortoise:
_groannnnn_

But in all serious, yeah, please do! PNG crushing is incredibly easy.

andytuba:
Thanks! It'd be neat if you documented your process / settings, too, but that's gilding the lily.

robotortoise:
Ah, well there is none!

All you do is use [OptiPNG's compiled 32-bit build ](http://optipng.sourceforge.net/)and drag and drop the PNGs onto the executable.

Honestly, I don't think you really need me. I'm happy to help out, but it's as simple as drag and drop.

PR #3064, opened 2016-06-10
Update to jQuery 3

erikdesjardins:


mc10:
We should probably wait for the release after the initial v3 to wait for any initial bugs and kinks to be worked out.

erikdesjardins:
There's been a prerelease out for a month, but sure, we're in no rush

erikdesjardins:
v3 has now been out for a month; AFAICT the only issue that has appeared is swallowing exceptions in `$(function)` / `$.ready()`, which we don't use.

At this point I doubt any major issues will show up, so let's go for it.

PR #3071, opened 2016-06-13
filteReddit fixes

larsjohnsen:


lib/modules/filteReddit.js
```diff
@@ -54,8 +54,8 @@ module.options = {
 	},
 	showFilterVisitedLinks: {
 		type: 'boolean',
-		value: process.env.BUILD_TARGET !== 'safari',
-		noconfig: process.env.BUILD_TARGET === 'safari',
+		value: process.env.BUILD_TARGET !== 'safari' && process.env.BUILD_TARGET !== 'edge',
+		noconfig: process.env.BUILD_TARGET === 'safari' && process.env.BUILD_TARGET === 'edge',
```
erikdesjardins: `&&` -> `||`

lib/modules/filteReddit.js
```diff
@@ -54,8 +54,8 @@ module.options = {
 	},
 	showFilterVisitedLinks: {
 		type: 'boolean',
-		value: process.env.BUILD_TARGET !== 'safari',
-		noconfig: process.env.BUILD_TARGET === 'safari',
+		value: process.env.BUILD_TARGET !== 'safari' || process.env.BUILD_TARGET !== 'edge',
```
erikdesjardins: Only on the second one, this should still be `&&`, sorry

Since you want `!value == noconfig`, and `!(!a && !b) == a || b`

PR #3072, opened 2016-06-13
Fix regex for short subreddit names

erikdesjardins:
Subreddits can apparently be fewer than 3 characters, e.g. /r/de


andytuba:
Some lucky very old subreddits, I think..

PR #3082, opened 2016-06-17
Setting link that can be relatively linked (backwards compatible)

corylulu:
Currently, since the current links (`#!settings`) begin with an exclamation mark `!` after the `#`, they cannot be relatively linked since reddit requires 'domains' to start with an alphanumeric character. 

So this is an invalid link:

`[RES Settings](#!settings)`

So I'm proposing changing to a valid link syntax:

`[RES Settings](#res/settings)`

So people on reddit can more easily link to reddit settings that work no matter what page (or page context level) you are viewing them from. 

This change is backwards compatible as well, so previous links will still work and be converted to the new format. The `hashComponents` doesn't have to be changed for this to work, but if it doesn't, it would greatly hinder the features discoverability.

### DIFF
--- a/lib/modules/settingsNavigation.js
+++ b/lib/modules/settingsNavigation.js
@@ -61,7 +61,7 @@ export function makeUrlHashLink(moduleID, optionKey, displayText, cssClass) {
 }
 
 export function makeUrlHash(moduleID, optionKey) {
-	const hashComponents = ['#res/settings'];
+	const hashComponents = ['#res:settings'];
 
 	if (moduleID) {
 		hashComponents.push(moduleID);
@@ -101,7 +101,7 @@ export function resetUrlHash() {
 
 function onHashChange() {
 	const hash = location.hash;
-	if (!hash.startsWith('#!settings') && !hash.startsWith('#res/settings')) return;
+	if (!hash.startsWith('#!settings') && !hash.startsWith('#res:settings')) return;
 
 	const params = hash.match(/\/(?:\w|\s|%20)+/g);
 	let moduleID;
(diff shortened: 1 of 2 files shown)


andytuba:
Great fix, my one big regret on settings navigation. Even better to remove the current url from makeurlhash.

lib/modules/settingsNavigation.js
```diff
@@ -101,7 +101,7 @@ export function resetUrlHash() {
 
 function onHashChange() {
 	const hash = location.hash;
-	if (hash.substring(0, 10) !== '#!settings') return;
+	if (hash.substring(0, 10) !== '#!settings' && hash.substring(0, 13) !== '#res/settings') return;
```
erikdesjardins: While we're at it, might as well update this to:

```js
if (!hash.startsWith('#!settings') && !hash.startsWith('#res/settings')) return;
```

erikdesjardins:
We can also remove the `baseUrl` from `makeOptionSearchResultLink` in search.js, yay!

corylulu:
@andytuba, Can you elaborate a bit more? Don't have a firm grasp of everything going on in regards to that.

andytuba:
@corylulu https://github.com/corylulu/Reddit-Enhancement-Suite/blob/dd213a5122dc18be66d23d44650fa3f225fd9b53/lib/modules/search.js#L314

corylulu:
@andytuba Ahh, got it. Updated that block.

lib/modules/settingsNavigation.js
```diff
@@ -118,7 +118,7 @@ function onHashChange() {
 
 function onPopState(event) {
 	const state = typeof event.state === 'string' && event.state.split('/');
```
erikdesjardins: This doesn't work anymore because URLs with `#res/settings` have more slashes than those with `#!settings`.

To fix this you could switch to `#res:settings` (like a pseudo-protocol), or change the parsing logic; I prefer the former, if you have no objections.

erikdesjardins:
Awesome, thanks!

PR #3084, opened 2016-06-17
showImages: Media max size (maxHeight, maxWidth) as percentage of viewport

larsjohnsen:
Also sets default max-width to 100% of viewport, since it works fine with `keepMediaVisible`. In addition it sets default max-height to 95% of viewport, as most media otherwise would require scrolling.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -378,8 +378,9 @@ const mediaStates = {
 };
 
 const resizeSources = {
-	MANUAL: 0,
-	OTHER: 1,
+	KEEP_VISIBLE: 0,
+	MANUAL: 1,
+	OTHER: 2,
 };
 
 function isWithinBuffer(ele) {
@@ -1260,7 +1261,7 @@ function keepMediaVisible(media) {
 	media.classList.add('res-media-keep-visible');
 
 	const adjustAlignment = frameDebounce(e => {
-		if (!media.offsetParent) return;
+		if (e.detail === resizeSources.KEEP_VISIBLE || !media.offsetParent) return;
 
 		// Realignment on manual resizing may be non-intutive
 		if (isManuallyMoved || e.detail === resizeSources.MANUAL) {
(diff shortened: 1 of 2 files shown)


lib/modules/showImages.js
```diff
@@ -380,6 +380,7 @@ const mediaStates = {
 const resizeSources = {
 	MANUAL: 0,
 	OTHER: 1,
+	KEEPVISIBLE: 2,
```
erikdesjardins: Nit: `KEEP_VISIBLE`

lib/modules/showImages.js
```diff
@@ -1303,6 +1300,22 @@ function setMediaClippyText(elem) {
 	elem.setAttribute('title', title);
 }
 
+function setMediaMaxSize(media) {
+	let value = module.options.maxWidth.value;
+	let maxWidth = parseInt(value, 10);
+	if (maxWidth > 0) {
+		if (_.isString(value) && value.match('%')) maxWidth *= document.documentElement.clientWidth / 100;
```
erikdesjardins: `value.endsWith('%')` or `value.includes('%')` instead of `match`

(same below)

lib/modules/showImages.js
```diff
@@ -72,14 +72,14 @@ module.options = {
...
-		description: 'Max width of image displayed onscreen (enter zero for unlimited).',
+		value: '100%',
+		description: 'Max width of media (in pixels, enter zero for unlimited). Percentage of window width may also be used (e.g. "100%").',
 		advanced: true,
 	},
 	maxHeight: {
 		type: 'text',
-		value: '0',
-		description: 'Max height of image displayed onscreen (enter zero for unlimited).',
+		value: '95%',
```
erikdesjardins: We should have migrators for these; you can add another one to the end of the array and remove [the existing one](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/ba1e2a871721245776cad1d7a35e2add4725eaf9/lib/core/migrate.js#L240).

larsjohnsen: It's seems that it can be either, as it seemingly is coerced to a number somewhere (`storage get RESoptions.showImages`)

erikdesjardins: Yep, looks like I was wrong, you _do_ need to use a number in the migrator.

erikdesjardins: Hmm, 95% is just tall enough that images with titles or descriptions require scrolling, and I think we should be a little conservative since we're increasing it from 640x480.

Maybe bring this down to 75 or 80%?

larsjohnsen: Yeah, I guess that suits most displays. Updated to 80%.

erikdesjardins:
Looks good, thanks!

larsjohnsen:
Related: Shortcut for removing height restriction (useful for multi-picture images) b7c304c57d8fae022bffcfa251d6e1f915259788

PR #3088, opened 2016-06-18
Add tests for some utility functions, notably pageType regexes

erikdesjardins:


package.json
```diff
@@ -35,7 +35,8 @@
     ],
     "require": [
       "babel-core/register"
-    ]
+    ],
+		"babel": "inherit"
```
mc10: Something happened to indentation here...

erikdesjardins: Looks like tabs instead of spaces, thanks

PR #3092, opened 2016-06-19
Remove subreddit style before pageload

### LINKED ISSUE #3037: add body.res-srstyle-enabled in stylesheet loader if applicable

### LINKED ISSUE #3079: Remove subreddit style asap
eg when head is all ready, or try MutationObserver on head element to watch for it per @Farow

http://redd.it/4oduft

erikdesjardins:
Fixes #3079, fixes #3037

There is a short flicker when auto-disabling subreddit style on subreddits that don't support nightmode, on the first load (before it's cached).

Otherwise (i.e. for subreddits with the style manually disabled), it's instant.

### DIFF
--- a/lib/modules/menu.js
+++ b/lib/modules/menu.js
@@ -1,6 +1,5 @@
 import { $ } from '../vendor';
 import * as Hover from './hover';
-import * as StyleTweaks from './styleTweaks';
 
 export const module = {};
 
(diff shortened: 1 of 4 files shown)


lib/modules/menu.js
```diff
@@ -104,7 +103,7 @@ export function onClickMenuButton(callback, isImportant) {
 function addLegacyStyling() {
 	const gearIcon = RESPrefsLink.querySelector('.gearIcon');
 	const backgroundImage = window.getComputedStyle(gearIcon).backgroundImage;
-	if (StyleTweaks.isSubredditStyleEnabled() && backgroundImage && backgroundImage !== 'none') {
```
erikdesjardins: Note: this check existed because styleTweaks would take way too long to remove the subreddit style.

Now that that happens during loadDynamicOptions/beforeLoad, it's always removed before this code runs (in afterLoad), so `getComputedStyle` is accurate.

PR #3097, opened 2016-06-20
Preload media when using mediaBrowseMode

larsjohnsen:
Preloads `browsePreloadCount` number of expandos when using mediaBrowseMode (keyboard navigation).

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -41,6 +41,7 @@ import {
 	filterMap,
 	forEachChunked,
 	forEachSeq,
+	idleThrottle,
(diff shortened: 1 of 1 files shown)


lib/modules/showImages.js
```diff
@@ -1029,6 +1048,20 @@ function getMediaButtonInfo(options) {
 	};
 }
 
+let lastPreloadIndex = 0;
+function preloadMedia(pieces: Array<any>) {
```
erikdesjardins: `Array<any>` should be able to be removed after rebasing onto master

PR #3113, opened 2016-06-25
Keyboard navigation: Cleanup

larsjohnsen:


lib/modules/keyboardNav.js
```diff
@@ -837,194 +869,56 @@ module.followSubreddit = function(newWindow) {
...
 
-	_moveUp(selected.thing);
-};
+function move(target, { scrollStyle = module.options.scrollStyle.value, tryMediaBrowse = false } = {}) {
+	const selected = SelectedEntry.selectedThing();
 
-module.moveUpComment = function() {
-	module.moveUp();
-};
+	if (target instanceof Function) {
```
erikdesjardins: Nit: `typeof target === 'function'` is usually used to do this (and is technically more accurate, though that doesn't really matter here)

lib/modules/keyboardNav.js
```diff
@@ -158,51 +159,71 @@ const commands = {
...
+				thing.getNextSibling({ direction: 'down' }) || thing.getNext({ direction: 'down' }),
+				{ scrollStyle: 'legacy' }
+			);
+		},
 	},
 	moveUpThread: {
 		include: ['comments'],
 		value: [75, true, false, true], // shift-alt-k
 		description: 'Move to the topmost comment of the previous thread (in comments).',
+		callback() { move(thing => thing.getThreadTop().getNextSibling({ direction: 'up' })); },
```
erikdesjardins: We need `scrollStyle: 'legacy'` for this and `moveDownThread` (looks like that's what it was before?)

larsjohnsen: You're right; updated.

PR #3120, opened 2016-06-26
guard against "source" button being added mutiple times

andytuba:
Bandaid. I thought we put one on a while ago already, but shouldn't hurt to add this too. ¯\\\_(ツ)_/¯

https://www.reddit.com/r/RESissues/comments/4pu2lf/bug_source_button_multiplied/

### DIFF
--- a/lib/modules/sourceSnudown.js
+++ b/lib/modules/sourceSnudown.js
@@ -40,6 +40,7 @@ function attachViewSourceButtons(entry) {
 		}
 
 		$menus::forEachChunked(item => {
+			if (item.querySelector('li.viewSource')) return; // guard against adding "source" button multiple times for whatever buggy reason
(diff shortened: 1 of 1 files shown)


lib/modules/sourceSnudown.js
```diff
@@ -40,6 +40,7 @@ function attachViewSourceButtons(entry) {
 		}
 
 		$menus::forEachChunked(item => {
+			if (item.querySelector('li.viewSource')) return; // guard against adding "source" button multiple times for whatever buggy reason
```
erikdesjardins: `item` needs to be `$(item).closest('.flat-list')` or something because it's the comments link, not the entire menu

andytuba: I should probably reuse the `$(item).closest('li')` and check for `$that.siblings('.viewSource')`

PR #3123, opened 2016-06-27
"reddit" media host, for image posts that link to comments page

### LINKED ISSUE #3116: Support for i.redd.it images in expando.
Appears images on i.redd.it arent expanding within RES show images. See: https://www.reddit.com/domain/i.redd.it/

Post titles appear to be self-post but are infact images, this is for images uploaded via Reddit's image uploader. e.g https://www.reddit.com/r/funny/comments/4pvo05/cant_place_the_building_here/

andytuba:
Closes #3116 

Targets posts linking to i.redd.it where the title link points at the comments page (for anonymous users) instead of the direct image link.

:warning: Only tested on https://www.reddit.com/by_id/t3_4r2r9g,t3_4r2r9g/?feature=preview_redirect_only  (fails on the second link for some reason). still needs testing in comments.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -617,8 +617,8 @@ async function resolveMediaLink(element, thing) {
 	}
 
 	if (commentsLink) {
-		const [ , , id ] = regexes.comments.exec(element.pathname);
-		const [ , data ] = await getPostMetadata({ id });
+		const [, , id] = regexes.comments.exec(element.pathname);
+		const [, data] = await getPostMetadata({ id });
 		return new URL(data.url, element.href);
 	}
 
@@ -653,8 +653,6 @@ async function checkElementForMedia(element) {
 	const mediaLink = await resolveMediaLink(element, thing);
 	const mediaInfo = getMediaInfo(element, mediaLink);
 	if (!mediaInfo) return;
-
-	thing = thing || new Thing(element);
 
 	const inText = !!$(element).closest('.md, .search-result-footer')[0];
 
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -613,7 +613,7 @@ async function resolveMediaLink(element, thing) {
 	const dataUrl = new URL(thing.$thing.data('url'), element.href);
 	const commentsLink = thing.getCommentsLink();
 	if (dataUrl && commentsLink && commentsLink.href !== dataUrl.href) {
-		return new dataUrl;
+		return dataUrl;
 	}
 
 	if (commentsLink) {
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -20,6 +20,7 @@ import {
 	expandos,
 	primaryExpandos,
 	Thing,
+	getPostMetadata,
 	addCSS,
 	batch,
 	CreateElement,
@@ -33,6 +34,7 @@ import {
 	waitForEvent,
 	watchForElement,
 	scrollToElement,
+	regexes,
 } from '../utils';
 import { addURLToHistory, ajax, isPrivateBrowsing, openNewTab } from '../environment';
 import * as SettingsNavigation from './settingsNavigation';
@@ -598,7 +600,23 @@ async function convertGifToVideo(options) {
 	}
 }
 
-function getMediaInfo(mediaLink) {
+async function getMediaLink(element, thing) {
+	const dataUrl = new URL(thing.$thing.data('url'), element.href);
+	const commentsLink = thing.getCommentsLink();
+	if (dataUrl && commentsLink && commentsLink.href !== dataUrl.href) {
+		return new dataUrl;
+	}
+
+	if (commentsLink) {
+		const [ , , id ] = regexes.comments.exec(element.pathname);
+		const [ , data ] = await getPostMetadata({ id });
+		return new URL(data.url, element.href);
+	}
+
+	return element;
+}
+
+function getMediaInfo(element, mediaLink) {
 	const matchingHosts = _.compact([
 		domainModuleMap[getMainDomainComponent(mediaLink.hostname)],
 		siteModules.default,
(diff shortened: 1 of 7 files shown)


andytuba:
Should I add this as an option here? https://www.reddit.com/r/Enhancement/comments/4q4n44/feature_request_direct_link_to_images_instead_of/

erikdesjardins:
@andytuba I think so; I'd be amenable to enabling it by default.

lib/utils/location.js
```diff
@@ -45,7 +45,7 @@ export function matchesPageLocation(includes, excludes) {
 /* eslint-disable key-spacing */
 export const regexes = {
 	frontpage:        /^\/(?:hot|new|rising|controversial|top)?(?:\/|$)/i,
-	comments:         /^\/(?:r\/([\w\.]+)\/)?comments(?:\/|$)/i,
+	comments:         /^\/(?:r\/([\w\.]+)\/)?comments(?:\/([a-z0-9]+)\/\w*)?(?:\/|$)/i,
```
erikdesjardins: I know this is a WIP but:

Probably drop the `\/\w*`, since that can be left out or mangled (with any character) and reddit will still accept it.

andytuba: Good point. How about `/^\/(?:r\/([\w\.]+)\/)?comments(?:\/([a-z0-9]+)/i`?

erikdesjardins: Why remove the `(?:\/|$)`? Are there any valid urls where the comment id or `comments` isn't followed by a slash or the end of the pathname? (it avoids matching a path section like `/comments_something_something`, though that may not be relevant here)

The comment id is optional in the current regex so that it matches `/r/enhancement/comments` (legacy behaviour). We could change that, but I'm not sure what a better page type would be and I don't really want to add a new one.

erikdesjardins: Also n.b. [we have tests for these](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/254de3e52b177cb92f57a367954c0eb4af2a547d/lib/utils/__tests__/location.js#L46) which you'll have to change, at the very least to add the comment id capture group

andytuba: Ah, I meant, I should definitely remove the `\/` and not just the `\/w+`.  Tests will be fixed up too!

lib/utils/thingMetadata.js
```diff
@@ -0,0 +1,16 @@
...
+export const getPostMetadata = batch(async requests => {
+	const byId = requests.map(r => r.id)
+		.map(id => `t3_${id}`).join(',');
+
+	const { data: { children } } = await ajax({
+		url: `/by_id/${byId}.json`,
+		type: 'json',
+	});
+
+	return children.reduce((map, { data }) => map.set(data.id, data), new Map());
```
erikdesjardins: I think you should be able to `return children.map(c => c.data)`, since results are returned in order (and not deduped, as you've seen 😜 )

lib/modules/showImages.js
```diff
@@ -598,7 +605,27 @@ async function convertGifToVideo(options) {
...
 
-function getMediaInfo(mediaLink) {
+async function resolveMediaLink(element, thing) {
+	if (!module.options.expandoCommentRedirects.value) {
+		return element;
+	}
+
+	const dataUrl = new URL(thing.$thing.data('url'), element.href);
+	const commentsLink = thing.getCommentsLink();
+	if (dataUrl && commentsLink && commentsLink.href !== dataUrl.href) {
```
erikdesjardins: Nit: `dataUrl` is always truthy

lib/utils/thingMetadata.js
```diff
@@ -0,0 +1,16 @@
...
+}, { size: 50 });
```
erikdesjardins: If you add `?limit=100` to the url you should be able to make this `100`, [e.g.](https://www.reddit.com/by_id/t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g,t3_4r2r9g.json?limit=100)

lib/utils/thingMetadata.js
```diff
@@ -0,0 +1,16 @@
...
+
+
```
erikdesjardins: Nit: there's an extra newline here

lib/modules/showImages.js
andytuba: Dang, I should check that `.data('url')` is valid instead.

PR #3132, opened 2016-07-06
TEMP: try to prevent babel-traverse from updating past 6.9.0 and causing issues

erikdesjardins:
Until https://github.com/babel/babel/pull/3557 is merged and released.

This works for fresh installs, but it may not work if you've already installed a version of babel-traverse with the issue.

`npm-shrinkwrap` would guarantee this, but hopefully it'll be fixed soon and we won't need to resort to that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "babel-plugin-transform-runtime": "6.9.0",
     "babel-preset-es2015": "6.9.0",
     "babel-preset-stage-0": "6.5.0",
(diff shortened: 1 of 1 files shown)


hzoo:
Will make another release this week

erikdesjardins:
Thanks! Don't rush it on my behalf; this works for now

hzoo:
Ok done 😄  https://github.com/babel/babel/releases/tag/v6.13.0

PR #3144, opened 2016-07-08
Misc fixes and optimiziations

larsjohnsen:


lib/modules/betteReddit.js
```diff
@@ -167,19 +162,6 @@ module.go = function() {
 		}
 	}
 
-	if (module.options.scoreHiddenTimeLeft.value && isPageType('comments')) {
-		$('.tagline').on('mouseenter', 'span:contains([)', function() {
```
erikdesjardins: I see why this is super inefficient, but I'm not sure that we should just remove it.

I think the usual pattern for doing stuff to a bunch of comments:

```js
applyScoreHiddenTimeLeft();
watchForElement(...);
// ...
function applyScoreHiddenTimeLeft() {
    querySelectorAll('.score-hidden')::forEachChunked(...)
}
```

would be a viable replacement. Might as well run it in afterLoad, since it's not noticeable until you hover.

lib/modules/filteReddit.js
```diff
@@ -869,15 +869,15 @@ function scanEntries(ele) {
 			// Run async (slow) filters separately
 			// Note: These won't be counted in numFiltered
 			(async () => {
-				filtered = removeVisitedLinksActive && await isURLVisited(postLink);
+				if (postLink) filtered = removeVisitedLinksActive && await isURLVisited(postLink.href);
```
erikdesjardins: Hmmm, now that I look at this a second time I think my concerns were misplaced.

`removeVisitedLinksActive` is always false on pageload, so we'll never hit an `await` during the initial filter, where performance really matters.

So you could put this back to how it was, without the async IIFE.

lib/modules/xPostLinks.js
```diff
@@ -1,5 +1,6 @@
 import { $ } from '../vendor';
-import { Thing, forEachChunked, watchForElement } from '../utils';
+import { Thing, forEachChunked, isPageType, watchForElement } from '../utils';
+
```
erikdesjardins: Nit: extra newline

lib/modules/filteReddit.js
larsjohnsen: I'll keep it as is for now since I have use for it in #3042

PR #3151, opened 2016-07-13
Fixes, tweaks

larsjohnsen:


package.json
```diff
@@ -51,6 +51,7 @@
     "jquery-sortable": "0.9.13",
     "konami": "1.3.3",
     "lodash": "4.13.1",
+    "element-resize-detector": "1.1.5",
```
erikdesjardins: Nit: npm will alphabetize this the next time someone runs a command with `--save`, so let's put it in alphabetical order now

PR #3153, opened 2016-07-14
Update dependencies to latest versions

BenMcGarry:
babel ignored due to build problems. No problems detected on tests.


erikdesjardins:
This significantly changes the coverage, but I believe it's more correct now. (24% was certainly wrong, at any rate)

PR #3155, opened 2016-07-15
keyboardNav: New shortcut: Move to parent's next sibling

larsjohnsen:
Combines `moveToParent` and then `moveDownSibling`.

I don't think `moveUpParentSibling` is necessary, as it would be the same as `moveToParent`. An symmetrical shortcut (`alt+k`) would be nice but I couldn't think of any good way to implement alternative shortcuts.

This reuses the `threadScrollStyle` option, as the action is similar. The option should probably also be used on `moveToParent`.

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -311,7 +311,12 @@ const commands = {
 		include: ['comments'],
 		value: [80, false, false, false], // p
 		description: 'Move to parent (in comments).',
-		callback() { move(thing => thing.getParent()); },
+		callback() {
+			move(
+				thing => thing.getParent(),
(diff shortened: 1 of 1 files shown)


erikdesjardins:
Hmm yeah then we start to blur the line between `threadScrollStyle`/`siblingScrollStyle`.

Maybe replace `threadScrollStyle` and `siblingScrollStyle` with `nonLinearScrollStyle` (or similar name), and use it for `move*Thread`, `move*Sibling`, `moveToParent`, and `move*ParentSibling`. I think that would still satisfy the need to change thread/sibling style.

>would be the same as `moveToParent`

We could have it move to the sibling before the parent, i.e. with:

```
A
  B
    C
  D
    E
    F <- selected
```

It would move to `B`.

PR #3159, opened 2016-07-16
showImages: Use fallback mechanism to update placeholder

### LINKED ISSUE #3158: .gifvs have invalid placeholder height in Firefox
@andytuba reports it happens always, I'm seeing it happening on any reopened gifv

image

larsjohnsen:
In Firefox the video bounding rect is not updated before the `loadedmetadata` event is emitted. This PR implmenets a fallback mechanism.

Fixes #3158

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1007,10 +1007,6 @@ function generateImage(options) {
 	element.ready = waitForEvent(image, 'load', 'error');
 
 	element.emitResizeEvent = () => { image.dispatchEvent(new CustomEvent('mediaResize', { bubbles: true })); };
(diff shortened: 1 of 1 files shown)


erikdesjardins:
>Edit: Nope, the browsers still doesn't agree.

This seems to work for me -- can you link a post where it doesn't work?

larsjohnsen:
@erikdesjardins When maxWidth or maxHeight limited the size, the aspect ratio of the element became incorrect.

I've amended the commit to use `elementResizeDetector` as fallback mechanism.

erikdesjardins:
Ah, okay. Thanks for the fix!

Issue #2794, opened 2016-03-17
New icon everywhere


andytuba:
- [ ] res.com favico
- [ ] Inside browser packages

githue:
- [ ] Remove the CSS filter from the console logo in nightmode.

mc10:
Related: #1178.

andytuba:
Related: #1071

erikdesjardins:
Favicon added to gh-pages site https://github.com/Reddit-Enhancement-Suite/Reddit-Enhancement-Suite.github.io/commit/cb9b1af477698471d8a14f42b209dfef52d7dbc1

PR #3165, opened 2016-07-17
UFO settings console icon

andytuba:
screen shot 2016-07-17 at 12 35 45 pm
screen shot 2016-07-17 at 12 38 25 pm

#2794


andytuba:
@githue, do you still think we should [remove the nightmode filter](https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2794#issuecomment-198122504)?

PR #3185, opened 2016-07-20
Version for beta.

BenMcGarry:


andytuba:
Context for close: 
> erikdesjardins: we'll use npm version and push directly to master with a tag to trigger the build

PR #3196, opened 2016-07-24
Refactor Chrome permissions message

erikdesjardins:
Now it opens an alert before opening the native popup, instead of only warning after the user denies it.

image

### DIFF
--- a/lib/environment/permissions.js
+++ b/lib/environment/permissions.js
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import { invokeAll } from '../utils';
+import { Alert, mutex } from '../utils';
 import { sendMessage } from 'browserEnvironment';
 
 function filterPerms(perms) {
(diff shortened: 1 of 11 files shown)


lib/environment/permissions.js
```diff
@@ -8,25 +8,39 @@ function filterPerms(perms) {
...
+		if (hasPermissions) return;
+
+		const urlStripRe = /((?:\w+\.)+\w+)(?=\/|$)/i;
+
+		await Alert.open(`
+			<p>To create expandos, RES needs permission to access the API(s) at:</p>
+			<br>
+			<p>${origins.map(url => `<code>${urlStripRe.exec(url)[0]}</code>`).join(', ')}</p>
+			<br>
+			<p>Be assured RES does not access any of your information on these domains - it only accesses the API.</p>
```
erikdesjardins: Wording? More info?

BenMcGarry: The only thing I'd recommend on that line is to put access/modify any of your information.

PR #3198, opened 2016-07-25
Firefox: append `beta` to beta version numbers

erikdesjardins:
For future release automation: since they'll be deployed to the same listing as stable RES, they need the suffix.

### DIFF
--- a/firefox/package.json
+++ b/firefox/package.json
@@ -6,7 +6,7 @@
 	"author": "honestbleeps",
 	"homepage": "http://redditenhancementsuite.com/",
 	"license": "GPL-3.0",
-	"version": "{{prop?version!../package.json}}",
+	"version": "{{to-string!./getVersion}}",
 	"main": "{{./background.entry.js}}",
 	"permissions": {
 		"private-browsing": true
--- a/build/isBetaVersion.js
+++ b/build/isBetaVersion.js
@@ -1,0 +1,5 @@
+/* eslint-disable import/no-commonjs */
+
+const semver = require('semver');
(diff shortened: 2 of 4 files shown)


firefox/package.json
```diff
@@ -6,7 +6,7 @@
 	"author": "honestbleeps",
 	"homepage": "http://redditenhancementsuite.com/",
 	"license": "GPL-3.0",
-	"version": "{{prop?version!../package.json}}",
+	"version": "{{to-string!./getVersion}}",
```
erikdesjardins: Note: `jpm` warns about the version because it's not valid semver, despite the fact that AMO _doesn't even use_ semver!

For reference, [here are the docs](https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Maintenance) suggesting to use `1.1alpha1`, which isn't valid (`1.1.0-alpha.1` would be), and a [published extension](https://addons.mozilla.org/en-US/firefox/addon/ublock-origin/#beta-channel) with version number `1.7.7b15`.

Ajedi32: FYI, JPM's requirement for Semver is documented here: https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json

erikdesjardins: Interesting. But amusingly, that page uses `0.1` in its example manifest, the "invalid semver" error message printed by jpm suggests `1.0.0rc4`, and AMO doesn't even let you submit extensions with a proper semver beta version:
image
Make up your mind Mozilla!

Issue #3194, opened 2016-07-23
indexedDB works with `jpm run` but not `jpm xpi`


erikdesjardins:
I have no words.

![image](https://cloud.githubusercontent.com/assets/7673145/17079793/86fa9478-50e8-11e6-86ca-3c2adec872d8.png)

erikdesjardins:
This was happening because my main Firefox profile was somehow corrupted, see [bug 994918](https://bugzilla.mozilla.org/show_bug.cgi?id=944918).

`jpm run` was unaffected because it always uses a fresh profile. Deleting and recreating my main profile fixed it for `jpm xpi`.

I really hope this isn't a common thing, because there's nothing we can do besides telling users to delete their Firefox profiles...

andytuba:
We could.. Um..  Put in an option to go back to simple storage :sob:

PR #3199, opened 2016-07-25
Firefox: show alert when db cannot be created

erikdesjardins:
This should give us an idea if #3194 is common or not (in the beta); hopefully we don't have to fall back to simple-storage and its horrible instability :cold_sweat:.

image

### DIFF
--- a/firefox/background.entry.js
+++ b/firefox/background.entry.js
@@ -138,7 +138,19 @@ let db;
 	}
 }
 
-addListener('storage', ([operation, key, value]) => {
+function storageFailureAlert(worker) {
+	sendMessage('alert', `
+		<p><b>An error occurred while creating the IndexedDB database.</b></p>
+		<p>Reddit Enhancement Suite will not function.</p>
+		<p>Your Firefox profile may be corrupted (<a href="https://bugzilla.mozilla.org/show_bug.cgi?id=944918">Bug 944918</a>).</p>
+		<br>
+		<p>Please report this to the relevant beta thread or /r/RESissues.</p>
+	`, worker);
+}
+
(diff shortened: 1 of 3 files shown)


firefox/background.entry.js
```diff
@@ -138,7 +138,19 @@ let db;
...
+		<p><b>An error occurred while creating the IndexedDB database.</b></p>
+		<p>Reddit Enhancement Suite will not function.</p>
+		<p>Your Firefox profile may be corrupted (<a href="https://bugzilla.mozilla.org/show_bug.cgi?id=944918">Bug 944918</a>).</p>
+		<br>
+		<p>Please report this to the relevant beta thread or /r/RESissues.</p>
+	`, worker);
+}
+
+addListener('storage', ([operation, key, value], worker) => {
+	if (!db) storageFailureAlert(worker);
```
erikdesjardins: Three or four storage requests are made before init crashes, so this will open multiple alerts in series.

But I don't think that's a problem because you're going to be immediately rolling back/disabling the extension if this happens (memoizing by worker and current URL would be mildly complex for little gain).

Ajedi32:
I wonder if we could have RES automatically report this somewhere. It'd be nice to have some hard data on how common this is.

erikdesjardins:
Redditors can be, \*ahem\* "security-conscious" (which is why we have zero telemetry), so probably not. We might have opt-in telemetry in the future, but that's a fairly large project.

Hopefully this won't be a problem for very long though, what with the planned stabilization of WebExtensions in Firefox 48, where _hopefully_ their implementation of `chrome.storage` will be stable.

PR #3211, opened 2016-07-28
showImages addDragListener: Safari compatibility

### LINKED ISSUE #3209: Image drag-to-zoom is broken in Safari
showImages' image zoom functionality (i.e., "drag to zoom" on expando images) is completely broken in Safari.

The culprit seems to be commit d16af250f60710ae5f45795e38db6edb4b115951 and subsequent commits to showImages.js by @larsjohnsen. Specifically, in the `addDragListener()` function (formerly `makeMediaZoomable()`), that commits adds code which checks `e.buttons`. Commit 35261f95314da9017b1e216540c3c9055b27d9e1 additionally adds code which checks `e.movementX `, and `e.movementY`. None of these attributes exist on Safari MouseEvent objects, so the drag event handler always bails out before doing any work because it thinks the drag has ended.

### LINKED ISSUE #3210: Support Safari, or stop advertising that you support Safari
Safari is not actually supported by the dev team in any way. As far as I can tell, no core dev runs Safari, even for testing purposes. Further, no reference sources are consulted when making code changes to see if they are Safari compatible.

As evidence, see #3192 and #3209. The former completely broke the extension in Safari for over three months because of a blatant bug in Safari-specific code, and no developer noticed; the latter broke drag-to-zoom on images due to relying on Safari incompatible mouse event attributes.

[MDN](https://developer.mozilla.org/en-US/) documents cross-platform support for nearly every web feature, and a quick glance at [`MouseEvent.buttons`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons), for example, shows that it is not supported in Safari. Checking this could have prevented breaking click-to-zoom in Safari.

Looking at @larsjohnsen's recent commits, you'll see "checked on Chrome, Firefox and Edge", and no sign of Safari. (Not picking on him, but using him as an example, since he committed the Safari incompatible code in click-to-zoom.) There is no CI server or other automated testing infrastructure to catch the regression either, for devs that don't have Macs. And apparently, no other dev ever thinks to start up Safari and install RES to check if it still works by hand.

If RES supports Safari, then support Safari. If, on the other hand, Safari isn't considered during development, or even run, then stop saying RES supports Safari.

larsjohnsen:
Safari does not expose `MouseEvent.movementX`, `MouseEvent.movementY`, and `MouseEvent.buttons`, which my implementation required. This provides a workaround in order to make `addDragListener` Safari compatible.

Please note that I have not had the opportunity to test this on Safari yet.

Fixes #3209

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1403,11 +1403,11 @@ function addDragListener({ media, atShiftKey, onStart, onMove }) {
 	let isActive, hasMoved, lastX, lastY;
 
 	const handleMove = frameDebounce(e => {
-		if (!e.movementX && !e.movementY) {
+		const movementX = e.clientX - lastX;
+		const movementY = e.clientY - lastY;
+
+		if (!movementX && !movementY) {
 			// Mousemove may be triggered even without movement
-			return;
-		} else if (e.buttons !== 1) {
-			stop();
 			return;
 		} else if (atShiftKey !== e.shiftKey) {
 			isActive = false;
@@ -1422,25 +1422,24 @@ function addDragListener({ media, atShiftKey, onStart, onMove }) {
 			document.body.classList.add('res-media-dragging');
 		}
 
-		onMove(e.clientX, e.clientY, e.clientX - lastX, e.clientY - lastY);
+		onMove(e.clientX, e.clientY, movementX, movementY);
 		({ clientX: lastX, clientY: lastY } = e);
 	});
 
 	function handleClick(e) {
 		if (hasMoved) e.preventDefault();
 		document.removeEventListener('click', handleClick);
-
-		stop();
 	}
 
 	function stop() {
 		document.body.classList.remove('res-media-dragging');
 
 		document.removeEventListener('mousemove', handleMove);
+		document.removeEventListener('mouseup', stop);
 	}
 
 	function initiate(e) {
-		if (e.buttons !== 1) return;
+		if (e.button !== 0) return;
 
 		({ clientX: lastX, clientY: lastY } = e);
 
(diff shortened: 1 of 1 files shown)


lib/modules/showImages.js
```diff
@@ -1400,13 +1400,16 @@ function setMediaMaxSize(media) {
 }
 
 function addDragListener({ media, atShiftKey, onStart, onMove }) {
-	let isActive, hasMoved, lastX, lastY;
+	let isActive, isReleased, hasMoved, lastX, lastY;
 
 	const handleMove = frameDebounce(e => {
-		if (!e.movementX && !e.movementY) {
+		if (
+			process.env.BUILD_TARGET !== 'safari') && // Safari lacks e.movement(X|Y), e.buttons
```
larsjohnsen: This may sometimes cause an issue where clicking on the image does not open it, but I don't know if this condition is ever met in Safari.

lib/modules/showImages.js
```diff
@@ -1400,13 +1400,16 @@ function setMediaMaxSize(media) {
 }
 
 function addDragListener({ media, atShiftKey, onStart, onMove }) {
-	let isActive, hasMoved, lastX, lastY;
+	let isActive, isReleased, hasMoved, lastX, lastY;
 
 	const handleMove = frameDebounce(e => {
-		if (!e.movementX && !e.movementY) {
+		if (
+			process.env.BUILD_TARGET !== 'safari' && // Safari lacks e.movement(X|Y)
```
larsjohnsen: This may sometimes cause an issue where clicking on the image does not open it, but I don't know if this condition (!e.movementX && !e.movementY) is ever met in Safari.

erikdesjardins: Yeah this prevents a click from opening the image in Safari. May have to add a `mousedown` listener to set the initial `lastX` and `lastY` and use `e.clientX - lastX` instead of `e.movementX`, if that would be equivalent.

erikdesjardins:
Confirmed to work in all browsers; thanks!

PR #3222, opened 2016-08-02
Update account switcher cookie removal for Firefox 48

### LINKED ISSUE #3052: Firefox 48 will change the signature of nsICookieManager::remove() again
https://www.reddit.com/r/RESissues/comments/4mo20g/bug_user_account_switcher_doesnt_work_in_firefox/

https://bugzilla.mozilla.org/show_bug.cgi?id=1259169

AFAICT, the fix isn't backwards-compatible.

related: #2779

erikdesjardins:
...now that it is stable.

Fixes #3052.

### DIFF
--- a/firefox/background.entry.js
+++ b/firefox/background.entry.js
@@ -59,6 +59,17 @@ function workerFor(tab) {
 	return worker;
 }
 
+/**
+ * @this {nsISimpleEnumerator}
+ * @param {nsIJSIID} iface
+ * @returns {Generator}
+ */
+function* asGenerator(iface) {
+	while (this.hasMoreElements()) {
+		yield this.getNext().QueryInterface(iface);
+	}
+}
+
 const {
 	_handleMessage,
 	sendMessage,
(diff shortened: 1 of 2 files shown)


firefox/background.entry.js
```diff
@@ -69,9 +80,17 @@ const {
 
 addCommonBackgroundListeners(addListener);
 
-addListener('deleteCookies', cookies =>
-	cookies.forEach(({ name }) => cookieManager.remove('.reddit.com', name, '/', false, false))
-);
+addListener('deleteCookies', cookies => {
+	for (const { host, name, path, originAttributes } of cookieManager.enumerator::asGenerator(Ci.nsICookie)) {
```
erikdesjardins: I'd love to be able to get the `originAttributes` without iterating through every single cookie, but I can't find any (documented) API that will let me do that (in [nsICookieManager](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsICookieManager) or [nsICookieManager2](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsICookieManager2)).

One might expect to be able to use [getCookiesFromHost()](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsICookieManager2#getCookiesFromHost()), but nope, it also requires `originAttributes`, though this is not documented:



In any case, we'll be switching to WebExtensions at some point in the (hopefully) near future and this code will disappear.

PR #3226, opened 2016-08-04
Added support for Super Stack images in the inline image viewer.

Dimes:
I've added a new host file in `lib/modules/hosts` that supports images from [Super Stack](https://superstack.io).

I've tested this in Firefox, Chrome, and Safari. The code is linted.

### DIFF
--- a/lib/modules/hosts/superstack.js
+++ b/lib/modules/hosts/superstack.js
@@ -3,13 +3,13 @@ import { ajax } from '../../environment';
 const IMAGE = 'IMAGE';
 const GALLERY = 'GALLERY';
 
-const superstack = {
+export default {
 	moduleID: 'superstack',
 	name: 'superstack',
 	domains: ['superstack.io'],
 	logo: '//superstack.io/favicon.ico',
-	detect({ href }) {
-		return (/^https?:\/\/(?:www\.)?superstack\.io\/v\/(\w+)(?:[/#]|$)/).exec(href);
+	detect({ pathname }) {
+		return (/^\/v\/(\w+)(?:\/|$)/).exec(pathname);
 	},
 	async handleLink(href, [, id]) {
 		const data = await ajax({
@@ -23,20 +23,15 @@ const superstack = {
 			throw new Error('Super Stack API error: no photos returned');
 		}
 
-		// Check if this is a single image or an album
-		if (data.photos.length === 1) {
-			return {
+		return {
+			type: GALLERY,
+			title: data.albumTitle,
+			src: data.photos.map(photo => ({
+				src: photo.source,
 				type: IMAGE,
-				src: data.photos[0].source,
-			};
-		} else {
-			return {
-				type: GALLERY,
-				title: data.albumTitle || data.photos[0].title,
-				src: data.photos.map(photo => photo.source),
-			};
-		}
+				title: photo.title,
+				caption: photo.description,
+			})),
+		};
 	},
 };
-
-export default superstack;
--- a/lib/modules/hosts/superstack.js
+++ b/lib/modules/hosts/superstack.js
@@ -9,7 +9,7 @@ const superstack = {
 	domains: ['superstack.io'],
 	logo: '//superstack.io/favicon.ico',
 	detect({ href }) {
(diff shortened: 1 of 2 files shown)


lib/modules/hosts/superstack.js
```diff
@@ -0,0 +1,42 @@
+import { ajax } from '../../environment';
+
+const IMAGE = 'IMAGE';
+const GALLERY = 'GALLERY';
+
+const superstack = {
```
erikdesjardins: This can be: `export default {`; the `superstack` var isn't used anywhere besides the export

lib/modules/hosts/superstack.js
```diff
@@ -0,0 +1,42 @@
...
+	moduleID: 'superstack',
+	name: 'superstack',
+	domains: ['superstack.io'],
+	logo: '//superstack.io/favicon.ico',
+	detect({ href }) {
```
erikdesjardins: This can be simplified to: `detect: ({ pathname }) => (/^\/v\/(\w+)(?:\/|$)/).exec(pathname),`

lib/modules/hosts/superstack.js
```diff
@@ -0,0 +1,42 @@
...
+		});
+
+		if (data.error) {
+			throw new Error(`Super Stack API error: ${data.error}`);
+		} else if (!data.photos || data.photos.length < 1) {
+			throw new Error('Super Stack API error: no photos returned');
+		}
+
+		// Check if this is a single image or an album
+		if (data.photos.length === 1) {
```
erikdesjardins: Albums with a single image are automatically treated as images (we're actually fixing the icon for that as we speak), so this can be removed

lib/modules/hosts/superstack.js
```diff
@@ -0,0 +1,42 @@
...
+			return {
+				type: IMAGE,
+				src: data.photos[0].source,
+			};
+		} else {
+			return {
+				type: GALLERY,
+				title: data.albumTitle || data.photos[0].title,
+				src: data.photos.map(photo => photo.source),
```
erikdesjardins: Individual images can have a `title` and `caption`, and should have a `type` ([see imgur](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/9644a2a0e49d8005c0b98c2af4ea060d2c30d20b/lib/modules/hosts/imgur.js#L181-L185); the docs can probably be improved)

lib/modules/hosts/superstack.js
```diff
@@ -0,0 +1,42 @@
...
+		// Check if this is a single image or an album
+		if (data.photos.length === 1) {
+			return {
+				type: IMAGE,
+				src: data.photos[0].source,
+			};
+		} else {
+			return {
+				type: GALLERY,
+				title: data.albumTitle || data.photos[0].title,
```
erikdesjardins: Due to the comment on the next line, this should be `title: data.albumTitle`

safari/Info.plist
```diff
@@ -60,6 +60,7 @@
 				<string>miiverse.nintendo.net</string>
 				<string>api.gyazo.com</string>
 				<string>codepen.io</string>
+				<string>superstack.io</string>
```
erikdesjardins: Usually if we need to add more permissions to Safari we also need to add them for Chrome/Edge. Are you sure that it won't work without this?

lib/modules/hosts/superstack.js
Dimes: Changed.

lib/modules/hosts/superstack.js
Dimes: Changed.

lib/modules/hosts/superstack.js
Dimes: Changed.

lib/modules/hosts/superstack.js
Dimes: Added titles and captions to the photos.

lib/modules/hosts/superstack.js
Dimes: Changed.

safari/Info.plist
Dimes: You are correct, this was not required.

erikdesjardins:
Thanks!

PR #3228, opened 2016-08-04
Transform single-image gallery (albums) expandos to images

### LINKED ISSUE #3227: Treat single-image galleries/albums as single images
@larsjohnsen, does one of your show image refactors already handle this?

andytuba:
Fixes #3227


andytuba:
This is not the best solution, and @larsjohnsen reports he has a better solution coming.

PR #3245, opened 2016-08-07
Tweak scrollStyle middle

larsjohnsen:
The previous solution missed the reason why scrolling in certain situations is not desirable, and as such did not work as intended all the time.

### DIFF
--- a/lib/utils/dom.js
+++ b/lib/utils/dom.js
@@ -131,13 +131,13 @@ function padBottom(scrollTop, viewportHeight) {
 	}
 }
 
-export function scrollToElement(element, { scrollStyle, direction }) {
+export function scrollToElement(element, { scrollStyle, direction: selectedDirection }) {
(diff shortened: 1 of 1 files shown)


lib/utils/dom.js
```diff
@@ -150,16 +150,16 @@ export function scrollToElement(element, { scrollStyle, direction }) {
...
 		scrollTo(0, top);
 	} else if (scrollStyle === 'middle') {
-		// Do not scroll if new element position does not exceed current middle point
-		if (
-			(direction === 'up' && target.top > viewport.height / 2) ||
-			(direction === 'down' && target.bottom < viewport.height / 2)
-		) {
-			return;
+		const buffer = (viewport.height - target.height) / 2;
+		top -= buffer;
```
erikdesjardins: Nit: use a different var for this, since it's not really the top of the target anymore

PR #3257, opened 2016-08-07
styleTweaks: implement floatingSideBar in pure CSS

### LINKED ISSUE #3247: floatingSideBar also (dis)appears with each scroll
Related: #3224

erikdesjardins:
Fixes #3247

### DIFF
--- a/lib/css/modules/_styleTweaks.scss
+++ b/lib/css/modules/_styleTweaks.scss
@@ -116,3 +116,14 @@ blockquote.twitter-tweet {
 .res-styleTweaks-postTitleCapitalization-lowercase {
 	.entry a.title { text-transform: lowercase; }
 }
+
+.res-styleTweaks-floatingSideBar {
+	body.with-listing-chooser .listing-chooser {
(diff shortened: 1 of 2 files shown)


lib/css/modules/_styleTweaks.scss
```diff
@@ -116,3 +116,14 @@ blockquote.twitter-tweet {
...
 	.entry a.title { text-transform: lowercase; }
 }
+
+.res-styleTweaks-floatingSideBar {
+	body.with-listing-chooser .listing-chooser {
+		position: fixed;
+		top: 0;
+
+		.contents {
+			margin-top: 65px;
```
erikdesjardins: Note: reddit hardcodes `65px` in [their own CSS](https://github.com/reddit/reddit/blob/3ea40a09cc38f3d62fc008cf2ba3f858f638b755/r2/r2/public/static/css/reddit.less#L9810) (and I don't think they'll change the header height out-of-the-blue), so I'm not concerned about hardcoding this.

PR #3263, opened 2016-08-09
Hide the ugly empty space above edit buttons if link is not multireddit.

pltarun:
Added :empty selector to the subreddit manager dropdown ul.

### DIFF
--- a/lib/css/modules/_subredditManager.scss
+++ b/lib/css/modules/_subredditManager.scss
@@ -177,7 +177,6 @@
 	margin-left: -5px;
 
 	> .RESShortcutsEditButtons {
-		border-top: 1px rgba(0, 0, 0, .5) solid;
 
 		.res-icon {
 			color: black;
@@ -205,6 +204,7 @@
 		overflow-y: auto;
 		max-width: 300px;
 		max-height: 500px;
+		border-bottom: 1px rgba(0, 0, 0, .5) solid;
 
 		li {
 			padding: 0;
--- a/lib/css/modules/_subredditManager.scss
+++ b/lib/css/modules/_subredditManager.scss
@@ -218,6 +218,10 @@
 				background-color: rgba(255, 255, 255, .6);
 			}
 		}
+		
+		&:empty {
+			display: none;
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pltarun:
http://i.imgur.com/f8RidFC.png
This is what I'm referring to.

andytuba:
:+1: 

When this module gets a larger refactor, e.g. a template for the dropdown menu and/or a redesign, then populating the dropdown should skip the `<ul>` if there aren't any subreddits. This PR is a good patch for the current system and can easily be applied to RES v4.6.

lib/css/modules/_subredditManager.scss
```diff
@@ -218,6 +218,10 @@
 				background-color: rgba(255, 255, 255, .6);
 			}
 		}
+
+		&:empty {
+			display: none;
```
andytuba: For bonus aesthetics, nix the extra line:

```css
& + .RESShortcutsEditButtons {
    border-top-width: 0; 
}
```

pltarun:
I've kept the border and moved it to the ul element instead of .RESShortcutsEditButtons so that it is hidden when empty.

lib/css/modules/_subredditManager.scss
```diff
@@ -205,6 +204,7 @@
 		overflow-y: auto;
 		max-width: 300px;
 		max-height: 500px;
+		border-bottom: 1px rgba(0, 0, 0, .5) solid;
```
andytuba: Sorry to throw you another curveball, but will this end up with a double border (ul & dropdown container) if the [dropdownEditButton](https://www.reddit.com/#!settings/subredditManager/dropdownEditButton) option is turned off? Could mitigate it by adding `bodyClass: true` to the `dropdownEditButton` option config and wrapping this `border-bottom` property in `.res-subredditManager-dropdownEditButton & { ... }`

pltarun:
I think this is a better solution, as it ensures that the fix is pure CSS.

andytuba:
Looks like you've covered all the edge cases I can think of now. Thanks for collaborating on this little fixup!

_waits patiently for travis green.._

Issue #3207, opened 2016-07-27
Hide child comments on non-top comments


andytuba:
https://www.reddit.com/r/Enhancement/comments/4urp6k/is_it_possible_to_collapse_a_comments_children/

thought this was already a GH issue..

pltarun:
I'm working on this. Not ready to make a pull request yet.
<img width="785" alt="screen shot 2016-08-10 at 21 16 18" src="https://cloud.githubusercontent.com/assets/5273804/17569390/e5f3d61e-5f3f-11e6-8cb2-27331a629d1c.png">
<img width="817" alt="screen shot 2016-08-10 at 21 17 09" src="https://cloud.githubusercontent.com/assets/5273804/17569391/e60faefc-5f3f-11e6-8e67-3ac157a3466f.png">

PR #3267, opened 2016-08-10
Allow hiding children of nested comments

pltarun:
Fixes issue #3207.

A link to hide child comments has been added to all comments that have children. This functionality was previously only for top level comments.

I added a new option to the module called noNested. This is enabled by default so as to preserve the current behaviour. When disabled, clicking "hide all child comments" or enabling the automatic option will collapse children of nested comments as well.

Previous wording felt a bit awkward, so I've slightly changed that too.

screen shot 2016-08-10 at 23 47 31
screen shot 2016-08-10 at 23 47 09

### DIFF
--- a/lib/modules/hideChildComments.js
+++ b/lib/modules/hideChildComments.js
@@ -32,11 +32,7 @@ module.go = () => {
 	toggleAllLink.setAttribute('title', 'Show only replies to original poster.');
 	toggleAllLink.addEventListener('click', function(e) {
 		e.preventDefault();
-		if (module.options.noNested.value) {
-			toggleComments(false, this.getAttribute('action'));
-		} else {
-			toggleComments(true, this.getAttribute('action'));
-		}
+		toggleComments(this.getAttribute('action'), module.options.noNested.value);
 		if (this.getAttribute('action') === 'hide') {
 			this.setAttribute('action', 'show');
 			this.setAttribute('title', 'Show all comments.');
@@ -51,7 +47,7 @@ module.go = () => {
 	const commentMenu = document.querySelector('ul.buttons');
 	if (commentMenu) {
 		commentMenu.appendChild(toggleButton);
-		const rootComments = document.querySelectorAll('div.commentarea > div.sitetable > div.thing > div.child > div.listing');
+		const rootComments = document.querySelectorAll('div.commentarea > div.sitetable > div.thing > div.child > div.listing, div.listing div.listing');
 		rootComments::forEachChunked(comment => {
 			const toggleButton = document.createElement('li');
 			const toggleLink = document.createElement('a');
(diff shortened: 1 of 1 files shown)


lib/modules/hideChildComments.js
```diff
@@ -6,12 +6,17 @@ export const module = {};
...
-module.description = 'Allows you to hide all comments except for replies to the OP for easier reading.';
+module.description = 'Allows you to hide child comments.';
 module.options = {
 	automatic: {
 		type: 'boolean',
 		value: false,
-		description: 'Automatically hide all but parent comments, or provide a link to hide them all?',
+		description: 'Hide children for all comments by default.',
+	},
+	noNested: {
```
erikdesjardins: I think the inverse of this would be easier to understand (no double-negative when it's disabled), i.e. "nested: Hide children for nested comments when hiding all child comments."

lib/modules/hideChildComments.js
```diff
@@ -42,7 +47,7 @@ module.go = () => {
 	const commentMenu = document.querySelector('ul.buttons');
 	if (commentMenu) {
 		commentMenu.appendChild(toggleButton);
-		const rootComments = document.querySelectorAll('div.commentarea > div.sitetable > div.thing > div.child > div.listing');
+		const rootComments = document.querySelectorAll('div.commentarea > div.sitetable > div.thing > div.child > div.listing, div.listing div.listing');
```
erikdesjardins: This can probably be simplified to something like

```js
const rootComments = document.querySelectorAll('.commentarea .listing');
```

We don't need to keep the original, very precise selector now that all comments should have a "hide child comments" button.

And maybe it should be renamed, since it's not just the `rootComments` anymore.

lib/modules/hideChildComments.js
```diff
@@ -74,10 +79,12 @@ module.go = () => {
 
 function toggleComments(action, obj) {
 	let commentContainers;
-	if (obj) { // toggle a single comment tree
-		commentContainers = $(obj).closest('.thing');
-	} else { // toggle all comments
+	if (obj === true) { // toggle only top level comments
```
erikdesjardins: Instead of using the second param for two things, I would break `toggleComments` into three functions, something like:
```js
function toggleComments(action, containers) {
    containers::forEachChunked(container => {
        ...
    });
}

function toggleIndividualComment(action, ele) {
     toggleComments(action, $(ele).closest('.thing'));
}

function toggleAllComments(action, includeChildren) {
     const containers = includeChildren ?
        document.querySelectorAll('div.commentarea div.sitetable div.thing') :
        document.querySelectorAll('div.commentarea > div.sitetable > div.thing');

    toggleComments(action, containers);
}
```

pltarun:
👍
Will do.

pltarun:
👍
Will do. 

Do we want to preserve previous behaviour (hide only top level children) or should the default be to hide everything?

andytuba:
I'd vote for preserve the previous behavior: hide only top-level children when pressing "hide all child comments".

erikdesjardins:
Merged in #3915, thanks!

PR #3273, opened 2016-08-13
hover: Indicate loading on async HoverInfoCard

larsjohnsen:
The previous content was shown until the new one is ready, which is confusing when it takes several seconds to fetch new data. This instantaneously replaces it with a loading indicator until the new content is ready.

### DIFF
--- a/lib/modules/hover.js
+++ b/lib/modules/hover.js
@@ -118,7 +118,6 @@ class Hover {
 	template = hoverDefaultTemplate;
 	_options = Hover._defaultOptions;
 	_enabled = true;
-	_loadIndicator = '<div class="RESHoverLoadIndicator"><span class="RESLoadingSpinner"></span></div>';
 
 	constructor(id) {
 		this.instanceID = id;
@@ -317,6 +316,7 @@ class Hover {
 
 class HoverInfoCard extends Hover {
 	template = hoverInfoCardTemplate;
+	_loadIndicator = '<div class="RESHoverLoadIndicator"><span class="RESLoadingSpinner"></span></div>';
 
 	_displayLoadIndicator($container) {
 		this._populate($container, null, this._loadIndicator);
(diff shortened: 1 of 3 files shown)


erikdesjardins:
This is effectively a bugfix because people keep complaining about filters loading then getting reset when they see the red +filter from the last time (that's not what's happening!), so pushing this forward to 4.7.x

lib/modules/hover.js
```diff
@@ -118,6 +118,7 @@ class Hover {
 	template = hoverDefaultTemplate;
 	_options = Hover._defaultOptions;
 	_enabled = true;
+	_loadIndicator = '<div class="RESHoverLoadIndicator"><span class="RESLoadingSpinner"></span></div>';
```
erikdesjardins: Probably move this to `HoverInfoCard` or inline it; it's not particularly generic over the class (e.g. I imagine you would want more/different elements if it were to be used for `HoverDropdownList`)

lib/modules/hover.js
```diff
@@ -203,6 +204,7 @@ class Hover {
 		const $container = this.getContainer();
 
 		if (this._callback) {
+			if (this._displayLoadIndicator) this._displayLoadIndicator($container);
```
erikdesjardins: Nit: make `_displayLoadIndicator` an empty method on `Hover` and remove the null-check (making it more obvious that this is something that a child class can/should override)

Also, _maybe_ call it `_onOpen` or `_beforeOpen` or something (i.e. why/when it's called rather than what the child class should do). Though I'm not sure if that's overly generic, since we'll probably never use it for anything but a loading icon, so feel free to leave it as-is.

erikdesjardins:
You can also remove the existing spinner from userInfo, i.e. replace [this line](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/50aa34e3d1ed5414a25ff8f630f401dffbc98d8c/lib/modules/userInfo.js#L114) with `update(username);`

erikdesjardins:
Great, thanks!

PR #3269, opened 2016-08-11
showImages: expando-button CSS clean-up

larsjohnsen:
Also fixes `.linkImg` expandos on `/comments` pages.

PR #3275, opened 2016-08-13
showImages: ensure we set `display: inline-block` on `.commentImg`

erikdesjardins:
re #3269

Otherwise we lose expando icons in comments with the native types (selftext, video):

image

And for `.comment > .parent > &.linkImg`:

image

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -42,6 +42,7 @@ img {
 	&.commentImg {
 		vertical-align: middle !important;
 		float: none;
+		display: inline-block;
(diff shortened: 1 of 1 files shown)


i-cant-git:
I use 5.5.4 Beta on Firefox 54

a1_

a2

OP comment expandos still use `display: block`

erikdesjardins:
That's not exactly what this PR is about, but okay... d84d5ffa39b6dc4d27ae71185d13d36d8674707d

i-cant-git:
Thanks Erik :wink:

PR #3277, opened 2016-08-14
Update dependences.

BenMcGarry:


erikdesjardins:
One of the issues is the sort-keys rule in ESLint, but we don't want that rule enabled anyways, so add `"sort-keys": 0,` [somewhere around here](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/6474bb526179eb3d045a26c678bf47c1df8728a2/.eslintrc.json#L182) (in alphabetical order).

The other is with power-assert (again), so we may have to downgrade `ava`. Will investigate.

PR #3284, opened 2016-08-16
Split changelog into files by version

andytuba:
and make version changelog file from `changelog/UNRELEASED.md` when running `npm version`.

⚠️ fails awkwardly if changelog/vNew.Version.Number.md already exists.


build/version.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable import/no-commonjs, import/no-nodejs-modules */
+
+const execSync = require('child_process').execSync;
+const prependFile = require('prepend-file');
+const metadata = require('../package.json');
+const unreleasedChangelog = 'changelog/UNRELEASED.md';
+const templateChangelog = 'changelog/_EXAMPLE.md';
+
+const version = metadata.version;
```
andytuba: Maybe check if newChangelog exists first?

build/version.js
```diff
@@ -0,0 +1,17 @@
...
+const newChangelog = `changelog/v${version}.md`;
+execSync(`git mv ${unreleasedChangelog} ${newChangelog}`);
+prependFile.sync(newChangelog, `## [v${version}](https://github.com/${metadata.repository.username}/${metadata.repository.repository}/releases/v${version})\n\n`);
+execSync(`git add ${newChangelog}`);
+
+execSync(`cp ${templateChangelog} ${unreleasedChangelog}`);
```
erikdesjardins: Use the equiv of `require('fs-extra').copySync` for Windows

Ajedi32:
I guess this is to make it easier to automate maintenance of the change log? That's pretty cool. Worth noting though that there are a couple minor disadvantages to this approach:

1. It makes it more difficult to Ctrl+F through the CHANGELOG to find out in what version of RES a specific change was implemented.
2. The sort order of the versions in `changelog/` won't necessarily be oldest to newest. (E.g. A theoretical 4.7.10 would be sorted _before_ 4.7.9 if you're sorting by name.)
3. It's marginally less discoverable, since you're no longer using the conventional `CHANGELOG` file in the root of the repo.

These are all definitely solvable problems though:

1. We could include the full, concatinated CHANGELOG in a page on http://redditenhancementsuite.com/
2. No real solution, but the solution for 1 above also lessens the impact of this.
3. We could keep a CHANGELOG.md file in the root that tells people to look in `changelog/`

andytuba:
> We could include the full, concatenated CHANGELOG in a page on http://redditenhancementsuite.com/

I like the sound of that idea. I set up the known issues page on the upcoming res.com in a similar fashion: http://reddit-enhancement-suite.github.io/knownissues/

> We could keep a CHANGELOG.md file in the root that tells people to look in changelog/

Good idea! And a pointer to res.com, too.

package.json
```diff
@@ -27,7 +29,8 @@
     "lint-fix": "eslint . --fix",
     "test": "cross-env NODE_ENV=test ava",
     "coverage": "cross-env NODE_ENV=test nyc ava",
-    "report-coverage": "nyc report --reporter=text-lcov | coveralls"
+    "report-coverage": "nyc report --reporter=text-lcov | coveralls",
+    "version": "babel-node build/version.js"
```
Ajedi32: Might want to use `postversion` for this, so that `build/version.js` isn't using the old version when creating the changelog file.

erikdesjardins: No, `version` is what we want to use. `preversion` has the old version number, `version` has the new version number and is before the commit (which we want to add to), and `postversion` is after the commit

build/version.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable import/no-commonjs, import/no-nodejs-modules */
+
+const execSync = require('child_process').execSync;
+const prependFile = require('prepend-file');
+const metadata = require('../package.json');
+
+const unreleasedChangelog = 'changelog/UNRELEASED.md';
+const templateChangelog = 'changelog/_EXAMPLE.md';
+
+const version = metadata.version;
```
Ajedi32: You should just be able to use `process.env.npm_package_version` here. See: https://docs.npmjs.com/misc/scripts#packagejson-vars

package.json
Ajedi32: Ah, really? Somebody needs to tell npm to update [their docs](https://docs.npmjs.com/misc/scripts#description):

> preversion, version: Run BEFORE bump the package version

:stuck_out_tongue_winking_eye:

Ajedi32: Oh. Haha. They just did: https://github.com/npm/npm/commit/c565d893a38efb6006e841450503329c9e58f100

erikdesjardins: Huh, I'm surprised that's not deployed yet (I guess only on a new npm release?).

The `npm version` [docs](https://docs.npmjs.com/cli/version) go into more detail and seem to be correct:

> 4\. Run the version script. These scripts have access to the new version in package.json (so they can incorporate it into file headers in generated files for example). Again, scripts should explicitly add generated files to the commit using git add.

build/version.js
andytuba: Rebased with `fs.openReadStream().pipe(fs.openWriteStream())`

build/version.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable import/no-commonjs, import/no-nodejs-modules */
```
Ajedi32: I guess `import/no-commonjs` can be removed now since you're using ES6 imports?

PR #3288, opened 2016-08-16
filteReddit: only add the quick toggle when the module is enabled and running

### LINKED ISSUE #3285: NSFWQuickToggle shows up even if the module is disabled

erikdesjardins:
Fixes #3285.

Previously, the toggle was added in `always()` to preserve legacy behaviour, where the toggle showed up everywhere, even on pages where filteReddit doesn't run. But this doesn't make much sense, because you're unlikely to want to toggle the NSFW filter while browsing a wiki page or something.


andytuba:
:+1:

and if anyone asks, they can learn about the less-quick toggles.

Issue #3208, opened 2016-07-27
Feature: Make the gear menu open on click, rather than on hover (+ other gear menu things)


ScoreUnder:
I'm really not a fan of hover menus at all, they feel very finicky to use. Having the hover action and the click action be different is unintuitive and often frustrating, and having the click action take me away from the page without being presented as a proper link is also frustrating. (Because there's no indication on the status bar that it's about to take me off the page, and even when I know it's there I don't have the option to open it in a new tab.) Hovering to open the gear menu is also inconsistent with the accounts menu, which requires a click to open.

Regarding use cases, *almost always* I'm clicking that menu to get to the settings dialog. I have never once clicked it wanting to read a RES announcement, and yet if an announcement is present that's what the software assumes I want.

What I'd like as part of this feature:

* The RES announcements link is a menu item inside the gear menu, and this link is never the gear icon itself.
* The RES announcements link is a real `<a href=...>` link.
* The click action for the gear menu opens it, and keeps it open regardless of mouse position until it is clicked out of or the gear is activated (by click or keyboard) again.
* The gear menu has the `role="button"` role, enabling it to be found by extensions like pentadactyl (and potentially by screen readers). Alternatively it could be a styled `<button>` which would give the same effect.
* The gear menu has a `tabindex` and responds to the return key press as it does to click.
* Each link in the menu has the `role="link"` role for the same reason.
* The z-index of the menu is reduced to something sensible. Currently there is only 1 possible greater z-index.

andytuba:
These are excellent ideas. 

By the way, the upcoming release does add the "announcement!" as a menu item. However, clicking the button still sends you to /r/RESAnnouncements.

One question: why reduce the z-index?

ScoreUnder:
I suppose I could use the z-index above that, but it feels a little awkward because at that point it's no longer possible to use any more z-indexes for any other purpose. What brought the z-index to my attention was invisible hint markers in pentadactyl. I tried to fix this by setting pentadactyl hint markers to z-index 2,000,000,000 and was surprised when they were still hidden by the menu. (Then I used the inspector to find the real z-index I should be using)

andytuba:
Oh, thanks for cluing me in to the pentadactyl conflict. The z-index was set silly high to ensure it showed above subreddit-styled features. I'm sure there's an acceptable middle ground there. Do you have any specific suggestions, like the default z-index for pentadactyl's hint markers?

[...]

PR #3290, opened 2016-08-16
move ! gear announcement badge to separate clickable icon

andytuba:
incidentally, show gear menu on click

screen shot 2016-08-17 at 4 45 39 pm

Addresses part of #3208

### DIFF
--- a/lib/modules/announcements.js
+++ b/lib/modules/announcements.js
@@ -31,22 +31,19 @@ const recheckPostAfter = HOUR;
 
 const now = Date.now();
 
-let post;
 let $biff;
 
-module.beforeLoad = async function() {
+module.go = async () => {
 	if (isCurrentSubreddit(subreddit)) {
 		setMarkedRead();
 	} else {
-		post = await postForNotification();
+		const post = await postForNotification();
+		if (post) {
+			notify(post);
+		}
 	}
 };
 
-module.go = () => {
-	if (post) {
-		notify(post);
-	}
-};
 
 async function postForNotification() {
 	const post = await getLatestPost();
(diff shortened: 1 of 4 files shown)


lib/modules/announcements.js
```diff
@@ -30,53 +31,70 @@ const recheckPostAfter = HOUR;
...
 	if (post.created_js <= await getMarkedRead()) {
 		// posted before last "marked read"
-		return;
+		return false;
 	}
 
+	return true;
+}
+
+const notify = _.once(async function(post) {
```
andytuba: This _.once is probably superfluous.

erikdesjardins: Even if it shouldn't be called more than once because it's only called in  `go()`, if you're using it to uphold an invariant then :+1:

erikdesjardins: Also `async post => {`

lib/modules/announcements.js
```diff
@@ -30,53 +31,70 @@ const recheckPostAfter = HOUR;
 
 const now = Date.now();
 
-module.go = () => {
+let post;
```
erikdesjardins: I don't really like module globals like this, I think we could call `postForNotification()` in `go()` (it doesn't need to be super fast in `beforeLoad`) and pipe everything down through arguments

lib/modules/announcements.js
```diff
@@ -30,53 +31,70 @@ const recheckPostAfter = HOUR;
...
+let $biff;
+
+module.beforeLoad = async function() {
```
erikdesjardins: `async () => {`

erikdesjardins:
Otherwise the code LGTM (not tested yet)

lib/modules/announcements.js
andytuba: Okay, I'll squash it into `go`. On the off-change it's slow and annoying, we can fix the UX with CSS in a hotfix or something.

erikdesjardins: Actually #3291 lets you do this without a global

PR #3299, opened 2016-08-18
Orangered fixups

andytuba:
mostly to ensure the native message count is up to date with the nre floating message, and unreadLinksToInbox is applied to every orangered element.

### DIFF
--- a/lib/modules/orangered.js
+++ b/lib/modules/orangered.js
@@ -171,6 +171,18 @@ const setupFaviconBadge = _.once(() => {
 	return favicon;
 });
 
+function setupNativeButtons() {
+	if (nativeInboxButton) return;
+
+	nativeInboxButton = document.querySelector('#header-bottom-right #mail');
+	nativeMailCount = document.querySelector('#header-bottom-right .message-count');
+	if (!nativeMailCount && (module.options.updateCurrentTab.value || module.options.updateOtherTabs.value)) {
+		nativeMailCount	= document.createElement('a');
+		nativeMailCount.style = 'display: inline-none;';
+		$(nativeInboxButton).after(nativeMailCount);
+	}
+}
+
 function setupFloatingButtons() {
 	if (!module.options.showFloatingEnvelope.value) return;
 	if ($floatingInboxCount) return;
--- a/lib/modules/orangered.js
+++ b/lib/modules/orangered.js
@@ -182,7 +182,7 @@ const nativeButtons = _.once(() => {
 	const nativeInboxButton = document.querySelector('#header-bottom-right #mail');
 	let nativeInboxCount = document.querySelector('#header-bottom-right .message-count');
 	if (!nativeInboxCount && (module.options.updateCurrentTab.value || module.options.updateOtherTabs.value)) {
-		nativeInboxCount	= document.createElement('a');
+		nativeInboxCount = document.createElement('a');
 		nativeInboxCount.style = 'display: none;';
 		$(nativeInboxButton).after(nativeInboxCount);
 	}
--- a/lib/modules/orangered.js
+++ b/lib/modules/orangered.js
@@ -173,9 +173,7 @@ const orangeredElements = () => ({
 });
 
 const $orangeredElements = () => {
-	const dict = orangeredElements();
-	const elements = Object.keys(dict).map(key => dict[key]);
-	return elements.reduce(($all, ele) => $all.add(ele), $());
+	return Object.values(orangeredElements()).reduce(($all, ele) => $all.add(ele), $());
 };
 
 const nativeButtons = _.once(() => {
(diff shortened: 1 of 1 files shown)


lib/modules/orangered.js
```diff
@@ -171,6 +167,18 @@ const setupFaviconBadge = _.once(() => {
 	return favicon;
 });
 
+function setupNativeButtons() {
```
erikdesjardins: Maybe replace this and two of the globals with something like

```js
const nativeButtons = _.once(() => {
    const inbox = document.querySelector('#header-bottom-right #mail');
    let mailCount = document.querySelector('#header-bottom-right .message-count');
    if (!mailCount && (module.options.updateCurrentTab.value || module.options.updateOtherTabs.value)) {
        mailCount = document.createElement('a');
        mailCount.style = 'display: inline-none;';
        $(inbox).after(mailCount);
    }
    return { inbox, mailCount };
});
```

Also `inline-none`?

andytuba: i think you are right about the inline-none

andytuba: the once is also a good idea

lib/modules/orangered.js
```diff
@@ -171,9 +166,32 @@ const setupFaviconBadge = _.once(() => {
...
+	...nativeButtons(),
+	...floatingButtons(),
+	$modmailButton: $('#modmail'),
+});
+
+const $orangeredElements = () => {
+	const dict = orangeredElements();
+	const elements = Object.keys(dict).map(key => dict[key]);
+	return elements.reduce(($all, ele) => $all.add(ele), $());
+};
```
andytuba: @erikdesjardins How's this read? Feels a little awkward, but it's really only called once.

lib/modules/orangered.js
```diff
@@ -171,9 +166,32 @@ const setupFaviconBadge = _.once(() => {
...
-	if ($floatingInboxCount) return;
+const orangeredElements = () => ({
+	...nativeButtons(),
+	...floatingButtons(),
+	$modmailButton: $('#modmail'),
+});
+
+const $orangeredElements = () => {
+	const dict = orangeredElements();
+	const elements = Object.keys(dict).map(key => dict[key]);
```
erikdesjardins: This is `Object.values(dict)`, I think :stuck_out_tongue_winking_eye:

lib/modules/orangered.js
erikdesjardins: Yeah that looks fine

lib/modules/orangered.js
```diff
@@ -171,9 +166,32 @@ const setupFaviconBadge = _.once(() => {
...
+	return elements.reduce(($all, ele) => $all.add(ele), $());
+};
+
+const nativeButtons = _.once(() => {
+	const nativeInboxButton = document.querySelector('#header-bottom-right #mail');
+	let nativeInboxCount = document.querySelector('#header-bottom-right .message-count');
+	if (!nativeInboxCount && (module.options.updateCurrentTab.value || module.options.updateOtherTabs.value)) {
+		nativeInboxCount	= document.createElement('a');
```
erikdesjardins:
```
nativeInboxCount = document
                ^ looks like this is a tab
```

lib/modules/orangered.js
```diff
@@ -171,9 +166,32 @@ const setupFaviconBadge = _.once(() => {
 	return favicon;
 });
 
-function setupFloatingButtons() {
-	if (!module.options.showFloatingEnvelope.value) return;
-	if ($floatingInboxCount) return;
+const orangeredElements = () => ({
```
erikdesjardins: You could also wrap this in `_.once` if you want so you aren't selecting `#modmail` every time

lib/modules/orangered.js
andytuba: Only in Firefox 😭

lib/modules/orangered.js
andytuba: huh.

lib/modules/orangered.js
erikdesjardins: Nah Babel transforms it https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/1ef2c155b88c2dbbdd4a36d2b97ce1947df239ba/lib/core/profiling.js#L99-L101

lib/modules/orangered.js
andytuba: I'll sidestep that: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3299/commits/42d273a278c85e573537a82646ec1e5cad5b5f30

lib/modules/orangered.js
andytuba: Neato! https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3299/commits/cacaebc0bccf66721825e2bae8d8ae02c4cc78f0

PR #3301, opened 2016-08-18
Fix show images, single click opener, etc. on NER p2+

andytuba:
by fixing the MutationObserver sanity check

### DIFF
--- a/lib/utils/dom.js
+++ b/lib/utils/dom.js
@@ -21,7 +21,8 @@ export function waitForEvent(ele, ...events) {
  * @returns {MutationObserver}
  */
 export function observe(ele, options, callback) {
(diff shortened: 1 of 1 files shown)


lib/utils/dom.js
```diff
@@ -21,7 +21,8 @@ export function waitForEvent(ele, ...events) {
  * @returns {MutationObserver}
  */
 export function observe(ele, options, callback) {
-	if (typeof MutationObserver !== 'function') {
+	if (typeof MutationObserver === 'undefined') {
```
erikdesjardins: Context: `typeof MutationObserver` is `object` (!), not `function` in Safari 9

PR #3307, opened 2016-08-19
Fixes bug with writing to style (strict compliance) causing edge to break filters.

BenMcGarry:
Fixes console error: "Assignment to read-only properties is not allowed in strict mode".


lib/modules/orangered.js
```diff
@@ -179,7 +179,7 @@ const nativeButtons = _.once(() => {
 	let nativeInboxCount = document.querySelector('#header-bottom-right .message-count');
 	if (!nativeInboxCount && (module.options.updateCurrentTab.value || module.options.updateOtherTabs.value)) {
 		nativeInboxCount = document.createElement('a');
-		nativeInboxCount.style = 'display: none;';
+		nativeInboxCount.style.cssText = 'display: none;';
```
andytuba: nativeInboxCount.style.display = 'none';

Issue #1330, opened 2014-06-27
Disable keyboardNav.LastIndex, neverEndingReddit.lastVisible.., newCommentCount in private mode


jewel-andraia:
http://www.reddit.com/r/Enhancement/comments/298wul/feature_request_stop_saving_history_while_private/

PR #3309, opened 2016-08-20
newCommentCount: remove the option to show all tracked threads

erikdesjardins:
...and only show threads that have been explicitly subscribed to.

https://www.reddit.com/r/Enhancement/comments/4ym1di/

This is effectively an implementation detail, so I'm not sure why this was visible in the first place. I suppose you could stop tracking new comments for a thread, but it's not like the `(x new)` text is very intrusive ¯\\\_(ツ)_/¯

Related to but not an implementation of #1330

### DIFF
--- a/lib/modules/newCommentCount.js
+++ b/lib/modules/newCommentCount.js
@@ -62,9 +62,6 @@ module.go = () => {
 		// add tab to dashboard
 		Dashboard.addTab('newCommentsContents', 'My Subscriptions');
 		// populate the contents of the tab
-		const $showDiv = $('<div class="show">Show </div>');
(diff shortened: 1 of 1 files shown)


andytuba:
:+1: ostensibly new comment count shouldn't store visits unless the user specifically opts into that, but this is a step towards protecting that privacy. 

And if one person does speak up looking for recently viewed posts, they can be educated how to filter their browser history.

andytuba:
*shouldn't store visits while incognito

erikdesjardins:
> ostensibly new comment count shouldn't store visits [while incognito] unless the user specifically opts into that

:+1: Yeah whenever we implement #1330 that's what it should probably do

Issue #3292, opened 2016-08-17
Move showImages.clippy into mediacontrols if available


andytuba:

thybag:
Think this one is probably close-able as of https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3316

PR #3316, opened 2016-08-23
showImages clippy: Integrate with mediaControls

larsjohnsen:
peek 2016-08-23 16-13

1st click: Show the info
2nd click: Disable it

Implements #3292

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -38,6 +38,7 @@ import {
 	regexes,
 } from '../utils';
 import { addURLToHistory, ajax, isPrivateBrowsing, openNewTab } from '../environment';
+import * as Options from '../core/options';
 import * as SettingsNavigation from './settingsNavigation';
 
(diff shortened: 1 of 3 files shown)


andytuba:
This looks small enough and totally great enough that I think we could push it to v4.7.8.

lib/modules/showImages.js
```diff
@@ -1403,21 +1415,17 @@ function keepMediaVisible(media) {
...
 	}
 
-	if (elem.tagName === 'IMG' && module.options.imageMove.value) {
+	if (module.options.imageMove.value) {
 		clippy.push('shift-drag to move');
 	}
 
-	const title = clippy.join(' or ');
-
-	elem.setAttribute('title', title);
```
andytuba: :+1: to showing clippy inside media controls, or not at all. It's always seemed more obnoxious than useful as a tooltip, at least from the feedback I've heard.

PR #3318, opened 2016-08-23
use Init.headReady instead of Safari rolling its own

### LINKED ISSUE #2858: RES leaking into Twitter.com?
I just saw "4.6.1" appear at the bottom of a Twitter oauth page (edit:) after approving slack for twitter oauth

andytuba:
Fixes #2858

### DIFF
--- a/safari/environment.js
+++ b/safari/environment.js
@@ -3,6 +3,7 @@
 import resCss from '../lib/css/res.scss';
 
 import { createMessageHandler } from '../lib/environment/_messaging';
+import * as Init from '../lib/core/init';
 import { nonNull } from '../lib/utils';
 
 // DOM Collection iteration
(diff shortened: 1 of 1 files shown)


safari/environment.js
```diff
@@ -16,12 +17,12 @@ window.onunload = () => {};
 // since safari's built in extension stylesheets are treated as user stylesheets,
 // we can't inject them that way.  That makes them "user stylesheets" which would make
 // them require !important everywhere - we don't want that, so we'll inject this way instead.
-nonNull(() => document.head, 100).then(() => {
+nonNull(() => Init.headReady, 30).then(() => Init.headReady.then(() => {
```
erikdesjardins: You can probably just use `_.defer` (same as `setTimeout(..., 0)`), since module loading is synchronous

PR #3321, opened 2016-08-24
Flow typing

erikdesjardins:
Flow finally runs on Windows; let's do this :smile:

I've typed [quickMessage.js](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3321/files#diff-159154909148c451ef693fe3720f3ffe) to demonstrate how little overhead this adds for normal development (though there may be more when the imported modules [commandLine, commentTools, etc.] are typed).


lib/utils/dom.js
```diff
@@ -52,24 +55,38 @@ export function waitForChild(ele, selector) {
 	});
 }
 
-export function click(obj, button = 0) {
-	const evt = document.createEvent('MouseEvents');
-	evt.initMouseEvent('click', true, true, window.wrappedJSObject, 0, 1, 1, 1, 1, false, false, false, false, button, null);
-	obj.dispatchEvent(evt);
+export function click(obj: EventTarget, button: number = 0): void {
+	obj.dispatchEvent(new MouseEvent('click', {
```
erikdesjardins: Tested in all browsers (+ with isProgrammaticEvent) and this behaves identically to the (_very_ deprecated) `initMouseEvent`

Ajedi32:
I'm a bit skeptical of changes like this since, unlike the ES6/7 transformations, this is adding additional syntax that isn't part of the JavaScript language. You're undoubtedly a bigger stakeholder in this though than I am, so if you and the other major contributors believe this will make development easier for you then go for it!

honestbleeps:
having been absent for so long due to being busy with work, I feel bad even chiming in because I feel less like "owner" than I once did... but I do tend to agree that adding not-actual-javascript syntax could be a concern for the following reasons:

- how long will this project exist / be supported?
- does the benefit(s) it provides outweigh the higher barrier to entry for open source contribution (people have to learn a new thing)?

I realize a lot of work has been done toward this and I wish I'd voiced my opinion earlier... I'm not saying "no, don't do this", but I do think it should be discussed before it's ultimately merged (or not merged, I suppose)...

erikdesjardins:
> how long will this project exist / be supported?

It's a [Facebook project](https://github.com/facebook/flow), used heavily in the React repo (and being expanded to other Facebook repos like graphql), and is the defacto replacement for `React.PropTypes`. If it does disappear (I'm willing to bet it won't, but let's entertain that), we can run the code through Babel with only the `flow-strip-types` transform<sup>*</sup>.

> does the benefit(s) it provides outweigh the higher barrier to entry

I think I can say "easily", because the barrier to entry is not much higher.

Let's look at [quickMessage.js](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3321/files#diff-159154909148c451ef693fe3720f3ffe): I changed ~20 lines to add null-checks, etc. But that's only vanilla js, so you don't need to learn anything else; it's effectively just replacing a review comment "you should null-check this" with automation. And in that regard Flow's errors are very easy to understand:

image

Besides that, only the two exported functions have type annotations (everything else is inferred), so your average contributor (adding a small feature or fix to a module, not adding or changing APIs) may never even see a type annotation.

If you're modifying core APIs or utils, you will have to understand Flow. But, to be blunt: if someone is unwilling to learn, I don't want them touching core APIs or utils.

To be honest, I do think the "barrier to entry" concern is a bit overblown. As long as we don't make a big deal out of using Flow, I doubt most new contributors will realize that we do. If the tools we use are too intimidating, then we should remedy this by improving our docs to emphasize that PRs don't have to be perfect, and that contributors should expect to work with us through review (in fact we should do this anyways, #3324); not by removing said tools.

And lastly: if it does turn out to be a barrier, the great thing about Flow is that it's not all-or-nothing. We could decide to type only `utils/`, `core/`, `environment/`, and leave `modules/` untyped. Of course this also misses out on a lot of the benefits, so I don't want to resort to it, but it is an option.

**TL;DR**: Flow is backed by Facebook; it's not going anywhere. Contributors don't need to know anything<sup>†</sup>, they can continue to develop in vanilla js without types (type errors don't prevent you from building), although CI may fail when they open a PR (but that's okay!). Review may take a bit longer when there are type errors, but it's justified by the increased safety/maintainability.

<sup>\*: but this will break whitespace formatting\*\*</sup>
<sup>\*\*: but ESLint can autofix (almost) all whitespace</sup>
<sup>†: a slight exaggeration</sup>

honestbleeps:
my apologies for causing alarm with my concern, I'm speaking from a place of both ignorance and being disconnected from the project a bit... you make a good case here, thank you for laying it out and getting me up to speed!

erikdesjardins:
Don't apologize; it was a reasonable concern, and I'm glad to explain :smile:

PR #3328, opened 2016-08-28
NER returnToPrev: Do not restore page on reload

larsjohnsen:
Fixes https://www.reddit.com/r/RESBetaTesting/comments/4zmkmb/announcement_res_v478_beta_released/d6z9ihi

API details: https://developer.mozilla.org/en-US/docs/Web/API/Navigation_timing_API

### DIFF
--- a/lib/modules/neverEndingReddit.js
+++ b/lib/modules/neverEndingReddit.js
@@ -167,7 +167,12 @@ function togglePause(pause, source) {
 }
 
 function initiateReturnToPrevPage() {
+	const navigation = window.performance.navigation;
+	// Only restore state if user returned here
+	const returned = ![navigation.TYPE_NAVIGATE, navigation.TYPE_RELOAD].includes(navigation.type);
+
 	if (
+		returned &&
 		history.state &&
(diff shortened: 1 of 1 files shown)


lib/modules/neverEndingReddit.js
```diff
@@ -167,7 +167,12 @@ function togglePause(pause, source) {
 }
 
 function initiateReturnToPrevPage() {
+	const navigation = window.performance.navigation;
```
erikdesjardins: Nit: use `performance.navigation` for future type checking (`window` is untyped)

lib/modules/neverEndingReddit.js
```diff
@@ -167,7 +167,12 @@ function togglePause(pause, source) {
...
+	// Only restore state if user returned here
+	const returned = ![navigation.TYPE_NAVIGATE, navigation.TYPE_RELOAD].includes(navigation.type);
```
erikdesjardins: Why not `performance.navigation.type === performance.navigation.TYPE_BACK_FORWARD` ?

Issue #2604, opened 2016-01-11
Don't try to expando dropbox links


andytuba:
https://www.reddit.com/r/RESissues/comments/40fwcv/bug_dropbox_image_viewing/

showImages has a domain blacklist, right?

Ajedi32:
Alternately, is there any way to get around those restrictions imposed by Dropbox? As a browser extension, I suspect RES has a lot more flexibility in this regard than a regular website does.

Or... maybe I should back up a step. _Should_ we try to get around these restrictions? Any idea why Dropbox is doing this in the first place?

andytuba:
We could write an official JavaScript client that calls file_get from the Dropbox API instead of hotlinking directly.. but will Dropbox API give RES an official URL that's okay for hotlinking?

Ajedi32:
I'm not actually sure whether there's a way to do this using the Dropbox API. My initial thought was to just rewrite the `referer` header in the request for the image to prevent Dropbox from detecting it as a hotlink. Not sure if there's a cross-browser way of doing that. I know you can do it with Chrome. https://developer.chrome.com/extensions/webRequest

I guess a feature like that might be more trouble than it's worth though. Just disabling expandos for those images is a reasonable solution, especially short-term.

andytuba:
I'd rather not resort to those kinds of hijinks since it may result in Dropbox going to more drastic measures to enforce their policy. It'd be better to see if Dropbox as a method they condone for hotlinking to images, or forego it.

Ajedi32:
Yeah, that's why I raised the concern in my previous comment of whether we _should_ try to get around such restrictions, and what Dropbox's motivation for putting them in place initially was.

I'd argue that what RES is doing isn't quite the same as hotlinking, since users have to click to expand the image anyway (which is essentially the same as them clicking on the link, only slightly more convenient).

That said, obviously using the official API is a much easier and better way of doing this if possible.

honestbleeps:
the purpose of dropbox having bandwidth limits is pretty clear, so I don't think we should really try and circumvent it.

official API usage is, to me, the only viable option and even then I'm not sure it'd be possible because we'd probably need some sort of key/secret that ought not be shared in our open source javascript.

Ajedi32:
[...]

PR #3330, opened 2016-08-29
Create mediaHost blacklist & add dropbox to it

thybag:
Edit: Depending on the outcome of discussions in https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2604 it may be better to go with a working mediahost & close this for now (assuming we don't have anything else we want to blacklist?)

-----

Provides an initial fix to https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2604 by blacklisting the dropbox domain from the image expando system.

As a bonus, it also means reddit's own dropbox expando's (which seem to work by using a reddit held copy of the image) will work on the index pages

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -292,6 +292,9 @@ module.exclude = [
 	/^https?:\/\/(?:[\-\w\.]+\.)?reddit\.com\/[\-\w\.\/]*\/submit\/?$/i,
 	/^https?:\/\/(?:[\-\w\.]+\.)?reddit\.com\/subreddits/i,
 ];
+module.mediaHostBlackList = [
+	"dropbox.com"
+];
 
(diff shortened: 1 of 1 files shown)


lib/modules/showImages.js
```diff
@@ -681,6 +684,13 @@ async function resolveMediaUrl(element, thing) {
 }
 
 function getMediaInfo(element, mediaUrl) {
+	// Check host is not blacklisted
+	for (const blackListedHost of module.mediaHostBlackList) {
+		if (mediaUrl.hostname.indexOf(blackListedHost) !== -1) {
+			return false;
+		}
+	}
```
Ajedi32: Probably a good idea to extract this to a method.

erikdesjardins:
Closing in favour of #3381

PR #3333, opened 2016-08-31
Update URLs for new structure

BenMcGarry:


erikdesjardins:
Might as well make all of these `https` since they'll get redirected there anyways

Ajedi32:
Oh? How does that work? I thought GitHub Pages doesn't support HTTPS for custom domains. Or are you using it with Cloudflare?

erikdesjardins:
Yep, cloudflare

PR #3347, opened 2016-09-02
Fix issues with Keyboard Navigation when `fixHideLinks` is turned off.

alexvanolst:
Resolves two issues around hiding links with a keyboard shortcut.

**Issue 1:** 
Null reference on pressing hide shortcut shortly after hiding or when no element is selected (this causes keyboard navigation to stop working). Quite an old issue, there was a post [here](https://www.reddit.com/r/RESissues/comments/40irwi/bug_issues_with_hide_shortcuthiding_submissions/?st=islzf9j7&sh=c3e95dd3) but it was deleted.

**Issue 2:**
Scrolling to the next item does not work nicely when hiding items, since after the item has been hidden the page content moves and the content that was previously scrolled to moves off screen. I've adapted larsjonsen's code to select the nearest item without scrolling after a hide.

https://www.reddit.com/r/RESissues/comments/4qc4xc/bug_hide_select_new_5_second_delay_hides_moves_up/?st=islz1ap4&sh=d7cffdd2 (only fixes case where `fixHideLinks` is off)

### DIFF
--- a/lib/modules/selectedEntry.js
+++ b/lib/modules/selectedEntry.js
@@ -144,6 +144,11 @@ export function select(thing, options) {
 	_select(thing, options);
 }
 
+export const selectClosestVisible = _.debounce(() => {
+	const target = selectedThing.getNext({ direction: 'down' }) || selectedThing.getNext({ direction: 'up' });
+	if (target) _select(target, { scrollStyle: 'none' });
+});
+
 export function unselect() {
 	const prevSelected = selectedThing;
 	_select(undefined);
--- a/lib/modules/selectedEntry.js
+++ b/lib/modules/selectedEntry.js
@@ -144,10 +144,10 @@ export function select(thing, options) {
 	_select(thing, options);
 }
 
-export const selectClosestVisible = _.debounce(() => {
+export function selectClosestVisible(options) {
 	const target = selectedThing.getNext({ direction: 'down' }) || selectedThing.getNext({ direction: 'up' });
-	if (target) _select(target, { scrollStyle: 'none' });
-});
+	if (target) _select(target, options);
+}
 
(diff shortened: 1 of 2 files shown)


larsjohnsen:
Related: I wrote `selectClosestVisible` which may be used here after `click(hide)` instead of the `scrollStyle: 'null'` workaround.

https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3042/commits/b532b9b7e45c2741726cb7c133c230e98631f596#diff-04a03be1356b2339100df6464d9f34d6R181
Feel free to copy it.

alexvanolst:
@larsjohnsen Thanks, the problem still occurs so I changed the behaviour of selectClosestVisible to not scroll.

lib/modules/selectedEntry.js
```diff
@@ -144,6 +144,11 @@ export function select(thing, options) {
 	_select(thing, options);
 }
 
+export const selectClosestVisible = _.debounce(() => {
```
erikdesjardins: `_.debounce()` with no delay is a basically a no-op, so you can remove that and `export function selectClosestVisible() {...`

We don't need a debounce for this usage, so @larsjohnsen can add it back in his PR if necessary

erikdesjardins:
Otherwise LGTM and tested

@larsjohnsen can I get a LGTM before merge?

larsjohnsen:
LGTM

lib/modules/selectedEntry.js
alexvanolst: Cool, have removed the debounce and added the _select options as a parameter to be consistent with the rest of the select functions

erikdesjardins:
Thanks all!

PR #3349, opened 2016-09-02
chrome minimum version bump

andytuba:
chrome 49 can't `for (const foo in bar)`


erikdesjardins:
LGTM

If only we could do this retroactively...

PR #3359, opened 2016-09-02
Request permissions flow

### LINKED ISSUE #3346: Make permissions explanation for expandos less confusing
I noticed some users [expressing confusion](https://www.reddit.com/r/Enhancement/comments/50qnyj/announcement_res_v500_release/d767tpt) at the new dialog for RES expando permissions:

screenshot-www reddit com 2016-09-02 10-03-39

Upon further reflection, this is definitely very confusing wording. "Expando" seems to be a word RES just kinda made up. I've honestly never heard that term outside of the context of RES, and even [Googling for it](https://www.google.com/#q=expando&rct=j) turns up no results which might indicate this feature has anything to do with inline image expansion.

This text should be reworded, especially since it seems the whole point of this dialog is to explain to users why these permissions are needed.

larsjohnsen:
This changes the permission request flow so that the popup don't appear before the user clicks on the associated expando.

image

After clicking on the expando:
image

Resolves #3346


larsjohnsen:
@erikdesjardins Should I eslint-ignore the `Unexpected path "../modules/notifications" imported in restricted zone  import/no-restricted-paths` error?

Ajedi32:
Yeah, this is way better. Not sure I like the icon though, it doesn't really feel like an expando, and I'm not sure users will be able to guess what it does before clicking it, especially if they're not already familiar with RES. A the very least, the `+` expand indicator needs to be retained. I'm no graphics artist, but maybe something like a lock icon over one of the existing media icons would look good?

As for the flow itself, I think it'd be neat if clicking the "locked" expando opened the "To create expandos, RES needs permission to access the API(s) at" prompt inline with the page, in place of the content that would otherwise be shown if permissions were already granted. That way it's clear that the sudden permissions prompt is associated with the content the user is trying to view. Thoughts?

larsjohnsen:
@Ajedi32 Something like this?
image

Ajedi32:
Yeah, that looks way nicer.

andytuba:
Seems good to show res-standard notification at same time as permission, to address branding and number of clicks mentioned in  https://www.reddit.com/r/RESissues/comments/50z4tk/please_do_not_do_this_anymore_popup_box_with_no/

andytuba:
:+1: lock treatment

lib/environment/permissions.js
```diff
@@ -9,52 +9,44 @@ function filterPerms(perms) {
...
-	(...perms) => perms.join(',')
-);
+	const urlStripRe = /((?:\w+\.)+\w+)(?=\/|$)/i;
+
+	Notifications.showNotification({
+		header: 'Permission required',
+		moduleID: 'permissions',
+		closeDelay: 20000,
+		message: `
+				<p>In order to inline expand content from ${siteName}, RES needs permission to access the API(s) at:</p>
```
Daniel15: A lot of users won't know what an "API" is - What about just saying "RES needs permission to access these sites"?

lib/css/modules/_showImages.scss
```diff
@@ -38,6 +39,22 @@ img {
 		padding: 0;
 	}
 
+	&-requires-permission {
```
erikdesjardins: Does this need to be in CSS? Ideally the lock icon would be directly in the spritesheet.

If there's nothing I'm missing, this spritesheet should work:

expandobuttons

lib/environment/permissions.js
```diff
@@ -9,52 +9,44 @@ function filterPerms(perms) {
 	return { permissions, origins };
 }
 
-const STORAGE_KEY = 'RES.cachedDeniedPerms';
+export const has = _.memoize(
+	async (...perms) => {
```
erikdesjardins: Nit: you can remove `async` and return the promise directly.

Also, I think all callsites now spread the array, so you could replace the `...perms` rest param with a single param taking an array.

lib/environment/permissions.js
```diff
@@ -9,52 +9,44 @@ function filterPerms(perms) {
...
  */
-export const request = _.memoize(
-	mutex(async (...perms) => {
-		const { permissions, origins } = filterPerms(perms);
-
-		const hasPermissions = await sendMessage('permissions', { operation: 'contains', permissions, origins });
-
-		// already granted, don't ask the user again
-		if (hasPermissions) return;
+export const request = mutex(async (siteName, ...perms) => {
```
erikdesjardins: And if you do that you'll be able to reorder these params; `siteName` could even be optional, perhaps

PR #3363, opened 2016-09-02
showImages: Destroy expando on hide

### LINKED ISSUE #3361: Expandos does not auto collapse on hide
https://www.reddit.com/r/RESissues/comments/50vmsd/bug_expanded_videos_continue_to_play_in_the/

larsjohnsen:
Fixes #3361

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -786,6 +786,9 @@ async function completeExpando(expando, thing, mediaInfo) {
 
 	Object.assign(expando, options);
 
+	const hideButton = thing.getHideElement();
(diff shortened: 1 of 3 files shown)


lib/modules/showImages.js
```diff
@@ -786,6 +786,9 @@ async function completeExpando(expando, thing, mediaInfo) {
 
 	Object.assign(expando, options);
 
+	const hideButton = thing.getHideElement();
```
erikdesjardins: Note: this adds a listener to both the hide and unhide buttons, but this is correct because the unhide button also removes the post from the page (the /u/me/hidden page, the only place it shows up)

PR #3364, opened 2016-09-02
selectedEntry: Select on mouse down instead of only left click

### LINKED ISSUE #3362: selectedEntry does not select on middle click
https://www.reddit.com/r/RESissues/comments/50tyhq/neverending_no_longer_keeps_a_hash_with_current/d777ldn

larsjohnsen:
I don't believe this will cause any issues.

Resolves #3362


andytuba:
:+1:

PR #3373, opened 2016-09-03
disable fixing "commentpage+preview -> original post"

andytuba:
It's abusing the reddit api too hard. Let's find a better solution in a few days.


erikdesjardins:
:+1:

Issue #3383, opened 2016-09-04
User tagger popup menu gets blocked by image expando


fireattack:
![image](https://cloud.githubusercontent.com/assets/4130991/18228991/20d79a10-722b-11e6-8e9f-277492e6a61c.png)

PR #3385, opened 2016-09-04
[Bugfix] Fix user tagger dialog being behind expandos

thybag:
Fix z-index so user tagger dialog shows above expando images

fixes https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/3383

### DIFF
--- a/lib/css/modules/_userTagger.scss
+++ b/lib/css/modules/_userTagger.scss
@@ -3,7 +3,7 @@
 	position: absolute;
 	width: 336px;
 	box-sizing: border-box;
-	z-index: 35;
+	z-index: $zindex-res-hover;
 
 	form > div {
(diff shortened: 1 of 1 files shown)


lib/css/modules/_userTagger.scss
```diff
@@ -3,6 +3,7 @@
 	position: absolute;
 	width: 336px;
 	box-sizing: border-box;
+	z-index: 35;
```
andytuba: There's a z-indexes.scss with lots of variables, can you use a variable from there? Maybe add one if needed.

thybag: Think all of em are more than high enough to get job done :p Have just gone with ` $zindex-res-hover`

Issue #3372, opened 2016-09-03
v5.0.0 wikipedia submissions have a expando button that doesn't expand anythin.


creesch:
![image](https://cloud.githubusercontent.com/assets/2502190/18226918/0c2eafe2-7215-11e6-8add-d102e1a3f155.png)

andytuba:
Sounds like noEmbed's wiki support is down at the moment?

erikdesjardins:
Deferring to 5.0.2; it could be coincidental that wiki support went down (or we killed it), but let's see if it comes back up soon.

thybag:
Assuming noembeds wiki support doesn't come back, Wikipedia appears to have its own api we may be able to use via an optional permission.  (That said I've not checked what their usage policies are for it)

AjaxGb:
@thybag
https://www.mediawiki.org/wiki/API:Etiquette
It seems pretty lenient; they just ask that you try not to overload the server, and that you provide a User Agent that "identifies your script or service and provides some type of means of contacting you (e.g., an e-mail address)." [Proof of concept.](https://jsfiddle.net/xyh4w2r0/6/)

thybag:
Just had a play around and think i have an example working with cross origin headers (rather than JSONP) https://jsfiddle.net/xyh4w2r0/8/

wataf1:
FYI here is the source code for how noembed handled wikipedia: https://github.com/leedo/noembed/blob/master/lib/Noembed/Provider/Wikipedia.pm

Don't know perl that well but it looks like they didn't even use the API and just scraped the page?

Nolan1971:
If the noembed tool was scraping Wikipedia pages, then it's not really a surprise that it's down. You're really not supposed to scrape Wikipedia, especially at high volume, as that loads the server down (at least, that's my understanding). noembed may be blocked, at this point.

The MediaWiki API is relatively easy to use and is quite useful.

PR #3392, opened 2016-09-04
Wikipedia Expando v2

thybag:
Resolves https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/3372

clipboard02

The new version uses Wikipedia's own API to generate the expando & now includes the mediabox + some extra styles for it. 

Little unsure how how well we meet https://www.mediawiki.org/wiki/API:Etiquette since RES is a browser extension & thus is going to create a very distributed form of load. I've identified the API user ( as per https://www.mediawiki.org/wiki/API:Main_page#Identifying_your_client ) as `Reddit-Enhancement-Suite/5.0 ( https://github.com/honestbleeps/Reddit-Enhancement-Suite/ )`

Main changes:

* Use wikipedia's own API
* All text expando's include a `res-expando-host-{{moduleId}}` class, so they can be targeted separately
* Some extra styles to improve the display (tested with nightmode)
* Support for wikipedia.com domains
* <del>I was unable to figure out how to replicate getting section content via the #this_section hashes, so now the expando will always just show the top section of the article. </del>

Hope it all looks okay :)


BenMcGarry:
Is there anyway we can contact mediawiki about this beforehand? Just to check there's not something we are missing and to get their blessing.

thybag:
@BenMcGarry Possibly an idea, although I've got now idea how we'd go about that. I guess potentally someone could try emailing them https://en.wikipedia.org/wiki/Wikipedia:Contact_us

lib/css/modules/_showImages.scss
```diff
@@ -505,3 +505,38 @@ body:not(.res-showImages-displayImageCaptions) {
...
+		&:before, &:after {
+			content: "";
+			display: table;
+		}
+		&:after {
+			clear: both;
+		}
+	}
+}
+.res-nightmode .res-expando-host-wikipedia .md table {
```
erikdesjardins: This can go directly in with the table styles (line 521) with `.res-nightmode & { /* css */ }`

lib/modules/hosts/wikipedia.js
```diff
@@ -1,22 +1,38 @@
...
 		const data = await ajax({
-			url: 'https://noembed.com/embed',
-			data: { url: `http://${country}.wikipedia.org/wiki/${article}` },
+			url: `https://${country}.wikipedia.org/w/api.php?action=parse&format=json&prop=text|displaytitle&section=0&page=${article}&origin=*`,
 			type: 'json',
+			headers: {
+				// See https://www.mediawiki.org/wiki/API:Main_page#Identifying_your_client
+				'Api-User-Agent': 'Reddit-Enhancement-Suite/5.0 ( https://github.com/honestbleeps/Reddit-Enhancement-Suite/ )',
+				'Origin': '*',
+				'Content-Type': 'application/json; charset=utf-8',
```
erikdesjardins: This probably isn't necessary because we aren't sending any json

lib/modules/hosts/wikipedia.js
```diff
@@ -1,22 +1,38 @@
...
+
 		const data = await ajax({
-			url: 'https://noembed.com/embed',
-			data: { url: `http://${country}.wikipedia.org/wiki/${article}` },
+			url: `https://${country}.wikipedia.org/w/api.php?action=parse&format=json&prop=text|displaytitle&section=0&page=${article}&origin=*`,
 			type: 'json',
+			headers: {
+				// See https://www.mediawiki.org/wiki/API:Main_page#Identifying_your_client
+				'Api-User-Agent': 'Reddit-Enhancement-Suite/5.0 ( https://github.com/honestbleeps/Reddit-Enhancement-Suite/ )',
+				'Origin': '*',
```
erikdesjardins: The origin header can't be changed:
image

lib/templates/text.mustache
```diff
@@ -1,4 +1,4 @@
-<div class="res-text usertext-body">
+<div class="res-text usertext-body res-expando-host-{{moduleId}}">
```
larsjohnsen: I think this better be placed in the expando-box element, perhaps by setting attribute `host`.

thybag: Good thought + makes it a little more generic so it can work for all mediahosts regardless of type 👍

thybag:
Hmm, guess needing two api calls (when a #bla) is detected ain't the end of the world. Will try and find some time to give it a go

Ajedi32:
Another option would be to request the full article content and section data with one API call, then use the section data (`byteoffset`) to decide what part of the content to display.

lib/modules/hosts/wikipedia.js
```diff
@@ -7,16 +7,31 @@ export default {
 	attribution: false,
 	detect: ({ href }) => (/^https?:\/\/([A-z]{2}|simple)?\.(?:m\.)?wikipedia\.org\/wiki\/(.+)/i).exec(href),
 	async handleLink(href, [, country, article]) {
```
jdforrester: BTW, this isn't a country code but a language code, but it doesn't matter much. :-)

jdforrester:
Hey there, RES user who's also a Wikimedia staffer, provisionally this looks fine. We don't mind distributed read API load like this — it's just natural reader traffic, after all. Using "Reddit-Enhancement-Suite/5.0" in the UA string is great, thanks for doing that.

andytuba:
@jdforrester Thanks for dropping in! To clarify, [double API requests for section](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3392#issuecomment-245422030) are okay too?

jdforrester:
Well, since you asked, if you want to be fancy, using [the REST API](https://en.wikipedia.org/api/rest_v1/) instead of [the actions API](https://en.wikipedia.org/w/api.php) would probably be faster for your users and lower load for us.

For example, https://en.wikipedia.org/api/rest_v1/page/html/Foobar response is the cached "desktop" HTML for the whole page. That's an "unstable" API, but we're not changing it without a bunch of warnings and deprecation time of at least a few weeks.

That doesn't solve the section issue though. Although the API takes a "sections" parameter, it currently means the top-level DOM nodes in the `<body>` — so if the user passed in https://en.wikipedia.org/wiki/Foobar#Example_use_in_code you'd have to pull the HTML from the first query, then work out where Example_use_in_code appears (it's `mwPA` as of writing, but that's not stable between revisions), work out that that section consists (as of writing) of the ids `mwPA`, `mwPQ`, and `mwPw` (and have code to know that the next blob, `mwQA`, isn't part of that section because it's an `<Hx>` where x is equal to or higher than the starting heading, and then either do the work yourself to trim the document or re-request from us, which'd be https://en.wikipedia.org/api/rest_v1/page/html/Foobar?sections=mwPA%2CmwPQ%2CmwPw

That's a bit meh, sorry. :-( We're still working on this area, trying to make it nicer for us as well as third parties to use, especially around sections.

If you want content that's been pre-processed to be more friendly for mobile devices (*e.g.* block templates like infoboxes and some inline media files like pronunciations are removed), you could use https://en.wikipedia.org/api/rest_v1/page/mobile-sections/Foobar to get the HTML split out by section. This has the advantage of letting you just show the section you want (by `sections.[].anchor`), but might not be quite the content you want.

There's also https://en.wikipedia.org/api/rest_v1/page/mobile-sections-lead/Foobar to get the HTML for just the lead section and the IDs for all the other sections, for (I presume) the main use case of a no-anchor entry, which would reduce bandwidth usage.

Hope this helps! Sorry it's a bit of a mess right now.

lib/modules/hosts/wikipedia.js
```diff
@@ -1,22 +1,33 @@
...
+			url: `https://${country}.wikipedia.org/w/api.php?action=parse&format=json&prop=text|displaytitle&section=0&page=${article}&origin=*`,
 			type: 'json',
+			headers: {
+				// See https://www.mediawiki.org/wiki/API:Main_page#Identifying_your_client
+				'Api-User-Agent': 'Reddit-Enhancement-Suite/5.0 ( https://github.com/honestbleeps/Reddit-Enhancement-Suite/ )',
+			},
 		});
 
+		// Clean up returned data
+		const $wikiData = $(`<div>${data.parse.text['*']}</div>`);
```
erikdesjardins: Maybe use `$('<div>', { html: data.parse.text['*'] });` so it degrades a bit more gracefully if `data.parse.text['*']` is null or undefined or invalid html

thybag:
Demo with #section loading https://www.reddit.com/r/todayilearned/comments/525rkb/til_when_rodney_dangerfield_was_asked_how_long/

lib/modules/hosts/wikipedia.js
```diff
@@ -6,10 +6,40 @@ export default {
 	name: 'wikipedia',
 	domains: ['wikipedia.org', 'wikipedia.com'],
 	logo: 'https://en.wikipedia.org/static/favicon/wikipedia.ico',
-	detect: ({ href }) => (/^https?:\/\/([A-z]{2}|simple)?\.(?:m\.)?wikipedia\.(?:com|org)\/wiki\/([^#\?]+)/i).exec(href),
-	async handleLink(href, [, country, article]) {
+	detect: url => {
+		return {
+			article: url.pathname.substr(6), // remove "/wiki/"
```
andytuba: Is it worth double-checking this is an article page before asking the wikipedia api?

lib/modules/hosts/wikipedia.js
```diff
@@ -6,10 +6,40 @@ export default {
...
+		let sectionId = 0;
+
+		// If we have a hash, try and look up its section number
+		if (hash !== '') {
+			const sections = await ajax({
+				url: `https://${language}.wikipedia.org/w/api.php?action=parse&format=json&prop=sections&page=${article}&origin=*`,
+				type: 'json',
+				headers: {
+					// See https://www.mediawiki.org/wiki/API:Main_page#Identifying_your_client
+					'Api-User-Agent': 'Reddit-Enhancement-Suite/5.0 ( https://github.com/honestbleeps/Reddit-Enhancement-Suite/ )',
```
andytuba: How about
import * as Metadata from '../core/metadata';

`Reddit-Enhancement-Suite/${Metadata.version} ( ${Metadata.updatedURL} )`

or even better, add to metadata.js: `export const homepageURL = packageInfo.homepage;` and for the user-agent here, `Reddit-Enhancement-Suite/${Metadata.version} ( ${Metadata.homepageURL} )`

thybag:
I \*think\* this may now be done :p

PR #3394, opened 2016-09-05
dashboard: fix tabs

larsjohnsen:
- Adds  `filter visited links` and `show images` buttons back to the dashboard.
- Limits running the dashboard module to when visiting `/r/dashboard`
- It also changes of the tab menu buttons a little, as they were too structurally different from the native ones to always work as intended.
 - Clicking on the padding didn't trigger a `change` event 

Side effects:
- In the gear menu, `my dashboard` is below `RES console settings` (this is IMO preferable)
- In the dashboard, the tab items are a little reordered and has different styles (as the style is defined in the subreddit stylesheet).

### DIFF
--- a/lib/css/modules/_dashboard.scss
+++ b/lib/css/modules/_dashboard.scss
@@ -374,3 +374,21 @@ ul.widgetStateButtons {
 #userTaggerTable tr:hover .deleteIcon {
 	color: #369;
 }
+
+.res-r-dashboard {
+	.tabmenu li:not(.res-tabmenu-button),
+	.pagename {
+		display: none;
+	}
+
+	.tabmenu.res-dashboard-tabmenu-ready { display: inline-block; }
+
+	.res-dashboard-tab {
+		font-variant: small-caps;
+		font-size: 1.2em;
+		vertical-align: bottom;
+
+		// We want the tabs to act as a radio, i.e. don't allow deselect the currently selected
+		&.selected { pointer-events: none; }
+	}
+}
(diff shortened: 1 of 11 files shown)


lib/css/modules/_dashboard.scss
```diff
@@ -374,3 +374,21 @@ ul.widgetStateButtons {
 #userTaggerTable tr:hover .deleteIcon {
 	color: #369;
 }
+
+.res-r-dashboard {
```
erikdesjardins: This class isn't guaranteed to be present since the stylesheet module can be turned off; we should maybe put it in /r/Dashboard CSS

erikdesjardins: Actually, instead let's add `module.bodyClass = true;` in dashboard (if it's not already) and use `.res-dashboard`

larsjohnsen: That works. Updated.

lib/utils/createElement.js
```diff
@@ -83,28 +84,30 @@ export function table(items, callback) {
...
-	const checkbox = element.querySelector('input');
-	checkbox.addEventListener('change', e => {
-		e.stopPropagation();
-		element.classList.toggle('selected', checkbox.checked);
+	let checked = options.checked;
+	const a = element.querySelector('a');
+	a.addEventListener('click', () => {
+		checked = !checked;
+		element.classList.toggle('selected', checked);
+		a.dispatchEvent(new CustomEvent('change', { detail: checked }));
```
erikdesjardins: Note: confirmed that this works in Firefox, which is very restrictive about what you can use in `e.detail`

PR #3400, opened 2016-09-05
selectedEntry: Selecting on mousedown may disrupt click handlers; use mouseup instead

larsjohnsen:
Fixes https://www.reddit.com/r/RESissues/comments/51aphm/bug_problematic_interaction_of_comment_expandos

Selecting something sometimes causes elements to move around, so that the user-click won't register on the element as intended. 

Regression caused by #3364


fireattack:
Does this also fix the shift problem caused by "Selected Entry" module? If not I can open another ticket for that. 

zz8mmk

larsjohnsen:
Yes

PR #3409, opened 2016-09-08
Completely hide the video advanced bar when showVideoControls is disabled

larsjohnsen:
Also tweak `sourceErrorFallback` a little so that the error reason is available.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1675,7 +1675,7 @@ function videoAdvanced(options) {
 		return source;
 	}))).appendTo(vid).get();
 	if (!sourceElements.length) {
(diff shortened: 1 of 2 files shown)


lib/modules/showImages.js
```diff
@@ -1675,7 +1675,7 @@ function videoAdvanced(options) {
 		return source;
 	}))).appendTo(vid).get();
 	if (!sourceElements.length) {
-		_.defer(sourceErrorFallback);
+		_.defer(() => sourceErrorFallback(new Error('No playable sources was found')));
```
erikdesjardins: s/was/were/

andytuba: `s/(was|were)\s*//`

PR #3410, opened 2016-09-08
Move mediaBrowse logic to showImages

larsjohnsen:
By removing the mediaBrowse logic from the `move` function, it can easier by invoked by other select functions.

Fixes https://www.reddit.com/r/RESissues/comments/51nu5e/bug_images_not_autoexpanding_next_post_after/?ref=search_posts

### DIFF
--- a/lib/modules/selectedEntry.js
+++ b/lib/modules/selectedEntry.js
@@ -117,7 +117,14 @@ module.go = () => {
 		addFocusBorder();
 	}
 
-	setTimeout(selectInitial, 100);
+	addListener((selected, unselected) => unselected && Hover.infocard('showParent').close(false));
+	addListener(updateActiveElement);
+	addListener(updateLastSelectedCache);
+};
+
+module.afterLoad = () => {
+	// Do not re-select if user has already selected something
+	if (!selectedThing) selectInitial();
 };
 
 const onClick = _.throttle(e => {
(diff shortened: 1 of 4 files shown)


lib/modules/selectedEntry.js
```diff
@@ -202,19 +211,6 @@ function onNewComments(entry) {
 	}
 }
 
-function scrollTo(selected, last, options) {
-	if (!selected) return;
-	scrollToElement(selected.entry, options);
-}
-
-// this is run immediately for a reason, but I don't know/remember why
```
larsjohnsen: I couldn't find the reason either, so I moved this to 'module.go'.

lib/modules/showImages.js
```diff
@@ -583,22 +586,24 @@ export function toggleThingExpandos(thing, scrollOnExpando) {
...
 
 	if (mediaBrowseModeActive) {
 		if (oldExpando) oldExpando.collapse();
-		if (newExpando) newExpando.expand();
-	}
 
-	return newExpando && mediaBrowseModeActive;
+		if (newExpando) {
+			newExpando.expand();
+			options.didAutoExpand = true;
```
erikdesjardins: I think I'd prefer moving this entire function to keyboardNav (but still separated from `move`), so you can replace this with `if (module.options.scrollOnExpando.value) options.scrollStyle = 'top';` instead of setting `didAutoExpand` here and reading it in a getter in another file, which is a bit too hacky IMO.

larsjohnsen: Does https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3410/commits/ea08234b43bc55a26420520810af898b21244e08 look OK to you?

erikdesjardins: That's fine too, yeah: less dynamic and the dataflow is one way

erikdesjardins:
Great, thanks!

PR #3412, opened 2016-09-08
showImages video: conserveMemory & defer play until visible

larsjohnsen:
- Unload video when outside of `bufferScreens`
- Start playing video when at least `visibleThreshold` (%) of it is visible
  - It will only check while it is within `bufferScreens`
- `maxSimultaneousPlaying` can be set to only have to have a given number playing at any time.
 - Prioritized by: 1. Thing is selected. 2. Visibility.  3. Nearest top.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -47,6 +47,7 @@ import {
 	isPageType,
 	waitForEvent,
 	watchForElement,
+	getPercentageVisibleYAxis,
 } from '../utils';
 import { addURLToHistory, ajax, isPrivateBrowsing, openNewTab, Permissions } from '../environment';
 import * as Options from '../core/options';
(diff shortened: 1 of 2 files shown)


lib/modules/showImages.js
```diff
@@ -1773,27 +1784,67 @@ function videoAdvanced(options) {
...
+		const visibleThreshold = 0.33;
+		const isWithinPlayThreshold = () => getPercentageVisibleYAxis(vid) > visibleThreshold;
+
+		const tryPlay = () => {
+			if (!autoplay || (isWithinPlayThreshold() !== vid.paused)) return;
+			else if (vid.paused) vid.play();
+			else vid.pause();
+		};
+
+		playWhenVisibleInterval = setInterval(tryPlay, 150);
```
erikdesjardins: I think `_.throttle(tryPlay, 150, { leading: false })` in a scroll listener would be preferred over `setInterval` so it's not checking unnecessarily, unless this was chosen for a reason (I haven't read through everything yet)

lib/modules/showImages.js
```diff
@@ -1773,27 +1784,67 @@ function videoAdvanced(options) {
 		}
 	});
 
-	element.collapse = () => {
+	let playWhenVisibleInterval;
+
+	function playWhenVisible() {
+		const visibleThreshold = 0.33;
```
erikdesjardins: I think we should play videos that have any percentage on the screen -- doing so would cost a marginal amount of performance (at most two videos, one at the top and bottom, assuming no overlap) but would marginally improve responsiveness, and I think that wins here.

larsjohnsen:
@erikdesjardins Updated with solution which only uses 1 global interval.

larsjohnsen:
(Github automatically closed this for some reason)

erikdesjardins:
Awesome, thanks!

PR #3418, opened 2016-09-09
host : added osu.ppy.sh/ss/ expando

Dexesttp:
Allows link with the osu.ppy.sh/ss/<id> format to be displayed with expandos.

Tested with https://www.reddit.com/domain/ppy.sh.

See [this /r/Enhancement post](https://np.reddit.com/r/Enhancement/comments/51zk40/feature_request_osuppyshss_screenshots_expandos/) for the feature request.

### DIFF
--- a/lib/modules/hosts/ppy.js
+++ b/lib/modules/hosts/ppy.js
@@ -1,0 +1,13 @@
+export default {
+	moduleID: 'ppy.sh',
+	name: 'ppy.sh',
+	domains: ['ppy.sh'],
+	logo: 'https://s.ppy.sh/favicon.ico',
+	detect: ({ href }) => (/^https?:\/\/osu.ppy.sh\/ss\/(\d+)/i).exec(href),
+	async handleLink(href, [, code]) {
+		return {
+			type: 'IMAGE',
+			src: `http://osu.ppy.sh/ss/${code}`,
+		};
+	},
+};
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/ppy.js
```diff
@@ -0,0 +1,13 @@
+export default {
+	moduleID: 'ppy.sh',
+	name: 'ppy.sh',
+	domains: ['ppy.sh'],
```
erikdesjardins: Can be `osu.ppy.sh` since we only care about that subdomain

lib/modules/hosts/ppy.js
```diff
@@ -0,0 +1,13 @@
...
+	logo: 'https://s.ppy.sh/favicon.ico',
+	detect: ({ href }) => (/^https?:\/\/osu.ppy.sh\/ss\/(\d+)/i).exec(href),
```
erikdesjardins: Can be simplified by looking only at the `pathname`:

```js
detect: ({ pathname }) => (/^\/ss\/(\d+)/i).exec(pathname),
```

lib/modules/hosts/ppy.js
```diff
@@ -0,0 +1,13 @@
...
+	async handleLink(href, [, code]) {
```
erikdesjardins: `async` can be removed here because nothing is `await`ed

Dexesttp:
Thanks for the quick feedback !

erikdesjardins:
LGTM; should be merged as soon as we start working on 5.1.0 (right now we're only merging bugfixes for 5.0.x)

PR #3420, opened 2016-09-11
Blank safety

honestbleeps:
for safety, let's be careful to add  rel="noopener noreferer" when we use target="_blank"


erikdesjardins:
LGTM

PR #3425, opened 2016-09-11
fall back to current subreddit if thing doesn't contain subreddit link

### LINKED ISSUE #3406: Thing .getSubreddit should fall back to current subreddit
https://www.reddit.com/r/Enhancement/comments/51i34z/is_there_any_possibility_to_filter_frontpage_by/d7c40bt

andytuba:
Fixes #3406

### DIFF
--- a/lib/utils/Thing.js
+++ b/lib/utils/Thing.js
@@ -183,7 +183,11 @@ export default class Thing {
 
 	getSubreddit() {
 		const element = this.getSubredditLink();
-		return element && regexes.subreddit.exec(element.pathname)[1];
+		if (element) {
+			return regexes.subreddit.exec(element.pathname)[1];
+		} else {
+			return currentSubreddit();
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


lib/utils/Thing.js
```diff
@@ -183,7 +183,11 @@ export default class Thing {
 
 	getSubreddit() {
 		const element = this.getSubredditLink();
-		return element && regexes.subreddit.exec(element.pathname)[1];
+		if (element) {
```
erikdesjardins: Could be a ternary but I don't really care in this case

erikdesjardins:
LGTM

erikdesjardins:
Ah we should have also removed any existing fallbacks like https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/012cefaa89778e06ceaba1c36b7dc227afe0eba9/lib/modules/commentTools.js#L747

PR #3429, opened 2016-09-11
If element is not inside a thing, fallback to first visible thing

### LINKED ISSUE #3426: Thing-based macro placeholders eg subreddit broken in reply to post
https://www.reddit.com/r/RESissues/comments/520jsx/bug_macros_with_placeholder_stuff_dont_work/

For adding a top-level comment, Thing(textarea) isn't finding the post.

andytuba:
Fixes #3426: if the reference element (e.g. "add a comment" text area on top-level comment) isn't inside a thing, use the first visible thing and assume it's the main thing, e.g. on a comments page, the original post.  May need refining later.

also some incidental fixups

### DIFF
--- a/lib/utils/Thing.js
+++ b/lib/utils/Thing.js
@@ -32,12 +32,17 @@ export default class Thing {
 		return Thing.visibleThingElements().map(ele => new Thing(ele));
 	}
 
+	static elementFor(element) {
+		return $(element).closest(Thing.thingSelector).get(0) || Thing.visibleThingElements()[0];
+	}
+
 	constructor(element) {
 		if (element instanceof Thing) return element;
 		if (things.has(element)) return things.get(element);
 
-		this.$thing = $(element).closest(Thing.thingSelector);
-		this.element = this.thing = this.$thing[0];
+		const thingElement = Thing.elementFor(element);
+		this.$thing = $(thingElement);
+		this.element = this.thing = thingElement;
 
 		if (!this.element) return this;
 
--- a/lib/utils/Thing.js
+++ b/lib/utils/Thing.js
@@ -134,7 +134,7 @@ export default class Thing {
 	}
 
 	isComment() {
-		return this.thing.classList.contains('comment');
+		return this.thing.classList.contains('comment') || this.thing.classList.contains('was-comment');
 	}
 
 	getTitle() {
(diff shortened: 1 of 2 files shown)


andytuba:
wait, this did not work

andytuba:
wait, this does work, but testing in inbox uncovered a related issue.

lib/utils/Thing.js
```diff
@@ -129,7 +134,7 @@ export default class Thing {
 	}
 
 	isComment() {
-		return this.thing.classList.contains('comment');
+		return this.thing.classList.contains('comment') || this.thing.classList.contains('was-comment');
```
andytuba: `was-comment` is a comment in your inbox

lib/utils/Thing.js
```diff
@@ -32,12 +32,17 @@ export default class Thing {
 		return Thing.visibleThingElements().map(ele => new Thing(ele));
 	}
 
+	static elementFor(element) {
+		return $(element).closest(Thing.thingSelector).get(0) || Thing.visibleThingElements()[0];
```
erikdesjardins: Acknowledging that this is kinda hacky, but I don't think it'll cause any problems since almost everything we interact with is within a `.thing` already

lib/utils/Thing.js
erikdesjardins: Looks like comments in your inbox previously would return false for both `isComment` and `isLink`, so this shouldn't break anything.

In the future maybe we should use the attr `data-type = link|comment|message`

erikdesjardins:
```
/home/travis/build/honestbleeps/Reddit-Enhancement-Suite/lib/modules/commentTools.js
  719:2  error  Mixed spaces and tabs  no-mixed-spaces-and-tabs
  722:2  error  Mixed spaces and tabs  no-mixed-spaces-and-tabs
```

Otherwise LGTM

lib/utils/Thing.js
andytuba: Deffo hacky. Hopefully this won't generate too many more bugs.

lib/utils/Thing.js
andytuba: and potentially `|comment-message` too, since PMs and comment replies have different markup in the inbox.

PR #3434, opened 2016-09-12
Don't track visits incognito

### LINKED ISSUE #3405: Don't automatically save data in private mode
https://www.reddit.com/r/Enhancement/comments/51b28e/res_breaching_private_browsing_mode/d7bk9xh

andytuba:
Fixes #3405 

 - [x] Subreddit Manager optionally won't track visits to subreddits normally (do track by default) or incognito (don't track by default)
 - [x] New Comment Count optionally won't track visits to comments pages (do track by default) incognito (don't track by default) ... and maybe incognito shouldn't show "new comment" notifications at all?

### DIFF
--- a/lib/modules/newCommentCount.js
+++ b/lib/modules/newCommentCount.js
@@ -14,7 +14,7 @@ import {
 	regexes,
 	watchForElement,
 } from '../utils';
-import { Storage } from '../environment';
+import { Storage, isPrivateBrowsing } from '../environment';
 import * as Dashboard from './dashboard';
 import * as Notifications from './notifications';
 
@@ -39,6 +39,19 @@ module.options = {
 		type: 'boolean',
 		value: true,
 		description: 'Show the Subscribe button?',
+	},
+	monitorPostsVisited: {
+		type: 'boolean',
+		value: true,
+		description: 'Monitor the number of comments on posts you have visited.',
+		advanced: true,
+	},
+	monitorPostsVisitedIncognito: {
+		dependsOn: 'monitorPostsVisited',
+		type: 'boolean',
+		value: false,
+		description: 'Monitor the number of comments on posts you have visited while browsing in incognito/private mode.',
+		advanced: true,
 	},
 };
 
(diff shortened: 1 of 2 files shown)


lib/modules/subredditManager.js
```diff
@@ -115,6 +115,19 @@ module.options = {
 		value: true,
 		description: 'Show last update information on the front page (work only if you have at least 50/100 subscription, see <a href="/r/help/wiki/faq#wiki_some_of_my_subreddits_keep_disappearing.__why.3F">here</a> for more info).',
 	},
+	trackVisitedSubreddit: {
```
erikdesjardins: Bonus points: don't draw the "last visited" column when this is disabled

lib/modules/subredditManager.js
```diff
@@ -115,6 +115,19 @@ module.options = {
...
+		type: 'boolean',
+		value: true,
+		description: 'Mark subreddits visited when you visit a post listing, comments page, wiki, etc.',
```
erikdesjardins: This might be interpreted as adding to history, maybe "Store the last visited time..."

lib/modules/subredditManager.js
andytuba: I'll save that for the bonus round ... some other night :P

lib/modules/newCommentCount.js
```diff
@@ -40,6 +40,19 @@ module.options = {
...
+	monitorPostsVisited: {
+		type: 'boolean',
+		value: true,
+		description: 'Monitor the number of comments on posts you have visited.',
+		advanced: true,
+	},
+	monitorPostsVisitedIncognito: {
+		dependsOn: 'monitorPostsVisited',
+		type: 'boolean',
+		value: false,
```
erikdesjardins: Let's make this `process.env.BUILD_TARGET !== 'firefox'` or more conservatively `process.env.BUILD_TARGET === 'chrome'` since Chrome users will have specifically enabled the extension in incognito mode and seen the persistent warning:
image

erikdesjardins:
Otherwise LGTM code-wise, but if anyone thinks that "private" browsing mode is a replacement for legitimate privacy, they're gonna have a bad time (hint: Windows users try running `ipconfig /displaydns`)

lib/modules/newCommentCount.js
andytuba: I kinda prefer keeping this consistent across platforms so as to not catch multi-platformers by surprise, even with Chrome's disclaimer.

erikdesjardins: I guess I don't feel that strongly about this

Ajedi32: In my mind that warning isn't an excuse for extensions to misbehave (by saving browsing history when users have implicitly requested not to have that data saved), it's just a warning to users that some extensions *might* and Chrome has no way of preventing it.

erikdesjardins: The thing is: most users don't care, they just expect it to work when they enable it. But like I said, I don't really care either, and these aren't really landmark features

PR #3436, opened 2016-09-13
Re-enable redditmedia/i.redd.it expandos

andytuba:
but with less abusing the reddit API this time

also applicable for other posts that link to comments page + preview (as seen in some experiments)


erikdesjardins:
LGTM, although it looks like the fix hasn't been deployed on reddit's side yet (or hasn't hit all the servers), so I can't test

PR #3441, opened 2016-09-13
Keyboard shortcut works even when singleClick [L=C] is disabled

### LINKED ISSUE #3350: L+C keyboard shortcut doesn't work when L=C is hidden

erikdesjardins:
Fixes #3350 

(untested)

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -16,7 +16,6 @@ import {
 	isPageType,
 	loggedInUser,
 	matchesPageLocation,
-	mousedown,
 	niceKeyCode,
 } from '../utils';
 import { openNewTab } from '../environment';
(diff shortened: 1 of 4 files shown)


lib/utils/dom.js
```diff
@@ -63,12 +63,6 @@ click.isProgrammaticEvent = e => {
 	return e.clientX === 1 && e.clientY === 1;
 };
 
-export function mousedown(obj, button = 0) {
```
andytuba: Is this still used for clicking native reddit elements, like keyboard nav invoking upvote/downvote?

lib/modules/keyboardNav.js
```diff
@@ -30,6 +29,7 @@ import * as SaveComments from './saveComments';
 import * as SelectedEntry from './selectedEntry';
 import * as ShowImages from './showImages';
 import * as ShowParent from './showParent';
+import * as SingleClick from './singleClick';
```
andytuba: :+1:

lib/utils/dom.js
erikdesjardins: 😜 

image

s/a use/any future use/ to clarify

andytuba: Oh, they use `click` instead of `mousedown`.

Issue #3320, opened 2016-08-23
Target announcements to specific browsers


andytuba:
@BenMcGarry has an idea: if a post to /r/RESAnnouncements mentions any browser names in the title, eg `[Safari]`, then only show the bell if that browser is in the title. Implement using build targets.

michaelskiles:
I gave this a whirl:

https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3442

Note that this only applies to browsers that are currently supported by RES>

PR #3442, opened 2016-09-13
Mute announcement alerts if they are not relevant.

michaelskiles:
This change ensures the Alert icon is only active if an annoucement is not browser specific, or if an  annoucement pertains to the user's current browser.

### DIFF
--- a/lib/modules/announcements.js
+++ b/lib/modules/announcements.js
@@ -197,8 +197,7 @@ function isPostRelevant(post) {
 		return true;
 	}
 
-
-	if (postTitle.search(`${BrowserDetect.browser}`) > -1) {
+	if (postTitle.search(BrowserDetect.browser) > -1) {
 		// If post mentions user's browser, post is relevant
 		return true;
 	}
--- a/lib/modules/announcements.js
+++ b/lib/modules/announcements.js
@@ -73,7 +73,7 @@ async function shouldNotify(post) {
 		return false;
 	}
 
-	if (!isPostRelevent(post)) {
+	if (!isPostRelevant(post)) {
 		return false;
 	}
 
@@ -191,16 +191,16 @@ function setMarkedRead() {
 	}
 }
 
-function isPostRelevent(post) {
+function isPostRelevant(post) {
 	const postTitle = post.title;
 
 	if (postTitle.search(/Chrome|Safari|Firefox|Opera|Edge/) === -1) {
-		// If post is not browser specific, post is relevent
+		// If post is not browser specific, post is relevant
 		return true;
 	}
 
 	if (postTitle.search(`${currentBrowser}`) > -1) {
-		// If post mentions user's browser, post is relevent
+		// If post mentions user's browser, post is relevant
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


andytuba:
Closes #3320

lib/modules/announcements.js
```diff
@@ -31,6 +31,8 @@ const recheckPostAfter = HOUR;
 
 const now = Date.now();
 
+const currentBrowser = BrowserDetect.browser;
```
andytuba: Could just check `${BrowserDetect.browser}` below instead of `${currentBrowser}`, doesn't seem necessary to alias it for a single use..

lib/modules/announcements.js
```diff
@@ -2,7 +2,7 @@
 import _ from 'lodash';
 import { $ } from '../vendor';
 import * as Metadata from '../core/metadata';
-import { DAY, HOUR, isCurrentSubreddit, CreateElement } from '../utils';
+import { DAY, HOUR, isCurrentSubreddit, CreateElement, BrowserDetect } from '../utils';
```
andytuba: Iinstead of BrowserDetect.browser, you could probably use process.env.BUILD_TARGET

lib/modules/announcements.js
```diff
@@ -184,3 +190,21 @@ function setMarkedRead() {
...
+
+function isPostRelevant(post) {
+	const postTitle = post.title;
+
+	if (postTitle.search(/Chrome|Safari|Firefox|Opera|Edge/) === -1) {
+		// If post is not browser specific, post is relevant
+		return true;
+	}
+
+	if (postTitle.search(`${currentBrowser}`) > -1) {
```
andytuba: How about simply putting in a variable? No need for formatting..

lib/modules/announcements.js
michaelskiles: Thanks for the feedback. This was my initial thought as well, but my debugger was reporting the module as being out of scope when I tried this. Turns out my debugger was just being goofy because I wasn't reloading the plug-in from the Extensions menu after each build -___-

I'll clean this up.

lib/modules/announcements.js
```diff
@@ -199,7 +197,8 @@ function isPostRelevant(post) {
 		return true;
 	}
 
-	if (postTitle.search(`${currentBrowser}`) > -1) {
+
+	if (postTitle.search(`${BrowserDetect.browser}`) > -1) {
```
andytuba: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3442#discussion_r78666796  still relevant 😁

michaelskiles: Sorry, I didn't quite follow your first comment. Are you suggesting removing the unnecessary variable assignment? I'm totally cool with that -  just making sure I'm not blind and missing something else. :)

andytuba:
```
- if (postTitle.search(`${BrowserDetect.browser}`) {
+ if (postTitle.search(BrowserDetect.browser) {
```

If you're not adding any text or formatting to a variable, then simply use the variable.

michaelskiles: *Face Palm* Of course. Thanks!

erikdesjardins:
Thanks!

PR #3443, opened 2016-09-14
dashboard: Restore sidebar dashbord shortcut

larsjohnsen:
Fixes regression from #3394

### DIFF
--- a/lib/modules/dashboard.js
+++ b/lib/modules/dashboard.js
@@ -1,5 +1,4 @@
 import { $ } from '../vendor';
-import { go } from '../core/init';
 import {
 	Alert,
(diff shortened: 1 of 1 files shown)


andytuba:
:+1:

lib/modules/dashboard.js
```diff
@@ -1,4 +1,5 @@
 import { $ } from '../vendor';
+import { go } from '../core/init';
```
erikdesjardins: Nit: use `import * as Init` to distinguish it from utils and other free functions (which can be used freely, whereas I want to discourage the use of `Init` in modules except where necessary)

PR #3445, opened 2016-09-14
dashboard: work around Edge issue with jQuery

### LINKED ISSUE #3357: RES Dashboard widgets not saving.
Repro Steps:
1.       Navigate to https://www.reddit.com/ with RES extension installed
2.       Click on Dashboard
3.       Click +subreddit widget
4.       Enter "Cat" in search field and click on +add
5.       Observe that no results appear and it is not possible to add subreddits 

Confirmed in Edge, other browsers untested.

erikdesjardins:
Fixes #3357

There's an issue in a very specific situation in Edge where the jQuery handler gets dropped (the native `addEventListener` handler is correctly added by jQuery, but when it fires, the `handlers` array in `dataPriv` for the element is empty).

I can't be bothered to expend any more effort trying to figure out what causes this; it's extremely resistant to reproduction. Copying the entire `attachEditComponent()` function verbatim to Codepen, same version of jQuery and everything else? Nope, it all works as expected. Unsurprisingly, trying to hand-code a minimal reproduction is right out. Stepping through with the debugger is unenlightening; I can watch the listener get added correctly and the event fire correctly (to an empty array of handlers), but something is messing with the handlers in between (or the element itself changes somehow) and it's nigh impossible to detect. There are no `.off()` calls in dashboard.js or anything that touches the same elements.

### DIFF
--- a/lib/modules/dashboard.js
+++ b/lib/modules/dashboard.js
@@ -234,7 +234,7 @@ function attachEditComponent() {
 		</div>
 	`);
 
-	$dashboardEditComponent.find('#editRedditForm').submit(e => {
+	$dashboardEditComponent.find('#editRedditForm').get(0).addEventListener('submit', e => {
 		e.preventDefault();
 		let thisBasePath = $('#editReddit').val();
 		if (thisBasePath !== '') {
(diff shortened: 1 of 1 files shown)


andytuba:
I wonder if the event handler is lost when the drop down edit container is added to the body? Anyway, it'd a solution.

PR #3450, opened 2016-09-14
Update dependencies to latest versions

BenMcGarry:
Built fine on Windows for Edge, will monitor build as we have had problems with some dependencies before.


package.json
```diff
@@ -55,46 +55,46 @@
...
     "eslint": "3.5.0",
-    "eslint-import-resolver-webpack": "0.5.1",
+    "eslint-import-resolver-webpack": "0.6.0",
     "eslint-plugin-ava": "3.0.0",
     "eslint-plugin-babel": "3.3.0",
     "eslint-plugin-consistent-return-legacy": "1.0.1",
     "eslint-plugin-dollar-sign": "0.0.5",
     "eslint-plugin-filenames": "1.1.0",
-    "eslint-plugin-import": "1.14.0",
+    "eslint-plugin-import": "1.15.0",
```
erikdesjardins: One thing to do when updating eslint-* dependencies is to check if they have any new rules and add them if they make sense.

In this case, `eslint-plugin-import` has added [no-absolute-path](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-absolute-path.md), which looks like something we'd want to enforce.

So you can add `"import/no-absolute-path": 2,` to .eslintrc.json along with the other `import/` rules

BenMcGarry:
Ava issue appears to be fixed in next release, see: https://github.com/avajs/ava/issues/1017

PR #3451, opened 2016-09-14
imgur: Enforce HTTPS

nfnty:
Fixes mixed content errors.

### DIFF
--- a/lib/modules/hosts/imgur.js
+++ b/lib/modules/hosts/imgur.js
@@ -6,7 +6,7 @@ const imgur = {
 	moduleID: 'imgur',
 	name: 'imgur',
 	domains: ['imgur.com'],
-	logo: '//i.imgur.com/favicon.ico',
+	logo: 'https://i.imgur.com/favicon.ico',
 	options: {
 		'prefer RES albums': {
 			description: 'Prefer RES support for imgur albums rather than reddit\'s built in support',
@@ -47,8 +47,8 @@ const imgur = {
 			}],
 		},
 	},
-	baseURL: 'http://imgur.com/',
-	cdnURL: '//i.imgur.com/',
+	baseURL: 'https://imgur.com/',
+	cdnURL: 'https://i.imgur.com/',
 	apiPrefix: 'https://api.imgur.com/3/',
 	apiID: '1d8d9b36339e0e2',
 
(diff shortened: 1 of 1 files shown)


nfnty:
Fair enough, scaled back to only fix mixed content errors.

nfnty:
Rebased.

honestbleeps:
Viewing on mobile so not easy to check but is this made optional? 

If not we may need to make it so or do extensive testing. I could swear we forced https for imgur a year or two ago and it caused problems for lots of users. It is possible I'm misremembering, but I swear we had problems related to forcing https

nfnty:
No, it's enforced for all links. All API calls and images are already HTTPS. This fixes the problem where Imgur's API returns HTTP links for gifv.

lib/modules/hosts/imgur.js
```diff
@@ -47,8 +47,8 @@ const imgur = {
 			}],
 		},
 	},
-	baseURL: 'http://imgur.com/',
-	cdnURL: '//i.imgur.com/',
+	baseURL: 'https://imgur.com/',
```
andytuba: Until imgur goes https-everywhere, this should stay http as well, or only switch to https if the user opts in.  It's only used in the links, while the mixed content errors result from the cdnURL and URLs returned from the API.

lib/modules/hosts/imgur.js
```diff
@@ -47,8 +47,8 @@ const imgur = {
...
+	cdnURL: 'https://i.imgur.com/',
```
andytuba: It would be nice to provide an option to allow `http://` for uses who might get firewalled via https and don't care about mixed content warnings.

lib/modules/hosts/imgur.js
```diff
@@ -173,12 +173,10 @@ const imgur = {
...
 	},
 	_handleSingleImage(href, info) {
-		const src = info.link
-			.replace('http:', 'https:')
-			.replace(`/${info.id}.`, `/${info.id}${ShowImages.module.options.imgurImageResolution.value}.`);
-
 		return {
-			src,
+			src: info.link
+				.replace(/^http:\/\//, 'https://')
```
andytuba: Same remark about http option on cdnURL

lib/modules/hosts/imgur.js
```diff
@@ -219,13 +217,13 @@ const imgur = {
 			type: 'VIDEO',
 			controls: false,
 			href: ShowImages.module.options.preferredImgurLink.value === 'share' ? `${imgur.baseURL}${info.id}` : `${info.link}`,
-			fallback: info.link,
+			fallback: info.link.replace(/^http:\/\//, 'https://'),
```
andytuba: Same remark about http option on cdnURL

lib/modules/hosts/imgur.js
```diff
@@ -219,13 +217,13 @@ const imgur = {
...
 			caption: info.description,
 			title: info.title,
 			loop: info.looping,
 			muted: true,
 			sources: [{
-				source: info.mp4,
+				source: info.mp4.replace(/^http:\/\//, 'https://'),
```
andytuba: Same remark about http option on cdnURL

PR #3452, opened 2016-09-15
Make channels not recognized as YouTube expandos.

### LINKED ISSUE #3353: YouTube channel links recognised as YouTube videos
https://www.reddit.com/r/RESissues/comments/50sbzt/bug_res_incorrectly_parses_url_to_a_youtube/

Test post: https://www.reddit.com/r/sweden/comments/50quuf/evanescence_pannkaka_av_tales_scales_p%C3%A5_youtube/d768ght

thesbros:
Fixes #3353.

Making the watch/embed not optional fixes this. Also moved the youtu.be in the regex to accommodate for this change.

So far this regex still works on every link I have tested:

```
https youtu.be
http youtu.be
https youtube/watch
http youtube/watch
https youtube/embed
http youtube/embed
https m.youtube/watch
http m.youtube/watch
```

And fixes the issue on the following links (as far as I know these are the only channel links):

```
/c/
/channel/
/user/
/[username]
```

### DIFF
--- a/lib/modules/hosts/youtube.js
+++ b/lib/modules/hosts/youtube.js
@@ -5,8 +5,18 @@ export default {
 	name: 'youtube',
 	attribution: false,
 	domains: ['youtube.com', 'youtu.be'],
-	detect: ({ href }) => (/^https?:\/\/(?:youtu\.be|(?:www\.|m\.)?youtube\.com)\/?(?:watch|embed)?(?:.*?v=|v\/|\/)([\w-]+)/i).exec(href),
-	handleLink(href, [, hash]) {
+	detect: ({ pathname, hostname, search }) => {
+		const params = new URLSearchParams(search);
+		// split path excluding first /
+		const split = pathname.substring(1).split('/');
+		// short url
+		if (hostname.endsWith('youtu.be')) return split[0];
+		// long url
+		return params.has('v') ?
+			params.get('v') : // ?v=
+			(/watch|embed|v/i).exec(split[0]) && split[1]; // watch/embed/v
+	},
+	handleLink(href, hash) {
 		let src = `https://www.youtube.com/embed/${hash}?enablejsapi=1&enablecastapi=1`;
 		const params = getUrlParams(href);
 		if (params.t) {


lib/modules/hosts/youtube.js
```diff
@@ -5,7 +5,7 @@ export default {
 	name: 'youtube',
 	attribution: false,
 	domains: ['youtube.com', 'youtu.be'],
-	detect: ({ href }) => (/^https?:\/\/(?:youtu\.be|(?:www\.|m\.)?youtube\.com)\/?(?:watch|embed)?(?:.*?v=|v\/|\/)([\w-]+)/i).exec(href),
+	detect: ({ href }) => (/^https?:\/\/(?:www\.|m\.)?(?:youtu\.be\/|youtube\.com\/?(?:watch|embed)(?:.*?v=|v\/|\/))([\w-]+)/i).exec(href),
```
erikdesjardins: Actually, there is one thing: I believe we need to replace `watch|embed` with `watch|embed|v` and remove the `v\/` from `(?:.*?v=|v\/|\/)`, so that links like https://youtube.com/v/iwGFalTRHDA will work.

Ajedi32: Honestly this Regex looks to be getting a bit bloated to the point where it's becoming difficult to read. Maybe it's time to drop the Regex here and use [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) or something like that instead? (Does `URL` work in Edge?)

erikdesjardins: We could do something like

```js
({ hostname, pathname, search }) => (
  hostname.endsWith('youtu.be') ?
    (/^\/([\w-]+)/).exec(pathname) :
    (/[?&]v=([\w-]+)/i).exec(search) || (/^\/(?:watch|embed|v)\/([\w-]+)/i).exec(pathname),
)
```

thesbros: I'll experiment with a few different methods to see what works.

thesbros: I also came up with this method:

```
const videoIdRe = /^([\w-]+)/;

[...]
	detect: ({ href }) => {
		let url = new URL(href);
		return url.hostname.endsWith('youtu.be') ?
			videoIdRe.exec(url.pathname.substring(1)) :
			url.searchParams.has("v") ?
				videoIdRe.exec(url.searchParams.get("v")) :
				videoIdRe.exec(url.pathname.split("/")[2]);
```

It works with every URL I've found. (including /v/)

I feel like this is a lot more understandable than the regex. Whatever you think is better I can implement.

erikdesjardins: I think I'd change it a bit more, but that looks good:

```js
detect: ({ pathname, hostname, search }) => {
    const params = new URLSearchParams(search);
    const split = pathname.substring(1).split('/');
    // short url
    if (hostname.endsWith('youtu.be')) return split[0];
    // long url
    return params.has('v') ?
        params.get('v') : // ?v=
        ['watch', 'embed', 'v'].includes(split[0]) && split[1]; // watch/embed/v
},
```

And you'd have to change the handleLink param from `[, hash]` to `hash`

I removed `videoIdRe` because unless the URL is _really_ messed up, it'll just match everything. But I don't feel strongly about this, you could put it back if you want.

thesbros: @erikdesjardins  :+1: Squashed and commited. Only change I did was the array into the regex. The regex is slightly faster, and case-insensitive.

erikdesjardins:
Great, thanks!

PR #3454, opened 2016-09-15
Adds download button to media controls.

### LINKED ISSUE #3440: missing button to download from imgur
https://www.reddit.com/r/RESAnnouncements/comments/50qer8/announcement_res_v500_release/d7lfgbn?context=3

thesbros:
Resolves #3440.

I hope this is the correct way to implement this. I simply added a download button to the media controls that downloads the image/videos source.

I also removed the downloadurl from video-advanced as this replaces that. However from my testing it seems that didn't even work in the first place.

### DIFF
--- a/lib/modules/hosts/imgur.js
+++ b/lib/modules/hosts/imgur.js
@@ -219,7 +219,6 @@ const imgur = {
 			type: 'VIDEO',
 			controls: false,
 			href: ShowImages.module.options.preferredImgurLink.value === 'share' ? `${imgur.baseURL}${info.id}` : `${info.link}`,
-			downloadurl: info.download,
 			fallback: info.link,
 			caption: info.description,
 			title: info.title,
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -23,6 +23,7 @@ import {
 	Thing,
 	addCSS,
 	batch,
+	click,
 	CreateElement,
 	scrollToElement,
 	forEachChunked,
@@ -1133,7 +1134,7 @@ function generateImage(options) {
 
 	setMediaMaxSize(image);
 	makeMediaZoomable(image);
-	const wrapper = setMediaControls(anchor, options.src);
+	const wrapper = setMediaControls(anchor, options.src, options.src);
 	makeMediaMovable(wrapper);
 	keepMediaVisible(wrapper);
 	makeMediaIndependentOnResize(element, wrapper);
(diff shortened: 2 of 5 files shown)


larsjohnsen:
I'm not sure how useful this button is, since it is equivalent to right clicking the video/image and `save video/image as…'.

lib/modules/showImages.js
```diff
@@ -1788,7 +1798,7 @@ function videoAdvanced(options) {
 
 	setMediaMaxSize(vid);
 	makeMediaZoomable(vid);
-	setMediaControls(vid);
+	setMediaControls(vid, null, options.sources[0].source);
```
erikdesjardins: Nit: in general we prefer `undefined` to `null` for omitted params so that default param values work (if there were any)

lib/modules/hosts/imgur.js
```diff
@@ -219,7 +219,6 @@ const imgur = {
 			type: 'VIDEO',
 			controls: false,
 			href: ShowImages.module.options.preferredImgurLink.value === 'share' ? `${imgur.baseURL}${info.id}` : `${info.link}`,
-			downloadurl: info.download,
```
erikdesjardins: Note: the reason imgur stopped working is because they (seemingly) stopped sending a `download` url; this is always undefined.

lib/templates/videoAdvanced.mustache
```diff
@@ -38,9 +38,6 @@
 					{{#source}}
 					<a class="video-advanced-link video-advanced-source" href="{{ source }}">source</a>
 					{{/source}}
-					{{#downloadurl}}
-					<a class="video-advanced-link video-advanced-download res-icon" href="{{ downloadurl }}" title="download"></a>
```
erikdesjardins: There's some CSS in _showImages.scss to style `.video-advanced-download` that can be removed if this is removed

lib/modules/hosts/imgur.js
andytuba: This would be a good place to fill in the downloadurl if it's  not sent anymore, e.g. `info.download || info.sources[0].source`

andytuba:
I think I prefer adding a download button to the media controls buttons:

1. tidy up video controls
1. just keep adding buttons to that toolbar 😁 
1. discoverability (context menu isn't necessarily discoverable, and I personally have trouble reading through all the options)
1. granular control over download url fwiw

lib/modules/hosts/imgur.js
erikdesjardins: I was suggesting something slightly different (using the advanced video controls); I think we can leave this as-is until we have a need for download URLs other than the first source (YAGNI).

`download` [is not in the imgur data model](https://api.imgur.com/models/image) so I don't think it's coming back.

thesbros:
@erikdesjardins Sure thing. Changed `null` to `undefined` and removed unused SCSS.

erikdesjardins:
Looks great, thanks!

PR #3467, opened 2016-09-18
floater: Consistent position, increased visiblity test frequency

larsjohnsen:
The top offset should not depend on tab menu height, as that varies across custom stylesheets (e.g. /r/spacex). This change places the floater at a consistent position (left: 8px, right: 8px) and hides it if the header is visible.

### DIFF
--- a/lib/utils/dom.js
+++ b/lib/utils/dom.js
@@ -188,19 +188,25 @@ export function scrollTo(x, y) {
 	window.scrollTo(x, y - headerOffset);
 }
 
-export const getHeaderOffset = _.once(() => {
-	let header;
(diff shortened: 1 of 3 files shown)


lib/utils/dom.js
```diff
@@ -188,31 +188,33 @@ export function scrollTo(x, y) {
...
+				headers.push(document.getElementById('header'));
 				break;
 			case 'none':
 			default:
 				break;
 		}
 	}
 
-	if (header) {
-		return header.offsetHeight + 6;
```
larsjohnsen: I couldn't find a reason for adding `6` if any header is pinned, so I removed it.

PR #3470, opened 2016-09-18
userTagger: Add option for whether to show tagging icon

larsjohnsen:
Tagging is still functional for when the option is off, but the tagging tool must be accessed through vote tracker.


andytuba:
For touch users, maybe (optionally) show the tag on all users in a post /comment when that thing is selected?

PR #3473, opened 2016-09-19
Placeholder Appveyor deploy script

erikdesjardins:


erikdesjardins:
We're going to deploy Edge from Travis so this isn't likely to be needed.

PR #3474, opened 2016-09-20
Wikipedia expando UI tweaks.

creesch:
Max-width: 110 em.
Prepend the logo before the article to make clear it is not self text.

### DIFF
--- a/lib/modules/hosts/wikipedia.js
+++ b/lib/modules/hosts/wikipedia.js
@@ -56,7 +56,7 @@ export default {
 		// Make all relative links direct
 		$wikiData.find('a[href^="/"]').attr('href', (i, old) => `https://${language}.wikipedia.org${old}`);
 		// Prepend wikipedia logo to make clear it is a wikipedia article
-		$wikiData.prepend('<img src="https://en.wikipedia.org/static/favicon/wikipedia.ico" alt="Hosted on wikipedia" title="Hosted on wikipedia">')
+		$wikiData.prepend('<img src="https://en.wikipedia.org/static/favicon/wikipedia.ico" alt="Hosted on wikipedia" title="Hosted on wikipedia">');
 
 		return {
 			type: 'TEXT',
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -510,7 +510,7 @@ body:not(.res-showImages-displayImageCaptions) {
 
 // Extras for certain text expando's
 .res-media-host-wikipedia {
-	img[alt="Hosted on wikipedia"] {
+	img[alt='Hosted on wikipedia'] {
 		float: left;
 		margin-right: 10px;
 	}


creesch:
Oh sloppy, I'll fix that right away at home.

lib/modules/hosts/wikipedia.js
```diff
@@ -55,6 +55,8 @@ export default {
 		$wikiData.find('.metadata, .hatnote, .mw-editsection, .reference, .references').remove();
 		// Make all relative links direct
 		$wikiData.find('a[href^="/"]').attr('href', (i, old) => `https://${language}.wikipedia.org${old}`);
+		// Prepend wikipedia logo to make clear it is a wikipedia article
+		$wikiData.prepend('<img src="https://en.wikipedia.org/static/favicon/wikipedia.ico" alt="Hosted on wikipedia" title="Hosted on wikipedia">')
```
larsjohnsen: It might better to add `<div class="res-expando-siteAttribution"></div>` in the top of `text.mustache`.

creesch: This does something differently though and site attribution in that place wouldn't work for most other expandos,

creesch:
@erikdesjardins \o/ it passes!

lib/modules/hosts/wikipedia.js
larsjohnsen: Note that this would only affect expandos with `type: 'TEXT`: 
- Deviantart (https://www.reddit.com/r/twitchplayspokemon/comments/53jkmg/kakuna_wars_chapter_six_phone_home/)
- Tumblr (https://www.reddit.com/r/ShingekiNoKyojin/comments/53rhnt/manga_spoilers_theory_titans_true_objectives/)
- Github (https://www.reddit.com/r/privacytoolsIO/comments/53hjkt/howto_privacy_security_conscious_browsing/)

creesch:
Thing is, I didn't add this as an attribution sort of thing. I did add it because it is utterly unclear that it is a wikipedia expando instead of a regular self post. 

I specifically didn't touch the attribution itself so it remains at the bottom just like is done with all other expandos's.

creesch:
Thing is, I didn't add this as an attribution sort of thing. I did add it because it is utterly unclear that it is a wikipedia expando instead of a regular self post. Which in my daily reddit browsing confused me a few times already. So I set out to somehow distinguish between self posts and wikipedia expandos more clearly.

I specifically didn't touch the original attribution itself so it remains at the bottom just like is done with all other hosts. I actually thought about simply moving it up but that doesn't make sense as it would make a primary element that draws attention to itself rather than a secondary element as it is now. 

I don't think people will appreciate the attribution first before every single image or other media type. I personally would be annoyed by it for sure, it only makes sense imho for the wikipedia expandos. 

So I am not going to do the requested changes as I do not agree with them and they are not in line with what I intended to do with this pull request.

honestbleeps:
On review of the linked examples (tumblr/deviantart/github) - the attributions are indeed on the bottom.

If I'm reading this thread correctly, my understanding is that it's being suggested they be put at the top instead of bottom for wikipedia?

why is that?

creesch:
No, I don't want to place the attribution itself on top. I want to better distinguish that this expando is a wikipedia article and not a selfpost.

Because, as I talked about with @andytuba on irc we came to the conclusion that currently when you expand things a wikipedia post almost looks identical to self post. 
This is how the change in pullrequest looks like on reddit. 

image


As you can see it is **not** the attribution it changes, it does simply add the wikipedia W inline with the text so people are immediately attended to the fact that this is in fact a wikipedia article. 

In short: I **don't want to change attribution** I want to **distinguish wikipedia articles better**. 

@larsjohnsen  and @erikdesjardins seem to have reviewed it with a different scope in mind. I guess I should have clarified that in the initial pull request more. But I figured that was clear enough as I talked it over with @andytuba and @thybag on irc.

erikdesjardins:
> it is utterly unclear that it is a wikipedia expando instead of a regular self post

I disagree, I think the sidebar (with embedded image in the majority of pages) is more than enough to disambiguate Wikipedia expandos. I'll also note that we've had no other reports of this (on /r/resissues).

> would make a primary element that draws attention to itself rather than a secondary element as it is now.

Do you think that the icon currently added by this PR is not a "primary element" and/or doesn't draw attention to itself? Is drawing attention not the point?

> I don't think people will appreciate the attribution first before every single image or other media type.

Again, this is _not for all expandos_, just text-only expandos, viz. Github gists, deviantART text expandos (not deviantART images), and Tumblr text expandos (not tumblr image posts). Those are also liable to be confused with self posts since they don't have a sidebar or other non-snudown formatting (besides some embedded images).

>  I talked about with @andytuba on irc we came to the conclusion that currently when you expand things a wikipedia post almost looks identical to self post.

Correct, and he suggested to move the logo (attribution) up:

image

creesch:
> I disagree, I think the sidebar (with embedded image in the majority of pages) is more than enough to disambiguate Wikipedia expandos.

Well you can disagree, but it did happen to me several times. More so when there is no sidebar to speak of. 

> Do you think that the icon currently added by this PR is not a "primary element" and/or doesn't draw attention to itself? Is drawing attention not the point?

Of course it is, but in this case just for these specific expandos as in this case they blend in a bit too much with self posts. 

> Again, this is not for all expandos, just text-only expandos

I must admit that I didn't realize this. My sincere apologies for that, that makes it indeed less of an impact. 
Though the other text expandos (except for deviantart which I couldn't get to work) have more distinct elements that make it more clear it isn't your regular selfpost.

Ajedi32:
FWIW, I agree that right now it's not always entirely clear when a link is expanding to a Wikipedia page vs a self-post or some other text post. The sidebar does help, but it's not always present for all articles.

As for how to distinguish between different sources for text posts, I think I'm leaning towards the position being advocated by @erikdesjardins, as I'm not really convinced this is a problem unique to Wikipedia.

Another alternative to moving the attribution text would be to just move or duplicate the site logo next to to the title. E.g. `{Wikipedia Logo} {Title} \n {Article Text} \n {Full Attribution}`

Issue #3489, opened 2016-09-29
Disable RES on /advertising


andytuba:
It doesn't look pretty.

andytuba:
Night mode!

erikdesjardins:
#3490

PR #3490, opened 2016-09-29
Don't run RES on advertising management pages

andytuba:
#3489

### DIFF
--- a/lib/core/init.js
+++ b/lib/core/init.js
@@ -20,6 +20,7 @@ export function init() {
 	if (
 		!location.hostname.endsWith('.reddit.com') ||
 		(/^(?:i|m|static|thumbs|blog|code)\.reddit\.com$/i).test(location.hostname) ||
+		(/^\/advertising/).test(location.pathname) ||
 		(/\.(?:compact|mobile|json|json-html)$/i).test(location.pathname)
 	) {
 		return;


lib/core/init.js
```diff
@@ -20,6 +20,7 @@ export function init() {
 	if (
 		!location.hostname.endsWith('.reddit.com') ||
 		(/^(?:i|m|static|thumbs|blog|code)\.reddit\.com$/i).test(location.hostname) ||
+		(/^(?:advertising)/).test(location.pathname) ||
```
erikdesjardins: Non-capturing group can be removed

lib/utils/location.js
```diff
@@ -62,12 +62,13 @@ export const regexes = {
 	domain:           /^\/domain\/([\w\.]+)(?:\/|$)/i,
 	composeMessage:   /^\/(?:r\/([\w\.\+]+)\/)?message\/compose(?:\/|$)/i,
 	liveThread:       /^\/live\/(?!create(?:\/|$))([a-z0-9]+)(?:\/|$)/i,
+	promoted:         /^\/promoted/i,
```
erikdesjardins: This is trivial but we should probably have tests for it anyways

andytuba: Actually, I'll take out promoted for now.

PR #3493, opened 2016-09-30
Use derpibooru.org as canonical domain for Derpibooru links.

Ajedi32:
Apparently there's been some sort of issue with Russia's regulatory agencies which resulted in Derpibooru having to deprecate their `.ru` vanity URL. The API no longer works when accessed through the `derpiboo.ru` domain, and instead yields this message:

image

Example: https://derpiboo.ru/api/v2/images/show.json?ids=0

This PR updates the Derpibooru media host to access the API through the `.org` url instead.

Haven't tested this yet as I'm having a bit of trouble getting the build to work, but it's a rather trivial change so I'm fairly confident it won't break anything.


erikdesjardins:
Great, thanks!

PR #3497, opened 2016-10-01
Housekeeping - Update dependencies.

BenMcGarry:
New rules added - discussion?
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unassigned-import.md
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-webpack-loader-syntax.md
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-dynamic-require.md
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-internal-modules.md

NOTE: konami is discontinued on npm


erikdesjardins:
FYI I like to add all rules to config--even if they're just set to off--so we know that a decision has been made. As for those rules:

`no-unassigned-import`: off; this will be triggered by the imports of some legacy jQuery modules, and I think if someone does this by accident it'll be a fairly obvious mistake

`no-webpack-loader-syntax`: if this passes with existing code, turn it on, otherwise turn it off

`no-dynamic-require`: off; we lint against `require` so any usages will probably be dynamic

`no-internal-modules`: off; seems overly restrictive for our use, and similarly to `no-unassigned-import` it should be a fairly obvious mistake

There is also [unambiguous](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/unambiguous.md) which I think we should turn on if it passes with existing code (although it looks like Node is going to use `.mjs` :sob:)

.eslintrc.json
```diff
@@ -76,23 +76,28 @@
...
 		"import/no-namespace": 0,
+		"import/no-webpack-loader-syntax": 0,
 		"import/extensions": [2, { "js": "never" }],
 		"import/order": 2,
 		"import/newline-after-import": 2,
 		"import/prefer-default-export": 0,
+		"import/unambiguous": 0,
 
 		"no-alert": 0,
+		"no-unassigned-import": 0,
```
erikdesjardins: These also belong under `import/`, I guess they just don't throw because they're disabled

.eslintrc.json
```diff
@@ -76,23 +76,28 @@
...
 		"no-bitwise": 0,
 		"no-case-declarations": 0,
 		"no-confusing-arrow": 0,
 		"no-console": 0,
 		"no-continue": 0,
+		"no-dynamic-require": 0,
```
erikdesjardins: Ditto

.eslintrc.json
```diff
@@ -76,23 +76,28 @@
...
 		"no-else-return": 0,
 		"no-empty-function": [2, { "allow": ["arrowFunctions", "methods"] }],
 		"no-extra-parens": [2, "functions"], // "all" conflicts with no-return-assign and some other rules, unfortunately
 		"no-implicit-coercion": [0, { "boolean": false }],
 		"no-inline-comments": 0,
 		"no-inner-declarations": 0, // ES6 has block-scoped functions
+		"no-internal-modules": 0,
```
erikdesjardins: Ditto

.eslintrc.json
```diff
@@ -173,7 +178,7 @@
 		"sort-keys": 0,
 		"sort-vars": 0,
 		"space-before-function-paren": [2, "never"],
-		"strict": [2, "never"], // Babel inserts it for us
+		"strict": [2, "never"], // Babel inserts it for us
```
erikdesjardins: Extra whitespace

.eslintrc.json
erikdesjardins: Same here...

.eslintrc.json
erikdesjardins: ...and here.

erikdesjardins:
Thanks

PR #3500, opened 2016-10-01
Fix multireddit navigation settings button link

### LINKED ISSUE #3495: Multireddit navigation settings button redirects to /me/m/
The button links incorrectly to `/me/m/` instead of `/me/m/multi/` which leads to a 404 page.

Repro steps:

- visit a multi
- hover over a multi
- click on Mulitreddit Navigation

Browser: Firefox 49.0.1 x64, RES: 5.0.2.

Princesseuh:
This fix #3495

I'm kinda surprised the solution was this simple so perhaps I didn't understand the issue fully? Anyway, according to my tests it now work as expected.

### DIFF
--- a/lib/modules/multiredditNavbar.js
+++ b/lib/modules/multiredditNavbar.js
@@ -96,7 +96,7 @@ const populateSectionMenu = baseUrl => [
 	module.options.sectionLinks.value
 		.map(link => populateSectionItem(baseUrl, link))
 		.reduce((prev, curr) => (curr ? prev.add(curr) : prev), $())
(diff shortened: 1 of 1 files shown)


andytuba:
Thanks for the hacktoberfest contribution!

I think your solution is correct, although I haven't tested it yet. I suspect a change elsewhere in the code, e.g. `SettingsNavigation.makeUrlHash`, might have caused this bug. (I might have also shipped it that bug and nobody pointed it out until now.)

erikdesjardins:
Thanks!

PR #3503, opened 2016-10-02
Avoid drawing 'last visited' from storeSubredditVisit setting

### LINKED ISSUE #3448: If Subreddit Manager storeSubredditVisit is disabled, avoid drawing the "last visited" column
Follow-up to #3434

Jtfinlay:
- Disable adding 'Last visited' to the subreddit table when storeSubredditVisit is false.
 - Fixes #3448 

Let me know if you were looking for any additional changes along with this. This change just prevents the columns from being added to the host table.


andytuba:
Thanks for the hacktoberfest contribution! This looks like it fits the bill great. We'll get it tested and merged soon.

If anyone's feeling enterprising, it would be great to also see a PR which moves all this gnarly DOM-building into a mustache template.

erikdesjardins:
Great, thanks!

PR #3506, opened 2016-10-05
Hide and Show the Edit Bar

alberto467:
### The Edit Bar is hidden and shown when the textarea is selected or it contains text.

I love the text editing tools, but they just take up too much space for my liking, and they're useless until you start writing a comment.

This is for my feature request on Reddit: [Reddit Link](https://www.reddit.com/r/Enhancement/comments/54fuy3/feature_request_hide_all_comments_editing_tools/)

I'm sorry if you see some duplicate commits, i'm not exactly the genius of GitHub, but if it's a problem i could fix it.

### DIFF
--- a/lib/modules/commentTools.js
+++ b/lib/modules/commentTools.js
@@ -329,6 +329,18 @@ const initializeEditorTools = _.once(() => {
 		});
 	}
 
+	// Hide the edit bar and show it when the textarea is selected or if it contains text
+	$body.on('focusin', '.usertext-edit textarea', function() {
+		const editBar = $(this).parent().parent().find('.markdownEditor-wrapper.slide');
+		editBar.addClass('slide-show');
+	});
+	$body.on('focusout', '.usertext-edit textarea', function() {
+		const editBar = $(this).parent().parent().find('.markdownEditor-wrapper.slide');
+		if ($(this).val().length === 0) {
+			editBar.removeClass('slide-show');
+		}
+	});
+
 	if (module.options.keyboardShortcuts.value) {
 		$body.on('keydown', '.usertext-edit textarea, #BigEditor textarea, #wiki_page_content, #ban_message', function(e: KeyboardEvent) {
 			if (e.keyCode === KEYS.ESCAPE) {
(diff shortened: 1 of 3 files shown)


alberto467:
@andytuba On Reddit you've talked about the idea of moving the big editor button, i tried it out and it's seems really good, but unfortunately i've had some problem since the edit bar and the big editor button are created from two separate modules.
Basically, you need to check for each module on the other and disable the caching on the edit bar, or edit the already made edit bar on the big editor button module. I wanted to keep the code as clean as possible, but if you want to see it i can make another commit with that.

andytuba:
:+1: This is super neat! I'll review it in a few days, this week is unfortunately pretty busy.

lib/modules/commentTools.js
```diff
@@ -259,6 +259,12 @@ module.go = () => {
 		});
 	}
 
+	// slideEditBar on comments only
+	$body.on('input', 'form[id^=\'commentreply\'] textarea', function() {
```
erikdesjardins: I feel like it should be possible to do this up front in `attachEditorToUsertext` (i.e. never add the handler) instead of covering up the issue later. It will probably be necessary to change [this line](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/41f4e2f757e9d6d2bd1b5863b88d92847130806c/lib/modules/commentTools.js#L606) to `return cachedEditBar.clone();` so the first call to `makeEditBar` doesn't return the cached one.

Unless this happens because reddit clones reply textareas from a textarea that we've already modified. Then things get more _interesting_...

alberto467: Nope, calling the `slideEditBar` function in the `attachEditorToUsertext` function doesn't work for comments, which is why i added this line.

The `updateCounter `function which was already in the code did the same since it needs to be called on comments too, exactly like mine.
That's also why i didn't edit the `makeEditBar `function, i fiddled around with it for a while, but then i saw the `updateCounter `function and i just decided to conform my code to that.

Editing the `makeEditBar `could also allow to move the big editor button, i tried to but then had some problem with the caching/cloning/_interesting things_/???. (See the comment to andytuba)

andytuba: IIRC reddit clones from the textarea that's set up at the top and empties it.

erikdesjardins: `updateCounter` listens for `input` because it needs to update the count on every input. That's not the case here. Right now this reattaches the listeners on every single keypress.

It looks like reddit does clone the textarea when you click "reply"--leaving the listeners attached through `attachEditorToUsertext` behind--so we should move the focus{in,out} listeners up here as delegated listeners (and get rid of this one).

erikdesjardins:
Looks like "add tools on focus" was implemented in #3638 (we're not big fans of the tools disappearing on blur and shifting the page), so I'm just going to close this. You'll still be credited in the changelog, of course :)

PR #3507, opened 2016-10-06
magicPlaceholders flair text for comments replies

amews:
Added flair text magicPlaceholder for comment replies.

### DIFF
--- a/lib/modules/commentTools.js
+++ b/lib/modules/commentTools.js
@@ -832,6 +832,17 @@ const magicPlaceholders = [
 			return date.toDateString();
 		},
 	}, {
+		matches: ['flair'],
+		description: 'On comments pages, returns the flair text of the original link',
+		handle() {
+			if (isPageType('comments')) { 
+				var linkflair = document.querySelector('.linkflairlabel').innerHTML;
+				if (linkflair) {
+					return `${ linkflair }`;
+				}
+			}
+		}
+	}, {
 		matches: ['escaped'],
 		description: 'The selected text, escaped for snudown/markdown. Useful for text emoji like ¯\\_(ツ)_/¯',
 		handle(macroText, selectedText) {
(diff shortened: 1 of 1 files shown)


lib/modules/commentTools.js
```diff
@@ -832,6 +832,17 @@ const magicPlaceholders = [
 			return date.toDateString();
 		},
 	}, {
+		matches: ['flair'],
+		description: 'On comments pages, returns the flair text of the original link',
+		handle() {
+			if (isPageType('comments')) { 
+				var linkflair = document.querySelector('.linkflairlabel').innerHTML;
```
andytuba: Could you use `textContent` instead of `inner HTML`?  There might be extra markup injected in there now or in the future, seems like only the text is interesting for comments.

lib/modules/commentTools.js
```diff
@@ -832,6 +832,17 @@ const magicPlaceholders = [
...
+				if (linkflair) {
+					return `${ linkflair }`;
```
andytuba: You can simply `return linkflairlabel` here, there's no point to using template strings unless you're adding more text.

lib/modules/commentTools.js
```diff
@@ -832,6 +832,17 @@ const magicPlaceholders = [
 			return date.toDateString();
 		},
 	}, {
+		matches: ['flair'],
```
andytuba: Could you change this to the more specific `linkflair`, to disambiguate from user flair?

lib/modules/commentTools.js
andytuba: For bonus points, you could create a new Thing from the OP and use the Thing's method to get the link flair text.

amews:
Thanks @andytuba - note changes.

amews:
@erikdesjardins @andytuba are you happy with these changes in their current state?

erikdesjardins:
Looks good, thanks!

Sorry about the delay.

PR #3511, opened 2016-10-06
Update swirl.xyz expando

klpl:
swirl.xyz has been renamed to vidgif.com

### DIFF
--- a/lib/modules/hosts/vidgif.js
+++ b/lib/modules/hosts/vidgif.js
@@ -1,7 +1,7 @@
 export default {
-	moduleID: 'vidgif',
+	moduleID: 'swirl', // vidgif.com was originally swirl.xyz
 	name: 'vidgif',
-	domains: ['vidgif.com'],
+	domains: ['vidgif.com', 'swirl.xyz'],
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/vidgif.js
```diff
@@ -1,15 +1,15 @@
 export default {
-	moduleID: 'swirl',
-	name: 'swirl',
-	domains: ['swirl.xyz'],
+	moduleID: 'vidgif',
+	name: 'vidgif',
+	domains: ['vidgif.com'],
```
erikdesjardins: Please keep the `swirl.xyz` domain in this array so old links still work (put it after `vidgif.com`)

lib/modules/hosts/vidgif.js
```diff
@@ -1,15 +1,15 @@
 export default {
-	moduleID: 'swirl',
-	name: 'swirl',
-	domains: ['swirl.xyz'],
+	moduleID: 'vidgif',
```
erikdesjardins: We shouldn't change the moduleID since it's used as the settings key (`name` is what's displayed anyways).

Maybe also add `// vidgif.com was originally swirl.xyz` so we remember not to change it in the future.

andytuba: Would the `title` property work here?

PR #3513, opened 2016-10-09
Add class to notifications

### LINKED ISSUE #2835: Add a class to notifications with module id and notification ID
This might be a duplicate

https://www.reddit.com/r/Enhancement/comments/4c9c5f/feature_request_add_a_class_or_id_to_the/

jvtrigueros:
Adds a class to [div.RESNotificationContent](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/notifications.js#L251) constructed from the notifications `moduleID` and `notificationID`.
```html
    <div class="RESNotificationContent moduleID-notificationID"></div>
```

Resolves #2835

### DIFF
--- a/lib/modules/notifications.js
+++ b/lib/modules/notifications.js
@@ -257,6 +257,7 @@ function createNotificationElement(contentObj, notificationType) {
 	});
 	$thisNotification.find('h3').append(contentObj.renderedHeader);
 	$thisNotification.find('.RESNotificationContent').append(contentObj.message);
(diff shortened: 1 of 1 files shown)


jvtrigueros:
The only problem with #2835 specifically is that the Konami module doesn't set `notificationID` or `moduleID`, thus when the PR above creates a class for that notification, we end up with an invalid class name.

Should we expect the module to change to provide a proper `moduleID` and `notificationID`?

andytuba:
Yeah, let's add a a module I'd and notification id to where the Konami notification is fired off.

PR #3514, opened 2016-10-09
Add moduleID and notificationID to Konami notification object

jvtrigueros:
Add `moduleID` and `notificationID` to notification object. This is needed to generate a proper class, see #3513 and #2835.


andytuba:
Note to core contributors: I think a migration is overkill for this.

PR #3516, opened 2016-10-09
Update jpm to fix bug with ff >50. Also other deps.

BenMcGarry:


package.json
```diff
@@ -63,11 +63,11 @@
     "konami": "1.3.3",
     "lodash": "4.16.4",
     "moment": "2.15.1",
-    "snudown-js": "1.4.0-2",
+    "snudown-js": "2.0.0",
```
erikdesjardins: Can you move this to another PR? I want the jpm update for tonight but I think we should leave this to 5.1.0 just in case there are any bugs (shouldn't be; there are extensive tests but I'm playing it extra safe).

PR #3517, opened 2016-10-09
bugfix - maintain subscription comment count on cache clear

### LINKED ISSUE #3515: New comment count prunes subscriptions prematurely
http://www.reddit.com/r/resissues/comments/56ew0h/_/

When cleaning comments, if subscription length option is higher than clean comments option, clean comments should not delete entries that are actively subscribed to.

ssonal:
Should fix #3515

### DIFF
--- a/lib/modules/newCommentCount.js
+++ b/lib/modules/newCommentCount.js
@@ -385,15 +385,10 @@ function updateCommentCount(mycomment) {
 async function cleanCache() {
 	await getLatestCommentCounts();
 	const now = Date.now();
-	const skip = module.options.cleanComments.value < module.options.subscriptionLength.value;
 	for (const i in commentCounts) {
 		if (commentCounts[i] && ((now - commentCounts[i].updateTime) > (86400000 * module.options.cleanComments.value))) {
 			// Do not automatically delete comments belonging to actively subscribed threads
-			if (skip) {
-				// Delete expired subscriptions
-				if (typeof commentCounts[i].subscriptionDate === 'undefined') {
-					delete commentCounts[i];
-				}
+			if (typeof commentCounts[i].subscriptionDate !== 'undefined') {
 				continue;
 			}
 			// commentCounts[i] = null;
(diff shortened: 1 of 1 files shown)


andytuba:
Nice turnaround ☺️ this will certainly solve the problem.  It would be neat to see an iteration that does clean out items which aren't subscribed, but this definitely helps.

ssonal:
I wanted to get this done with minimal effort because I'm not very familiar with the codebase so I didn't want to risk breaking stuff elsewhere.

andytuba:
Nice turnaround! It would be neat to specifically clean out comments that arent actively subscribed, instead of skipping the cleaning entirely.

andytuba:
Wait sorry, I think k you did exactly that and I need to read closer..

lib/modules/newCommentCount.js
```diff
@@ -385,8 +385,13 @@ function updateCommentCount(mycomment) {
 async function cleanCache() {
 	await getLatestCommentCounts();
 	const now = Date.now();
+	const skip = module.options.cleanComments.value < module.options.subscriptionLength.value;
```
erikdesjardins: This is correct, but I think we should leave this out and ignore all comment counts where `subscriptionDate` is defined, regardless of `subscriptionLength`. That way the logic for removing thread subscriptions and removing comment counts are kept separate.

lib/modules/newCommentCount.js
```diff
@@ -389,7 +389,11 @@ async function cleanCache() {
 	for (const i in commentCounts) {
 		if (commentCounts[i] && ((now - commentCounts[i].updateTime) > (86400000 * module.options.cleanComments.value))) {
 			// Do not automatically delete comments belonging to actively subscribed threads
-			if (skip && typeof commentCounts[i].subscriptionDate !== 'undefined') {
+			if (skip) {
+				// Delete expired subscriptions
+				if (typeof commentCounts[i].subscriptionDate === 'undefined') {
```
andytuba: subscriptions should be cleaned per the subscription length option, comment threads that aren't subscribed should be cleared per the clean comments option. More specifically it looks like Erik has the good ideas.

lib/modules/newCommentCount.js
ssonal: Makes sense. Good way to avoid bugs from future changes.

ssonal: I think the latest commit ought to do it.

PR #3520, opened 2016-10-13
Improve neverEndingReddit performance

larsjohnsen:
By manipulating the new `siteTable` before attaching it to the document, it is earlier ready and the webpage is more usable while it is processed.

Typical before (@ #3518):
image

Typical after:
image

### DIFF
--- a/lib/utils/watchers.js
+++ b/lib/utils/watchers.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import { $ } from '../vendor';
-import { Thing, isPageType, observe, downcast } from './';
+import { isPageType, observe, downcast } from './';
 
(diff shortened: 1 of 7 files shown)


lib/utils/watchers.js
```diff
@@ -27,23 +27,16 @@ export function watchForElement(type, callback) {
...
-		const siteTable = Thing.thingsContainer();
+export function newSitetable(siteTable) {
+	// when a new sitetable is loaded, we need to add new observers for selftexts within that sitetable...
+	for (const expando of siteTable.querySelectorAll('.entry div.expando')) {
+		addSelfTextObserver(expando);
+	}
+	return Promise.all(watchers.siteTable.map(v => v(siteTable)));
+}
 
-		if (siteTable) {
```
larsjohnsen: As far as I can see, `siteTable` is only added via NER.

PR #3526, opened 2016-10-15
Update Konami.js library

jvtrigueros:
The Konami library being used in RES was dated to version 1.3.3, which can be obtained via NPM. However, this library is no longer published to NPM, therefore, after talking to @andytuba, we decide that the best course of action was to ~~place the library in `vendors`~~ take needed functionality and port into `easterEgg.js`.

This PR fixes the issue of [Konami broken by default settings](https://www.reddit.com/r/resissues/comments/56gbov/_/), since using the latest version of this library does not require that the user press [`ENTER` to activate the code](https://github.com/snaptortoise/konami-js/issues/9).

~~**Note: I disabled `global-require` for eslint, I wasn't sure if this was the correct approach, please reprimand if incorrect usage.**~~

**UPDATE:**

Most of what I've written above no longer applies. Here's a list of what actually happened:

- Instead of copying code into `vendors`, a modified version of the original code was copied into `easterEgg.js`
- Removed all support for IE 11 < and mobile devices, RES doesn't support those anyway
- Add support to determine if the Konami code was almost entered
- Modified existing library to make the linter happy
- Kept license and attribution for konami-js because we're awesome and respect other open source projects (also cause it's illegal to not do so)


lib/modules/easterEgg.js
```diff
@@ -23,11 +21,10 @@ module.go = () => {
...
 		setTimeout(() => $baconBit.addClass('makeitrain'), 500);
-	};
+	});
 	konami.iphone.load = () => {}; // nix touch support; occasional false positives on touchscreen laptops
-	konami.load();
 };
 
 export function konamiActive() {
-	return konami && konami.almostThere;
+	return konami;
```
andytuba: This is really the crux of the issue: this property is supposed to indicate if the user is 90-100% done with the Konami code sequence and RES shouldn't follow through with keyboard navigation. However,  It looks like  this version of the Konami library doesn't expose that directly. How about for this method, checking if konami exists and if Konami.input mostly matches konami.code?

jvtrigueros: Ah, ok that makes sense. I'll take a stab at solving this problem. Do you have a use case that I can test with? I'm not sure what other actions are being interrupted by the konami code.

erikdesjardins: It looks like `almostThere` was added manually by honestbleeps, since it doesn't show up anywhere [in the source](https://github.com/snaptortoise/konami-js). Alas, [it predates our Git history](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blame/3658a69c4fa1f3829ba680c45f8cc5421d266bd1~3/lib/reddit_enhancement_suite.user.js#L15685) so who knows.

We could just bring back the [old local konami.js](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/cfda25d9d0a707a63d1d490d50ba9f6feee673c5/lib/vendor/konami.js) before it was replaced with the npm module.

jvtrigueros: Yes, I see the issue now. I've resurrected the old local konami code and added these changes. I have questions a specific line. Follow up incoming.

lib/vendor/konami.js
```diff
@@ -0,0 +1,116 @@
...
+				if (konami.input.length > konami.pattern.length)
+					konami.input = konami.input.substr((konami.input.length - konami.pattern.length));
+				if (konami.input == konami.pattern) {
+					konami.code(link);
+					konami.input = "";
+					e.preventDefault();
+					return false;
+				} else if ((konami.input == konami.prepattern) || (konami.input.substr(2, konami.input.length) == konami.prepattern)) {
+					konami.almostThere = true;
+					setTimeout(() => konami.almostThere = false, 2000);
```
jvtrigueros: @andytuba, so here's the line in question. This is line cancels the `almostThere` after a short period of time. While this is nice, I believe it should also reset the input. Otherwise, if I enter all of the Konami sequence save for the last character, and I go browse reddit and then press the last char, `A`, the login menu pops up. What do you say?

lib/modules/easterEgg.js
```diff
@@ -23,9 +25,27 @@ module.go = () => {
 			notificationID: 'konami',
 		});
 		setTimeout(() => $baconBit.addClass('makeitrain'), 500);
-	};
-	konami.iphone.load = () => {}; // nix touch support; occasional false positives on touchscreen laptops
-	konami.load();
+	}
+
+	konami.addEvent(document, "keydown", function (e, ref_obj) {
```
erikdesjardins: At this point we might as well just drop konami.js and move the rest of the implementation (that we need) into this module.

We don't even use half of it anyways (mobile and IE support).

lib/vendor/konami.js
```diff
@@ -0,0 +1,107 @@
+/*
+ * Konami-JS ~
```
erikdesjardins: If we do that, we should copy the attribution parts of this comment into `easterEgg.js` (the copyright notice is required by the MIT license, the rest to be nice).

lib/modules/easterEgg.js
jvtrigueros: That's what it was feeling like. If I get both of your blessings then I shall proceed with this.

lib/modules/easterEgg.js
```diff
@@ -23,11 +21,59 @@ module.go = () => {
...
+ */
+function Konami(callback) {
+	const konami = {
+		input: '',
+		pattern: '38384040373937396665',
+		almostThere: false,
+		prepattern: '383840403739373966',
+		load: () => {
+			document.addEventListener('keydown', event => {
+				konami.input += event ? event.keyCode : '';
```
erikdesjardins: Nit: we can also remove the null-check from this, i.e. just `konami.input += event.keyCode;`

lib/modules/easterEgg.js
```diff
@@ -23,11 +21,59 @@ module.go = () => {
...
+
+				if (konami.input.length > konami.pattern.length) {
+					konami.input = konami.input.substr((konami.input.length - konami.pattern.length));
+				}
+
+				if (konami.input === konami.pattern) {
+					konami.code();
+					konami.input = '';
+					event.preventDefault();
+					return false;
```
erikdesjardins: Nit: we can also remove this `return false;`, it doesn't do anything anymore

erikdesjardins:
Actually those are simple, I'll take care of it

erikdesjardins:
Awesome, thanks!

Issue #3129, opened 2016-07-02
Constrain subreddit logo size in nightmode


andytuba:
https://www.reddit.com/r/Enhancement/comments/4qv723/disabling_sidebar_image_when_disable_subreddit/

because some subreddits have huge logos which are then resized with CSS.

andytuba:
and let's do it in resupdates.css too.

PR #3528, opened 2016-10-16
Added max-width and max-height to #header-img-a (nightmode)

Lakston:
Fixes [#3129](https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/3129)

I couldn't find the resupdates.css though so I only added this in the nightmode file.

There are 2 commits because on my first push the code formatting wasn't correct.

### DIFF
--- a/lib/css/modules/_nightMode.scss
+++ b/lib/css/modules/_nightMode.scss
@@ -123,12 +123,6 @@
 		background-color: rgb(105, 105, 105);
 	}
 
-	#header-img-a {
-		max-width: 200px;
-		max-height: 200px;
-		overflow: hidden;
-	}
-
 	#jumpToContent {
 		color: #ddd;
 		border: 1px solid;
--- a/lib/css/modules/_nightMode.scss
+++ b/lib/css/modules/_nightMode.scss
@@ -122,6 +122,12 @@
 	.liveupdate-home .content {
 		background-color: rgb(105, 105, 105);
 	}
+	
+	#header-img-a {
+    max-width: 200px;
+    max-height: 200px;
+    overflow: hidden;
+	}
 
 	#jumpToContent {
 		color: #ddd;
--- a/lib/css/modules/_nightMode.scss
+++ b/lib/css/modules/_nightMode.scss
@@ -122,11 +122,11 @@
 	.liveupdate-home .content {
 		background-color: rgb(105, 105, 105);
 	}
-	
+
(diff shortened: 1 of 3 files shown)


andytuba:
> There are 2 commits because on my first push the code formatting wasn't correct.

For RES development, you can run `npm lint-fix` to automatically fix and highlight formatting issues. You can also `git commit --amend` or `git rebase` to squash your fixups into one commit, which is fine if nobody else has looked at your branch yet.

andytuba:
resupdates.css is a special snowflake that lives in an external repo, so I'll TODO: @andytuba take care of that.

andytuba:
Instead of "this should take care of", if you phrase it "Fixes #..." or "Closes #..." then GitHub will automatically close the issue when your PR is merged.

andytuba:
I realize now that if the user has subreddit styles enabled in night mode, this might break a little ... but I think this solution is still generally better than that new problem. I'll make a follow up item to move this into "styles that apply when subreddit stylesheet is disabled".

Lakston:
Hey Andy, thanks for all the advice ! I'm fairly new to OS and PRs so it is always nice when someone helps me get better.

lib/css/modules/_nightMode.scss
```diff
@@ -123,6 +123,12 @@
 		background-color: rgb(105, 105, 105);
 	}
 
+	#header-img-a {
```
erikdesjardins: To make sure this only takes effect when subreddit style is disabled, you'll need to change two things:

1. Add `module.bodyClass = true;` [here](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/60be97828c22d9fb4386aa639c8039035c300762/lib/modules/styleTweaks.js#L22) - this will add the `res-styleTweaks` class to `<body>` when the styleTweaks module is enabled

1. Move this CSS to the bottom of `_styleTweaks.css` and wrap it in this condition:

```css
body.res-styleTweaks:not(.res-srstyle-enabled) {
    ...css here...
}
```

andytuba:
> @andytuba we already have machinery for that (res-srstyle-enabled), why not do it now?

Oh right, thanks for the reminder. Yeah, let's push forward with https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3528#discussion_r84326078

erikdesjardins:
Thanks!

PR #3529, opened 2016-10-17
Fix checkElementForMedia for non-Thing links

larsjohnsen:
Resolves https://www.reddit.com/r/RESissues/comments/57xhtn/bug_view_images_not_working/


erikdesjardins:
Thanks!

---

For those keeping score: this is what Flow excels at. It definitely would have caught this for us.

PR #3531, opened 2016-10-18
Improve keyboard navigation (move) responsiveness

larsjohnsen:
Especially noticeable when using NER.

Before:
image

After:
image


lib/modules/selectedEntry.js
```diff
@@ -167,12 +167,36 @@ function _selectedThing() {
...
 
+const runCallbacks = (() => {
+	function runListeners(listeners, newSelected, oldSelected, options) {
+		for (const listener of listeners) listener(newSelected, oldSelected, options);
+	}
+
+	let idleCallback;
+	const requestIdle = window.requestIdleCallback ?
+		window.requestIdleCallback :
+		fn => requestAnimationFrame(() => requestAnimationFrame(fn));
```
larsjohnsen: `requestIdleCallback` is only available in Chrome / Opera

lib/modules/selectedEntry.js
```diff
@@ -167,12 +167,36 @@ function _selectedThing() {
...
+	const runIdle = callback => {
```
larsjohnsen: Should probably be in `utils` alongside with `frameDebounce`

lib/modules/selectedEntry.js
```diff
@@ -167,12 +167,36 @@ function _selectedThing() {
...
+	listeners[when].push(callback);
 }
 
+const runCallbacks = (() => {
+	function runListeners(listeners, newSelected, oldSelected, options) {
+		for (const listener of listeners) listener(newSelected, oldSelected, options);
+	}
+
+	let idleCallback;
+	const requestIdle = window.requestIdleCallback ?
```
erikdesjardins: Should probably move this to `utils/async.js` (maybe not export it for now, until we need something other than `idleThrottle` in module code)

lib/modules/selectedEntry.js
```diff
@@ -167,12 +167,36 @@ function _selectedThing() {
...
+		window.requestIdleCallback : // XXX Only available in Chrome / Opera
+		fn => requestAnimationFrame(() => requestAnimationFrame(fn)); // Approximation for "Don't do anything this frame"
+	const runIdle = callback => {
```
erikdesjardins: IIUC this ensures that the callback will only be called once if `runIdle` is called multiple times before an idle period.

If so, this abstraction can be exported from `async.js` as `idleThrottle` (just like `frameDebounce`, which should really be "throttle" rather than "debounce").

i.e. we would end up with

```js
const runIdle = idleThrottle(runListeners);
```

erikdesjardins:
Also :heart: perf profiles

PR #3533, opened 2016-10-18
Drag resize for iframe expando's

thybag:
Adds a draggable bar to iframe expando's, allowing them to be resized.

Todo:
- [x] Cleanup code
- [x] Maybe port markup to own view (there is a lot of it now)
- [x] Review design

Examples of current design: 
clipboard04

clipboard02


larsjohnsen:
I envisioned something like this
image

thybag:
@larsjohnsen Hmm, i guess i could theoretically use the `res-expando-siteAttribution` bar as the drag bar, which would save on some of the markup. (Having just the handle isn't ideal as its quite easy to drag that behind a sidebar) - what are your thought on giving it a colour, so it stands out from the background in some way?

larsjohnsen:
It's not a problem that it has a separate element, as the `siteAttribution`-element is not always available. As for color, it's probably not desirable that it is too distinct; the icon only needs to convey that it is possible to resize the element by dragging in that area.

larsjohnsen:
> (Having just the handle isn't ideal as its quite easy to drag that behind a sidebar)

To avoid this, modify `makeMediaIndependentOnResize(element, iframeNode);` to `makeMediaIndependentOnResize(element, element);`

thybag:
Unsure if there are any use-cases where it would be useful to disable the "resize" feature on iframe expandos (ie. if the content being embedded just isn't resizable). If so, its probably worth me adding a toggle to disable it etc.

erikdesjardins:
@larsjohnsen whenever you have time can you review?

lib/modules/showImages.js
```diff
@@ -1268,7 +1266,48 @@ function generateIframe(options) {
 	element.independent = true;
 
 	keepMediaVisible(iframeNode);
-	makeMediaIndependentOnResize(element, iframeNode);
+	makeMediaIndependentOnResize(element, element);
```
larsjohnsen: I see that my earlier comment regarding this was wrong. For `makeMediaIndependentOnResize` to work, the 2nd element must be in the subtree of `element`.

lib/modules/showImages.js
```diff
@@ -1268,7 +1266,48 @@ function generateIframe(options) {
...
+
+	// Handle resize for iframe
+	let initialWidth, initialDiagonal, minWidth, left, top, ratio;
+
+	function getDiagonal(x, y) {
```
larsjohnsen: Since documents don't have a fixed aspect ratio, keeping this is unnecessary. IMO it's more user-friendly to resize the iframe to basically w = offsetLeft, h = offsetTop (though this would mean only allow resize initiated from the right bottom corner).

lib/modules/showImages.js
```diff
@@ -1268,7 +1266,48 @@ function generateIframe(options) {
...
+			// Calc ratio
+			ratio = iframeNode.clientHeight / iframeNode.clientWidth;
+
+			minWidth = Math.max(1, Math.min(iframeNode.clientWidth, 100));
+		},
+		onMove(x, y) {
+			const newWidth = Math.max(minWidth, getDiagonal(x, y) / initialDiagonal * initialWidth);
+			const newHeight = `${newWidth * ratio}px`;
+
+			resizeMedia(iframeNode, newWidth);
```
larsjohnsen: The `resizeMedia` function is not suitable to this, since it presumes `height` will automatically follow from the width.

lib/modules/showImages.js
```diff
@@ -1268,7 +1266,48 @@ function generateIframe(options) {
...
+			iframeNode.style.height = newHeight;
```
larsjohnsen: To avoid setting width / height on all these elements, rather use a wrapper which you update the dimensions for, and set the height / width of the children to `100%`.

lib/modules/showImages.js
```diff
@@ -1268,7 +1266,48 @@ function generateIframe(options) {
...
+			resizeHelper.style.pointerEvents = 'all';
+
+			({ left, top } = iframeNode.getBoundingClientRect());
+
+			initialDiagonal = getDiagonal(x, y);
+			initialWidth = iframeNode.clientWidth;
+			// Calc ratio
+			ratio = iframeNode.clientHeight / iframeNode.clientWidth;
+
+			minWidth = Math.max(1, Math.min(iframeNode.clientWidth, 100));
```
larsjohnsen: `minWidth` is already specified in CSS. You may also want to set a `minHeight`.

lib/modules/showImages.js
```diff
@@ -1268,7 +1266,48 @@ function generateIframe(options) {
...
+
+			resizeMedia(iframeNode, newWidth);
+			iframeNode.style.height = newHeight;
+			// Sync heights to overlay
+			resizeHelper.style.width = `${newWidth}px`;
+			dragHandle.style.width = `${newWidth}px`;
+			resizeHelper.style.height = newHeight;
+		},
+		onEnd() {
+			resizeHelper.style.pointerEvents = 'none';
```
larsjohnsen: Since this style is already set in CSS, it would be clearer to just hide the `resizeHelper` element by setting `resizeHelper.hidden = true`.

thybag:
@larsjohnsen Have applied the more straight forward changes you mentioned.

> To avoid setting width / height on all these elements, rather use a wrapper which you update the dimensions for, and set the height / width of the children to 100%.

I'm unsure how to easily achieve this as the structure (assuming I don't move away from the `makeMediaIndependentOnResize` entirely) doesn't really lend itself to it to well.

	<div class="res-iframe-expando">
		<div>
			<div class="res-media-independent">
				<iframe src=""></iframe>
			</div>
		</div>
		<div class="res-iframe-expando-helper" ></div>
		<div class="res-icon res-iframe-expando-drag-handle"></div>
	</div>

I can't really see any obvious way to wrap the elements in a shared sizing container as the helper is absolute, and the drag handle only shares the width (not the height), and the outer container needs the height of the iframe + the drag handle.

Possible I'm over thinking this / missing something obvious.

> Since documents don't have a fixed aspect ratio, keeping this is unnecessary. IMO it's more user-friendly to resize the iframe to basically w = offsetLeft, h = offsetTop (though this would mean only allow resize initiated from the right bottom corner).

In most cases i think locking to an aspect ratio probably is still preferable, as a lot of the media we embed with iframes are things like videos (vimeo/youtube etc.) rather than standard documents - granted there are some exceptions (ala strawpoll). I also think it'd be a bit of a shame to loose the ability to resize from arbitrary locations on the resize bar.

larsjohnsen:
Try this template
```
<div>
	<div class="res-iframe-expando">
		<iframe src="{{url}}" allowFullscreen="true"></iframe>
		<div class="res-icon res-iframe-expando-drag-handle"></div>
	</div>
</div>
```
and then you can pass the `.res-iframe-expando` element to `makeMediaIndependentOnResize` as the second argument. Also, by setting `pointer-events` directly on the iframe, you can forego using a resize helper.

thybag:
@larsjohnsen Thanks :D No idea why just setting the pointer events on the iframe itself didn't occur to me >.<

Have updated with changes, plus reworked it a little to ensure the siteAttribution stuff gets wrapped in the res-iframe-expando (for correct styling). Everything appears to be working for me + codes hopefully a little cleaner.

thybag:
Have updated for compatibility with latest version.

lib/css/modules/_showImages.scss
```diff
@@ -517,6 +517,38 @@ body:not(.res-showImages-displayImageCaptions) {
 	}
 }
 
+.res-iframe-expando {
+
+	&-drag-handle {
```
larsjohnsen: If you wrap the drag handle and `siteAttribution` (`<div class="res-expando-siteAttribution"></div>`, which `addSiteAttribution` will detect) in a new element with rules
```
    display: flex;
    flex-direction: row-reverse;
    justify-content: space-between;
```

lib/css/modules/_showImages.scss
```diff
@@ -517,6 +517,38 @@ body:not(.res-showImages-displayImageCaptions) {
...
+		cursor: nwse-resize;
+		box-sizing: border-box;
+		padding: 4px;
+		text-align: right;
+		display: block;
+		width: 100%;
```
larsjohnsen: you can remove rules `text-align`, `display`, `width`

lib/css/modules/_showImages.scss
```diff
@@ -517,6 +517,38 @@ body:not(.res-showImages-displayImageCaptions) {
...
+	}
+
+	iframe {
+		border: none;
+		min-width: '100px';
+		width: 100%;
+		max-width: initial;
+	}
+
+	.res-expando-siteAttribution {
```
larsjohnsen: and this block will no longer be necessary

lib/css/modules/_showImages.scss
```diff
@@ -517,6 +517,38 @@ body:not(.res-showImages-displayImageCaptions) {
...
+		&::after {
+			content: '\F0A5';
+			transform: scaleX(-1);
+			display: inline-block;
+		}
+	}
+
+	iframe {
+		border: none;
+		min-width: '100px';
```
larsjohnsen: (this is not necessary; determined in resize logic)

lib/css/modules/_showImages.scss
```diff
@@ -517,6 +517,38 @@ body:not(.res-showImages-displayImageCaptions) {
...
+		width: 100%;
+		max-width: initial;
```
larsjohnsen: (this is not necessary; it's not set to anything other anywhere)

thybag:
Have applied the suggested updates :) Everything seems to work - although with this version only the drag-icon can be used for resizing (rather than the full bar).

Is this preferred? Personally I'd be tempted to turn the `res-iframe-expando-footer` in to the drag handle, which would restore the ability to resize using the full bar (although would keep the rest of the improvements.

larsjohnsen:
Allowing drag-start from anywhere on the footer shouldn't be an issue, as long as it don't interfere with siteAttribution.

thybag:
Cool. Have updated to make the footer in to the drag handle :)

lib/modules/showImages.js
```diff
@@ -1352,7 +1350,46 @@ function generateIframe(options: IframeMedia) {
 	element.independent = true;
 
 	keepMediaVisible(iframeNode);
```
larsjohnsen: One last thing: This should be `keepMediaVisible(iframeFooter)` so that the footer follows the iframe when it is moved towards the left.

thybag: Done 👍

thybag:
With the alignment tweaks

clipboard02

PR #3535, opened 2016-10-19
Add ability to swap the preview and editor when using the big editor for comments / text posts

### LINKED ISSUE #2306: Option for big editor to switch textarea/preview panes
https://www.reddit.com/r/Enhancement/comments/3asgn1/any_way_to_move_the_preview_to_the_left_in_the/

dhensche:
Not as glorious as a solution as using flexbox layout, but just an overriding option that swaps the functionality of the .BELeft and .BERight css classes

Fixes #2306

### DIFF
--- a/lib/css/modules/_commentPreview.scss
+++ b/lib/css/modules/_commentPreview.scss
@@ -13,18 +13,6 @@
 	}
 }
 
-.BELeft.swapped {
-	left: auto !important;
-	right: 0 !important;
-	margin-left: 0 !important;
-}
-
-.BERight.swapped {
-	right: auto !important;
-	left: 0 !important;
-	margin-right: 0 !important;
-}
-
 #banned {
 	position: relative;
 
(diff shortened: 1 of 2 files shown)


andytuba:
How about a slightly alternative approach: mark the option as `bodyClass: true` and wrap the CSS rules in `.res-commentTools-swapBigEditorLayout { ... }` ?

dhensche:
Thoughts? I just sort of jumped in without fully digging in to see how all the options operate, but the actual CSS class used for this bodyClass type option is `.res-commentPreview-swapBigEditorLayout`. I'm guessing it is generated based on the modulID?

andytuba:
Oh yeah, I forgot that code got moved into the `commentPreview` module, good catch.

Also, could you edit the title of this PR to describe what it actually does? then put "Fixes #..." into the body of the PR message.

dhensche:
@andytuba How do I edit the body of the PR message?

andytuba:
Find the ✏️  icon in the top right corner of your comments. You may need to log into github.com desktop site.

dhensche:
Duh, I thought my first message at the top of the PR Conversation page was a comment, not the PR message body.

andytuba:
I think it's technically a comment but since it's the first comment I think of it as the PR "body" :grin:

erikdesjardins:
Thanks!

PR #3537, opened 2016-10-20
Updated lines mentioned in #3483 to use _.startCase

### LINKED ISSUE #3483: make more friendly names for the settings
Each of the settings is now a camel cased string, but it's probably quite trivial to change that to a regularly readable with something like [sugarjs](https://github.com/andrewplummer/Sugar/blob/2.0.0/lib/string.js#L1002) or underscore.

dhensche:
fixes #3483


erikdesjardins:
Thanks!

PR #3538, opened 2016-10-20
Added settings gear to dashboard tabs

### LINKED ISSUE #2803: Add settings gear link to appropriate modules on /r/dashboard tabs
maybe with an option to turn them off :P

dhensche:
this only adds settings gears to the tabs added via 'addTab' in the dashboard module. It is not configurable, but we could add it as an option to the dashboard module. 

We could also add these gears to the other tabs (show images/filter visited links), but the simple solution would mean they would be present/not present for each page they are on, not just the dashboard. This would include all subreddits if the user has configured the modules to be present on each subreddit.
fixes #2803


lib/modules/dashboard.js
```diff
@@ -88,14 +89,20 @@ module.go = async () => {
...
+export function addTab(tabID, tabName, tabModule) {
 	const $tabPage = $('<div class="dashboardPane">').appendTo('#siteTable.linklisting').hide();
 
 	const tabMenuItem = CreateElement.tabMenuItem({
 		text: tabName,
 		className: 'res-dashboard-tab',
 		prepend: true,
 	});
+	// if a module with options has been provided then we automatically append the gear link
+	if (tabModule && tabModule.options) {
```
erikdesjardins: I think this function should take `tabModuleId` instead of an entire module; then this logic isn't hidden and the caller can decide whether they want a gear link by (not) passing a moduleId.

Also, technically we could have a module with no options that could still be enabled/disabled, so we'd still want to show the gear.

lib/css/res.scss
```diff
@@ -1836,6 +1836,21 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
...
+	&.selected a {
+		&:first-child {
+			border-right: none;
+		}
+
+		&:last-child {
+			border-left: none;
+		}
+
+		&:only-child {
```
larsjohnsen: If you modify `&.selected a {` to `&.selected a:not(:only-child) {` then you can remove this block.

dhensche: good point, it's been a little while since I've used some of these css selectors :grin:

lib/modules/dashboard.js
dhensche: okay. I had understood #2803 as only wanting the settings gear if there were additional options for the module

erikdesjardins:
Great, thanks!

PR #3539, opened 2016-10-21
Use giphy api to load inline giphy gifs

### LINKED ISSUE #3005: giphy host should use giphy api and support long url
https://github.com/Giphy/GiphyAPI#get-gif-by-id-endpoint

because the direct urls are prone to change

http://giphy.com/gifs/screamqueens-emma-roberts-scream-queens-chanel-oberlin-3o85xCu8FfNyWoXYQg

dhensche:
fixes #3005


dhensche:
So I think the npm servers were having an issue when the AppVeyor build was supposed to happen...

andytuba:
Looking pretty good to me. Haven't tested yet, though. @erikdesjardins any final comments?

lib/modules/hosts/giphy.js
```diff
@@ -1,27 +1,35 @@
...
+
 export default {
 	moduleID: 'giphy',
 	name: 'giphy',
 	domains: ['giphy.com'],
-	logo: '//giphy.com/static/img/favicon.png',
-	detect: ({ pathname }) => (/^\/gifs\/(?:\w+-)*(.*?)(\/html5)?$/i).exec(pathname),
-	handleLink(href, [, id, isHtml5]) {
+	logo,
+	detect: ({ pathname }) => (/^(?:\/gifs|\/media|)\/(?:\w+-)*([^/.]*)(\/html5)?.*$/i).exec(pathname),
```
erikdesjardins: Nit: `.*$` at the end of the regex is redundant; it can be removed. But to make the regex a bit more precise, I think we should replace it with `(?:\/|$)`

Is the trailing `|` in `(?:\/gifs|\/media|)` intended? If so, do you have an example link? If not, we can make that a bit cleaner: `\/(?:gifs|media)`

And while we're at it might as well replace `([^/.]*)` with `([^/.]+)` since an empty id isn't very useful (I'm not sure why it was this way originally).

erikdesjardins: Also, we can remove `(\/html5)?`, see below (my other suggestions still apply as-is)

lib/modules/hosts/giphy.js
```diff
@@ -1,27 +1,35 @@
...
+	async handleLink(href, [, id, isHtml5]) {
+		const { data } = await ajax({
+			url: `http://api.giphy.com/v1/gifs/${id}?api_key=dc6zaTOxFJmzC`,
```
erikdesjardins: `https` please, and we can make the query string a bit cleaner:

```js
url: `https://api.giphy.com/v1/gifs/${id}`,
data: { api_key: 'dc6zaTOxFJmzC' },
```

lib/modules/hosts/giphy.js
```diff
@@ -1,27 +1,35 @@
...
+			type: 'json',
+		});
+
 		if (isHtml5) {
```
erikdesjardins: This condition can be removed completely (and always take the first branch).

It seems that that giphy API always returns an mp4, and if it doesn't we'll just fall back to the same `.gif`.

lib/modules/hosts/giphy.js
dhensche: For the second point, it was to deal with links to i.giphy.com which have links with the following format: http://i.giphy.com/MhR4xjfLE2eNq.gif

lib/modules/hosts/giphy.js
dhensche: I'm fine with removing the special html5 logic. Based on the data returned from the giphy api, I didn't think it was necessary either but didn't know if there was another reason to keep it around.

erikdesjardins:
Sweet, thanks!

PR #3543, opened 2016-10-21
Never ending comments

### LINKED ISSUE #3432: Never-Ending Comments
1. When user scrolls near the bottom of the page, click the first top-level "load more comments" button at the bottom of the comments list.
2. When user pauses to start reading and a comment tree contains "load more comments", optionally (default off) click the first 2 (or optionally 0->10).  bonus points: min/max depth option

dhensche:
Added a Never Ending Comments module that by default only auto-loads
the "load more comments" link at the bottom of the page. There is an
option to also load these links for any child comments. When this is
enabled all links that are in the viewport are loaded, started from
the top. This causes lower links to potentially be pushed out of the
viewport and then not loaded until the user scrolls them down into
view again. On the other hand, it allows for nested "load more comments"
links to be loaded automatically as they appear.

fixes #3432

### DIFF
--- a/lib/modules/neverEndingComments.js
+++ b/lib/modules/neverEndingComments.js
@@ -1,19 +1,22 @@
+/* @flow */
+
 import * as _ from 'lodash';
 import { $ } from '../vendor';
+import { Module } from '../core/module';
 import { elementInViewport } from '../utils/dom';
 import * as SettingsConsole from './settingsConsole';
 
-export const module = {};
+export const module: Module<*> = new Module('neverEndingComments');
 
-module.moduleID = 'neverEndingComments';
-module.moduleName = 'Never Ending Comments';
-module.category = 'Comments';
-module.description = 'For those ultra-long comment pages, this will keep them flowing';
+module.moduleName = 'necName';
+module.category = 'commentsCategory';
+module.description = 'necDescription';
 module.options = {
 	loadChildComments: {
 		type: 'boolean',
 		value: false,
-		description: 'Whether or not on-screen \'load more comments\' links should be expanded when paused',
+		description: 'necLoadChildCommentsDesc',
+		title: 'necLoadChildCommentsTitle',
 	},
 };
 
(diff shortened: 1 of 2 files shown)


dhensche:
If it is desired I can change the implementation (when the load child comments option is enabled) to load the bottommost links first, resulting in all links in the viewport when scrolling stops being loaded (instead of lower ones being pushed offscreen).

lib/modules/neverEndingComments.js
```diff
@@ -0,0 +1,50 @@
...
+	'comments',
+];
+
+module.go = () => {
+	window.addEventListener('scroll', _.debounce(handleScroll, 300));
+};
+
+function visibleLoaders() {
+	const loaders = module.options.loadChildComments.value ? $('span.morecomments a') : $('span.morecomments:last a');
+	return loaders.filter((i, el) => elementInViewport(el));
```
andytuba: How about replacing `.filter(...)` with `.find(...)` for efficiency?

dhensche:
What sort of profiling information would you want? I haven't noticed any impact to scrolling during testing but obviously it working on my machine is not indicative of performance on other machines

lib/modules/neverEndingComments.js
dhensche: What about eagerly breaking out of the filtering function? something like 
```javascript
let found = false;
return loaders.filter((i, el) => {
	if (found) return false;
	found = elementInViewport(el);
	return found;
});
```

not a huge fan of using the flag though...

erikdesjardins: Either:

Convert the jQuery object to an array and call .find on that:
```js
return loaders.toArray().find(...);
```

Use `document.querySelectorAll` instead of jQuery, convert that to an array, and call .find on that:
```js
const loaders = ... document.querySelectorAll('span.morecomments a') ...
return Array.from(loaders).find(...);
```

Use `document.querySelectorAll` instead of jQuery, and use its iterator (actually jQuery is iterable too):
```js
const loaders = ... document.querySelectorAll('span.morecomments a') ...
for (const el of loaders) {
    if (elementInViewport(el)) return el;
}
```

erikdesjardins:
Great, thanks!

PR #3550, opened 2016-10-23
Support only Safari 10+

erikdesjardins:
There doesn't seem to be a way to set a minimum Safari version in the manifest 😕

[without whitespace changes](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3550/files?w=1)

### DIFF
--- a/.babelrc.json
+++ b/.babelrc.json
@@ -16,24 +16,6 @@
 		"lodash"
 	],
 	"env": {
-		"safari": {
-			"presets": [
-				"es2015"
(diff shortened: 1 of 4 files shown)


honestbleeps:
I don't even have a machine that has / will have Safari 10...

erikdesjardins:
Hah, well, that's a non-starter.

I thought Yosemite could run on any Mac made in the last 7 years, some even older...is yours really that old?

PR #3553, opened 2016-10-23
Fix various accessibility issues

wojtekmaj:
As DIVs were used, the elements were not properly behaving according
to accessibility guidelines, most notably weren't accessible using
keyboard

Please check if the buttons are rendering properly, they should have, although I couldn't get my dev env set up fully, so I just applied my changes using developer tools to make sure they look okay.

### DIFF
--- a/lib/css/res.scss
+++ b/lib/css/res.scss
@@ -188,6 +188,13 @@ body.res-console-open {
 	display: inline-block;
 }
 
+#RESSettingsButton {
+	user-select: none;
+	font-size: 0;
+	border: none;
+	background-color: transparent;
+}
+
 .RESDropdownList {
 	position: absolute;
 	list-style-type: none;
@@ -303,6 +310,7 @@ body.res-console-open {
 	text-align: center;
 	width: 24px;
 	height: 24px;
+	padding: 0;
 	z-index: 1000;
 	cursor: pointer;
 	box-sizing: border-box;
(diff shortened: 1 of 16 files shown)


andytuba:
Thanks, this is a great contribution, and more accessibility is always better.

I'm a bit concerned about adding `!important` in the CSS though. It looks like you're trying to fix styling issues created by switching from `div` and `span` to more accessible `button` tags. Is there a good compromise, like adding `aria` tags or using `a` instead of `button`?

andytuba:
Also the whitespace is a little wonky, maybe somebody else can help with the menial lint fixups right before merging.

You mentioned trouble setting up your dev environment. Did you get stuck at some particular step that we could help with?

wojtekmaj:
You are absolutely right @andytuba and I share your concerns. It's the matter [CSS specifity](https://css-tricks.com/specifics-on-css-specificity/). I could switch these classed from `.className` to `tagName.className` to make them "stronger" than `tagName { ... }`, which probably is not a bad idea, I just need to make sure it won't break anything. Hang on then :)

erikdesjardins:
`.className` is already more specific than `tagName` (inline styles > ids > classes > elements), so I think most of those selector changes shouldn't be necessary.

wojtekmaj:
Ah, nice catch. I don't know why it didn't work for me at first. Fixed :)

wojtekmaj:
Could we have a second look on this? It's likely to be affected by i18n changes a lot in the nearest future.

PR #3557, opened 2016-10-24
Support yarn

erikdesjardins:
We will continue to support developing with `npm` unless there's some big improvement that requires dropping it [read: for the foreseeable future].

https://github.com/yarnpkg/yarn


Ajedi32:
👍 Cool.

Might want to tell Travis to cache the Yarn cache directory as described in [the docs](https://yarnpkg.com/en/docs/install-ci#travis-tab):

```
cache:
  directories:
  - $HOME/.yarn-cache
```

mc10:
Is the `yarn.lock` supposed to be checked into version control?

erikdesjardins:
Planning on it, going to wait a few builds so we have a baseline (since it made zero difference before: #3298).

erikdesjardins:
Ah, the question everyone asks ([yes](https://yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control)).

PR #3558, opened 2016-10-25
Open settings URLs in same tab

### LINKED ISSUE #2968: "open links in comments in new tabs" should ignore {currenturl}#!settings URLs

ssonal:
Settings URLs in comments should be opened in the same tab even if commentLinksNewTabs is set.

I think this addresses the issue but I might've misunderstood the scope of this change. 

Closes #2968

### DIFF
--- a/lib/modules/betteReddit.js
+++ b/lib/modules/betteReddit.js
@@ -219,8 +219,12 @@ module.go = () => {
 function commentsLinksNewTabs(ele = document.body) {
 	const links = ele.querySelectorAll('.thing div.md a');
 	links::forEachChunked(link => {
-		link.target = '_blank';
-		link.rel = 'noopener noreferer';
+		const { baseURI, href } = link;
+
+		if (!(href.includes(baseURI) && (href.includes('#!settings') || href.includes('#res:settings')))) {
+			link.target = '_blank';
+			link.rel = 'noopener noreferer';
+		}
 	});
 }
 
--- a/lib/modules/betteReddit.js
+++ b/lib/modules/betteReddit.js
@@ -14,6 +14,7 @@ import {
 	watchForElement,
 } from '../utils';
 import { ajax } from '../environment';
+import { isSettingsUrl } from './settingsNavigation';
 import * as AccountSwitcher from './accountSwitcher';
 import * as SubredditManager from './subredditManager';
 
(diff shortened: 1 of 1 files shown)


lib/modules/betteReddit.js
```diff
@@ -219,8 +219,12 @@ module.go = () => {
 function commentsLinksNewTabs(ele = document.body) {
 	const links = ele.querySelectorAll('.thing div.md a');
 	links::forEachChunked(link => {
-		link.target = '_blank';
-		link.rel = 'noopener noreferer';
+		const { baseURI, href } = link;
+
+		if (!(href.includes(baseURI) && (href.includes('#!settings') || href.includes('#res:settings')))) {
```
andytuba: The `href.includes(...)` checks are already written for you, so you can `import * as SettingsNavigation from './settingsNavigation';` at the top of this file (in alphabetical order please) and replace the `href.includes...` part with `SettingsNavigation.isSettingsUrl(href)`.

andytuba: I also would personally do `if (!(is this a settings link)) return;` to preserve the blame and save indentation, but that's just a matter of taste.

andytuba: It would be nice to check if the link's pathname is different from the location (like https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2968#issuecomment-250956314 describes), for cases like linking to a particular post listings (nsfw, some subreddit/domain) with the filteReddit settngs already open.  However, we could save that for another time, since the vast majority of settings links should load in the current page.

ssonal: Sounds good! Will make the first two changes.

With ```if (href.includes(baseURI) && isSettingsUrl(href))``` I check for changes in location and only then checks for the settings hash. 
So in that case, responses from the AutoModerator with say a link to settings at ```https://www.reddit.com/#!settings/search``` still open in a new tab.
 

Is this what the comment described? Can you give me a more concrete example on Reddit so I can test this?

andytuba: I was thinking a link eg to https://www.reddit.com/domain/siteIWantToFilter.com#res:settings/filteReddit/domain, but don't worry about it -- I'd rather get this in and working well for the general case.

andytuba:
Looks good to me, thanks! Anyone else wanna double-check we didn't miss some crucial bit of logic?  (Check out the collapsed conversation, too, where I talk about how it's good and fine to open settings links in the current tab by default.)

erikdesjardins:
Thanks!

PR #3563, opened 2016-10-26
Date custom filter condition

andytuba:
Filter posts if today is before or on/after a certain date(time). Useful for filtering noise leading up to an event such as an election or hiding spoilers around a release date.

I opted for plaintext since momentjs has an excellent parser. A future update could add a `datetime` builderField to settingsConsole.


larsjohnsen:
I don't think it is necessary to add any additional fields, since it appears that this is a browse context.

larsjohnsen:
Though moment could also be used in post cases. A generic `postDate` could likely replace `postAfter` and `postAge`

PR #3565, opened 2016-10-26
Use improved Thing methods for PostNavigator ShowLink

### LINKED ISSUE #3488: PageNavigator ShowLink shows empty () on ads pages
Find an ad post, eg from the spotlight,, and look in the top right corner

andytuba:
Fixes #3488

### DIFF
--- a/lib/utils/Thing.js
+++ b/lib/utils/Thing.js
@@ -143,7 +143,16 @@ export default class Thing {
 	}
 
 	getTitleElement() {
-		return this.entry.querySelector('a.title, a.search-title');
+		return this.entry.querySelector('a.title, a.search-title') ||
+			this.entry.querySelector('.title');
+	}
+
+	getTitleUrl() {
+		const element = this.getTitleElement();
+		if (element) {
+			return element.href;
+		}
+		return '';
 	}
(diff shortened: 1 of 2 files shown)


lib/utils/Thing.js
```diff
@@ -218,8 +245,28 @@ export default class Thing {
...
+		}
+		return `/domain/${this.getPostDomain()}/`;
+	}
+
 	getPostDomainLink() {
-		return this.thing.querySelector('.domain > a');
+		return this.thing.querySelector('.domain a');
+	}
+
+	getPostDomainText() {
```
andytuba: I feel like there are some redundancies here that could be consolidated.

erikdesjardins: You could have stuff like:

```js
_getDomainFromAttrs() {
	return this.thing.getAttribute('data-domain');
}
```

but I don't think it's a huge deal.

andytuba: Yeah, I think I'll just stick with some minor duplication. Even if that logic changes a little, it's all tidily grouped together.

andytuba:
:haircut_man:

PR #3568, opened 2016-10-28
Don't show QuickMessage is modifier key is pressed

### LINKED ISSUE #2027: message mods / user button shouldn't launch QuickMessage if any modifier key is held down
it breaks browser functionality (e.g. right-click menu, open in new tab) and may cause weirdness (@allthefoxes)

sionide21:
Fixes #2027


erikdesjardins:
Thanks!

PR #3570, opened 2016-10-30
Change XKCD expandos to use XKCD API instead of noembed

### LINKED ISSUE #3384: xkcd parsing issues on certain posts
https://www.reddit.com/r/RESissues/comments/514vxm/bug_xkcd_expando_cuts_off_title_text_of_859/

Overv:
The problem with the noembed API Is that it returns invalid HTML for
certain comics that include quotes in the alt text.

Example:
https://noembed.com/embed?url=http://xkcd.com/1726/
https://www.reddit.com/r/programming/comments/5a5tba/utf8_everywhere_now_please/d9e0jzn/

Unfortunately the noembed repository does not seem to be very active
anymore, so I think the best solution is to fix this on the RES side by
switching to the official XKCD API.

Edit(erikdesjardins): fixes #3384

### DIFF
--- a/chrome/manifest.json
+++ b/chrome/manifest.json
@@ -64,6 +64,7 @@
 		"*://backend.deviantart.com/oembed?url=*",
 		"*://api.gyazo.com/api/oembed*",
 		"*://codepen.io/api/oembed*",
-		"*://api.tumblr.com/v2/blog/*/posts?api_key=*&id=*"
+		"*://api.tumblr.com/v2/blog/*/posts?api_key=*&id=*",
+		"https://xkcd.com/*/info.0.json"
 	]
 }
--- a/edge/manifest.json
+++ b/edge/manifest.json
@@ -63,6 +63,7 @@
 		"*://backend.deviantart.com/oembed?url=*",
 		"*://api.gyazo.com/api/oembed*",
 		"*://codepen.io/api/oembed*",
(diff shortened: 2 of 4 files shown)


chrome/manifest.json
```diff
@@ -64,6 +64,7 @@
 		"*://backend.deviantart.com/oembed?url=*",
 		"*://api.gyazo.com/api/oembed*",
 		"*://codepen.io/api/oembed*",
-		"*://api.tumblr.com/v2/blog/*/posts?api_key=*&id=*"
+		"*://api.tumblr.com/v2/blog/*/posts?api_key=*&id=*",
+		"https://xkcd.com/*"
```
Ajedi32: Couldn't this be made a bit more restrictive, like the other permissions above it? (I.E. `https://xkcd.com/*/info.0.json`)

Ajedi32:
We were avoiding this before due to it requiring an extra permission, but there have been some rather recent user-experience and API improvements to the way the permissions system in RES works so maybe now this way would work fine.

chrome/manifest.json
Overv: Good idea, I've made the URL more restrictive.

erikdesjardins:
Yeah, now it's not as big of a deal to require permissions (we should still avoid it where possible though)

erikdesjardins:
Thanks!

PR #3583, opened 2016-11-03
Remove hideUnreadCount

### LINKED ISSUE #3348: Orangered hideUnreadCount turned on by migration?

erikdesjardins:
<as previously discussed>&nbsp;

This was erroneously turned on by a migration for some users.

The only reason it exists was to preserve legacy behaviour: in the past reddit didn't show a message count, so RES had an option called `showUnreadCount` which would generate one. When this was implemented on reddit's side and removed from RES, the option was inverted to `hideUnreadCount`.

But it's kinda silly, and I can't imagine why you'd want to hide the message count alone (without, e.g. also hiding the orangered envelope).

So in lieu of trying to migrate it back, we'll just remove it.

If anyone still wants to hide the message count, they can use a CSS snippet: `.message-count { display: none !important; }`

Resolves #3348

### DIFF
--- a/lib/core/migrate.js
+++ b/lib/core/migrate.js
@@ -203,12 +203,6 @@ const migrations = [
 			await migrators.generic.moveOption('stylesheet', 'loadSubredditStylesheets', 'stylesheet', 'loadStylesheets', migrators.specific.subredditToUrl);
 		},
 	}, {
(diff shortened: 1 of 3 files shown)


andytuba:
lgtm

PR #3585, opened 2016-11-05
add a count of multis a SR is in next to the subscribe button

### LINKED ISSUE #3047: Add "how many multis are you in" to subreddit subscribe button
https://www.reddit.com/r/Enhancement/comments/4mknme/feature_request_highlightalter_subscribe_button/

entails subreddit manager tracking your multis.  bonus: show them on the "my subredits" dropdown list

dhensche:
adds a count badge to the right hand SR info area and the subreddit info popover seen when hovering over a SR link

fixes #3047

### DIFF
--- a/lib/css/modules/_nightMode.scss
+++ b/lib/css/modules/_nightMode.scss
@@ -1148,7 +1148,8 @@
 	.snoovatar-page .gold-accent a,
 	.combined-search-page .search-result-subreddit .search-subscribe-button .add.active,
 	.combined-search-page .search-result-subreddit .search-subscribe-button .remove.active,
-	.fancy-toggle-button a {
+	.fancy-toggle-button a,
+	.fancy-toggle-button .multi-count {
 		color: rgb(255, 255, 255);
 	}
 
--- a/lib/css/res.scss
+++ b/lib/css/res.scss
@@ -939,6 +939,13 @@ a.RESConsoleButton {
 	}
 }
 
+span.active.multi-count {
+  background-color: #FF7500;
+  position: relative;
+  top: -5px;
+  left: -5px;
+}
+
 .SearchRES-result-header {
 	margin-bottom: 12px;
 	color: #666;
--- a/lib/css/res.scss
+++ b/lib/css/res.scss
@@ -922,7 +922,7 @@ a.RESConsoleButton {
 
 // Universal subreddit toggle button (subscribe, filter, shortcut, etc)
 // Borrowed from fancy-toggle-button
-.res-fancy-toggle-button {
+%res-fancy-toggle-button {
 	font-weight: bold;
 	font-size: 10px;
 	color: white;
@@ -933,17 +933,20 @@ a.RESConsoleButton {
 	background-image: linear-gradient(to bottom, rgb(123, 184, 80), rgb(117, 168, 73));
 	border-radius: 3px;
 	cursor: pointer;
+}
+
+.res-fancy-toggle-button {
+	@extend %res-fancy-toggle-button;
 
 	&.remove {
 		background-image: linear-gradient(to bottom, rgb(207, 97, 101), rgb(190, 94, 96));
 	}
 }
 
-span.active.multi-count {
-  background-color: #FF7500;
-  position: relative;
-  top: -5px;
-  left: -5px;
+span.multi-count {
+	@extend %res-fancy-toggle-button;
+	background: #FF7500;
+	color: white;
 }
 
 .SearchRES-result-header {
--- a/lib/modules/subredditInfo.js
+++ b/lib/modules/subredditInfo.js
@@ -205,6 +205,9 @@ async function showSubredditInfo(ele, update) {
 		const $subscribeToggle = $('#RESHoverInfoSubscriptionButton');
 		$subscribeToggle.attr('data-subreddit', jsonData.data.display_name.toLowerCase());
 		updateToggleButton($subscribeToggle[0], subscribed);
+		if (Modules.isEnabled(SubredditManager)) {
+			$subscribeToggle.after(SubredditManager.getMultiCounts(jsonData.data.display_name));
+		}
 		$subscribeToggle.fadeIn('fast');
 	}
 
--- a/lib/modules/subredditInfo.js
+++ b/lib/modules/subredditInfo.js
@@ -206,7 +206,7 @@ async function showSubredditInfo(ele, update) {
 		$subscribeToggle.attr('data-subreddit', jsonData.data.display_name.toLowerCase());
 		updateToggleButton($subscribeToggle[0], subscribed);
 		if (Modules.isEnabled(SubredditManager)) {
-			$subscribeToggle.after(SubredditManager.getMultiCounts(jsonData.data.display_name));
+			$subscribeToggle.after(await SubredditManager.getMultiCounts(jsonData.data.display_name));
 		}
 		$subscribeToggle.fadeIn('fast');
 	}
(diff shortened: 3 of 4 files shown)


lib/css/res.scss
```diff
@@ -922,7 +922,7 @@ a.RESConsoleButton {
 
 // Universal subreddit toggle button (subscribe, filter, shortcut, etc)
 // Borrowed from fancy-toggle-button
-.res-fancy-toggle-button {
+%res-fancy-toggle-button {
```
andytuba: %?

lib/css/modules/_nightMode.scss
```diff
@@ -1148,7 +1148,8 @@
 	.snoovatar-page .gold-accent a,
 	.combined-search-page .search-result-subreddit .search-subscribe-button .add.active,
 	.combined-search-page .search-result-subreddit .search-subscribe-button .remove.active,
-	.fancy-toggle-button a {
+	.fancy-toggle-button a,
+	.fancy-toggle-button .multi-count {
```
andytuba: :+1: so considerate

lib/modules/subredditInfo.js
```diff
@@ -204,6 +204,9 @@ async function showSubredditInfo(ele, context, update) {
 		const $subscribeToggle = $('#RESHoverInfoSubscriptionButton');
 		$subscribeToggle.attr('data-subreddit', jsonData.data.display_name.toLowerCase());
 		updateToggleButton($subscribeToggle, subscribed);
+		if (Modules.isEnabled(SubredditManager)) {
+			$subscribeToggle.after(SubredditManager.getMultiCounts(jsonData.data.display_name));
```
andytuba: What do you think of not showing the count if it's 0?

lib/modules/subredditManager.js
```diff
@@ -1072,24 +1074,50 @@ async function getSubreddits() {
...
+	populateSubredditDropdown();
+}
+
+let myMultis = [];
+async function getMultis() {
+	if (gettingMultis) return;
+	gettingMultis = true;
+
+	myMultis = [];
+	const multis = await ajax({
```
andytuba: This should probably be cached for at least a day, and keyed by `${loggedInUser()}/multis` or something.. @erikdesjardins remind me how to phrase this?

lib/modules/subredditManager.js
```diff
@@ -1072,24 +1074,50 @@ async function getSubreddits() {
...
-	gettingSubreddits = false;
-	populateSubredditDropdown();
+	return data;
 }
 
 // if badJSON is true, then getSubredditJSON ran into an error...
 async function populateSubredditDropdown(sortBy = 'subreddit', badJSON) {
 	$srList.html('');
-
+	console.log(mySubreddits);
```
andytuba: This should be tidied up.

lib/modules/subredditManager.js
```diff
@@ -1263,6 +1291,12 @@ async function populateSubredditDropdown(sortBy = 'subreddit', badJSON) {
 	$srList.append(theBody);
 }
 
+export function getMultiCounts(displayName) {
+	const count = myMultis.filter(multi => !!multi.data.subreddits.find(sr => sr.name === displayName)).length;
```
andytuba: _.differenceWith / _.differenceBy might be your friend here.

andytuba:
Looking mostly good, thanks!  Mostly I'm concerned about calling `getMultis` before we actually need that data, and making sure to cache it.

lib/modules/subredditManager.js
dhensche: I figured it should be cached, just was unsure how that worked (is the caching smart so that if a user edits/creates a multi the cache is invalidated?)

lib/modules/subredditManager.js
dhensche: ah yes, debugging statements...

lib/modules/subredditInfo.js
dhensche: the 'getMultiCounts' just returns a blank string if the count is 0, resulting in no extra HTML being added. If you have a better solution, I'm all for it

lib/modules/subredditManager.js
andytuba: RES would need to monitor the two multi creation UI elements (frontpage sidebar and subscribe button popup)  and bust the cache manually. definitely worth doing but imo we could defer that to a separate PR.

lib/modules/subredditManager.js
```diff
@@ -176,6 +176,7 @@ module.go = async () => {
 	}
 
 	await getLatestShortcuts();
+	await getMultis();
```
andytuba: You could `const myMultis = await getMultis();` down below. You'd need to mark this function async and I think RES ajax handles de duplicating requests so you can get rid of the early return too.

andytuba: Sorry, it's already a sync, I really should do this from a desktop..

lib/css/modules/_nightMode.scss
dhensche: yeah, I always have night mode on so I noticed the difference immediately :smile:

lib/modules/subredditManager.js
dhensche: so for the `getMultis` function, I was just copying the pattern I saw for `getSubreddits`. I guess I'm unsure if you want a rewrite of the function, move the call to the function to load the multis, or both.

lib/modules/subredditManager.js
```diff
@@ -155,8 +155,8 @@ module.options = {
 };
 
 let shortCutsContainer, $subredditGroupDropdown, subredditGroupDropdownUL, $srList,
-	$editShortcutDialog, deleteButton, $sortMenu, gettingSubreddits, subredditsLastViewed,
-	sortShortcutsButton, shortCutsTrash, myRandomEnabled, myRandomGold;
+	$editShortcutDialog, deleteButton, $sortMenu, gettingSubreddits, gettingMultis,
```
erikdesjardins: We can drop the `gettingMultis` flag

lib/modules/subredditManager.js
```diff
@@ -1072,18 +1074,45 @@ async function getSubreddits() {
...
+	});
+
+	if (multis) {
+		myMultis = organizeMyData(multis, multi => multi.data.display_name);
+	}
+
+	gettingMultis = false;
+}
+
+function organizeMyData(data, dispFn) {
```
erikdesjardins: This can be left in `getSubreddits`: multis don't need to be sorted since we only care about their number; they don't need to be deduped since we're only making one API call

lib/modules/subredditManager.js
```diff
@@ -1072,18 +1074,45 @@ async function getSubreddits() {
...
+	populateSubredditDropdown();
+}
+
+let myMultis = [];
+async function getMultis() {
+	if (gettingMultis) return;
+	gettingMultis = true;
+
+	myMultis = [];
+	const multis = await ajax({
```
erikdesjardins: This `ajax()` call can be inlined into `getMultiCounts`; we can drop everything else in this function

lib/modules/subredditManager.js
erikdesjardins: Yeah we can drop this line

lib/modules/subredditManager.js
```diff
@@ -1263,6 +1292,12 @@ async function populateSubredditDropdown(sortBy = 'subreddit', badJSON) {
 	$srList.append(theBody);
 }
 
+export function getMultiCounts(displayName) {
```
erikdesjardins: This will become async as a result, which will be marginally annoying on line 210 but shouldn't be a huge deal (just make `manageSubreddits` async)

erikdesjardins:
Neat, thanks!

Issue #3499, opened 2016-10-01
Hide multireddit nav menu if unused


andytuba:
If users clicks a multireddit link from the left sidebar several times (5?) before clicking that nav menu, disable the nav menu module.

dhensche:
I'd like to contribute on this, but I'm not 100% sure what the requested change is. Are you saying if a user clicks on a specific multi link from the left sidebar several times (5?) we should no longer show the nav menu that usually appears when they hover over the link?

andytuba:
@dhensche Thanks for offering - check out the work I've done on #3587 and see if you have any suggestions?  I also changed the specifications to "show that menu less quickly/less often, and only disable if the user approves it."

dhensche:
I really like the solution you are working on there. Adding the capability for feature enable/disable to be a bit smarter.

PR #3587, opened 2016-11-05
Automatically throttle/disable unused features

andytuba:
If a feature reports that it presents itself to the user automatically or that the user dismisses it, and does not report the user taking advantage of it, then the Penalty Box eventually throttles or (optionally) disables that feature.

This incidentally includes core enhancements and new options to toggle existing module features.

Future plans, if people complain "why did this feature turn off:"

* in the settings console, flag options as "disabled because in the penalty box"
* add dashboard page highlighting features "in the penalty box" with buttons to re-enable them

And if people complain "why did it turn off *again*":

* when a feature in the penalty box is re-enabled, erase the penalty and turn off monitoring for that feature.  (This is currently only applied if the user opts to "re-enable that feature!" from the "feature disabled" notification.)

Closes: #3499


lib/modules/RESTips.js
```diff
@@ -66,6 +77,7 @@ function prevTip() {
 }
 
 function nextPrevTip(idx) {
+	FeaturePenaltyBox.alterFeaturePenalty(module.moduleID, 'sectionMenu', -20);
```
andytuba: Copypasta

lib/modules/featurePenaltyBox.js
```diff
@@ -0,0 +1,85 @@
+import * as Options from '../core/options';
+import * as Modules from '../core/modules';
+
+export const module = {};
+module.MIN_PENALTY = 1;
```
andytuba: Nah, I want these usable from the outside, like RESTips calculating `MAX_PENALTY / tips.length`

lib/core/options/table.js
```diff
@@ -68,9 +68,22 @@ export function getMatchingValueOrAdd(opaqueId, optionKey, valueIdentifier, hydr
 export function mapValueToObject(option, value) {
 	const object = {};
 
+	let index = 0;
 	for (const [field, val] of zip(option.fields, value)) {
-		object[field.name] = val;
+		Object.defineProperty(object, field.name, {
+			get: getMappedValue.bind(value, index),
+			set: setMappedValue.bind(value, index),
```
andytuba: That's probably a good idea, but this is intellectually stimulating!

lib/modules/featurePenaltyBox.js
andytuba: Oh, that's `export const`

andytuba:
Looks like I should either move the `mapValueToObject` function into some util instead of options.table, or build it deeper into the core.

dhensche:
I may have missed it, but are there any plans to show some sort of notification when a feature gets disabled? I see you have future plans to highlight features that have been disabled due to penalties, but I'm wondering if notifying them when it is disabled might be a good idea. A user may not even be aware that some behavior was a feature and then it just disappears on them? Just a thought

andytuba:
Yes, it would be a good idea to add a notification when a feature is disabled. Although I initially thought it would be annoying, it seems like a much better idea now that the default behavior is throttling and the user has to opt into disabling.

lib/modules/featurePenaltyBox.js
```diff
@@ -0,0 +1,161 @@
...
+	if (!option) {
+		console.warn('Could not find option', moduleID, featureID, optionKey);
+		return;
+	}
+	if (option.type !== 'boolean') {
+		console.warn(`${module.moduleID} could not disable option`, moduleID, featureID, optionKey);
+		return;
+	}
+	const oldValue = option.value;
+	const newValue = !option.value;
```
andytuba: I was hoping to punt on that nicety 😅

lib/modules/featurePenaltyBox.js
```diff
@@ -0,0 +1,161 @@
...
+	return obj;
+}
+
+async function suspendFeature(moduleID, featureID) {
+	if (!module.options.suspendFeatures.value) {
+		return;
+	}
+
+	const featureModule = Modules.get(moduleID);
+	const optionKey = featureID;
```
andytuba: I had originally had two motivations behind `featureID` abstraction:

1. Affects related options, e.g. `hoverDelay` for pop-up menus
1. Could be the whole module
2. Potentially support for other types of options besides booleans.

On reconsidering, 

1. There's typically a main option related options are always dependent on the primary option
2. It's easy to add a top-level option, or add `export const MODULE_FEATUREID = '---'` to support "module as feature"
3. For handling other types of options, we could extend the `penalizedDelay / option.penalizedValue` paradigm to other types, e.g. for enums `options.penalizedValue = [ 'disabled', 'partly-enabled-value', 'mostly-enabled-value', 'fully-enabled' ]` and map the penalty score onto that.

tldr: `s/featureID/optionKey/g`

lib/modules/featurePenaltyBox.js
```diff
@@ -0,0 +1,176 @@
...
+		description: 'resPenaltyBoxSuspendFeaturesDesc',
+	},
+	features: {
+		description: 'resPenaltyBoxFeaturesDesc',
+		type: 'table',
+		advanced: true,
+		addRowText: 'manually register feature',
+		fields: [
+			{
+				name: 'moduleID',
```
erikdesjardins: For compatibility with i18n, we should have a `key` or `index` (or something obviously not user-readable) field used by `mapScalarToObject`

See https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3739/files#r92008640

andytuba: How about [`id`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3587/commits/0f2bc0eeb4c49c64337fc5c4082a73e803115ae1) with a fallback to `name`?

PR #3590, opened 2016-11-05
hosts: Remove Futurism

larsjohnsen:
The API appears to be broken and requests takes a long time to time out / be rejected, causing mysterious blank expandos.


andytuba:
@dansutera, @mverderese do you know what's up with futurism.co's API endpoint?

jetlej:
@andytuba - We took it offline briefly during a caching issue, but now it's back up

PR #3591, opened 2016-11-05
Account Switcher Overflow

jontonsoup4:
# Description

Currently, whenever a user has a large number of accounts, the overflow div becomes the height of all of those accounts, causing the list expand beyond never ending reddit. This PR fixes the issue by adding overflow to the account switcher div. 

## Before

screen shot 2016-11-05 at 5 02 43 pm

## After

screen shot 2016-11-05 at 5 01 59 pm

### DIFF
--- a/lib/css/modules/_accountSwitcher.scss
+++ b/lib/css/modules/_accountSwitcher.scss
@@ -1,8 +1,10 @@
 #RESAccountSwitcherDropdown {
-	min-width: 110px;
-	width: auto;
 	display: none;
+	height: 400px;
+	min-width: 110px;
+	overflow: scroll;
 	position: absolute;
+	width: auto;
 	z-index: 1000;
 
 	li {
--- a/lib/css/modules/_accountSwitcher.scss
+++ b/lib/css/modules/_accountSwitcher.scss
@@ -1,8 +1,9 @@
 #RESAccountSwitcherDropdown {
 	display: none;
-	height: calc(100vh - 80px);
+	max-height: calc(100vh - 80px);
 	min-width: 110px;
-	overflow: scroll;
+	overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


lib/css/modules/_accountSwitcher.scss
```diff
@@ -1,8 +1,10 @@
 #RESAccountSwitcherDropdown {
-	min-width: 110px;
-	width: auto;
 	display: none;
+	height: 400px;
```
larsjohnsen: If this is replaced by `height: calc(100vh - 80px)` , then you avoid unnecessarily limiting the height of the account switcher.

lib/css/modules/_accountSwitcher.scss
```diff
@@ -1,8 +1,10 @@
 #RESAccountSwitcherDropdown {
-	min-width: 110px;
-	width: auto;
```
erikdesjardins: Nit: don't move these lines if you aren't changing them

lib/css/modules/_accountSwitcher.scss
```diff
@@ -1,8 +1,10 @@
...
 	display: none;
+	height: calc(100vh - 80px);
```
erikdesjardins: This forces the height to always be `calc(100vh - 80px)`. It should use `max-height`.

lib/css/modules/_accountSwitcher.scss
```diff
@@ -1,8 +1,10 @@
...
+	min-width: 110px;
+	overflow: scroll;
```
erikdesjardins: This forces scrollbars to always display. It should be `auto`, or, for completeness:

```css
overflow-x: hidden;
overflow-y: auto;
```

PR #3592, opened 2016-11-06
Clean-up RESTips, featureTips guiders (for filterline)

larsjohnsen:
FeatureTips are multi-step guiders which may (given they're configured to do so) show guiders the first time the feature appears.

- Refactors RESTips so that only the one to be shown is generated at initialization (this improves the first-run start-up time with > 1s)
- The first guider "About RES" attaches (as intended to) to the settings cog
- Guiders for Filterline

### DIFF
--- a/lib/modules/RESTips.js
+++ b/lib/modules/RESTips.js
@@ -29,6 +29,15 @@ module.options = {
 		value: true,
 		description: 'RESTipsDailyTipDesc',
 	},
+	newFeatureTips: {
+		type: 'boolean',
+		value: true,
+		description: 'Show tips about new features when they appear for the first time.',
+	},
+};
+
+module.loadDynamicOptions = () => {
+	addFeatureTips();
 };
 
 const lastTooltipStorage = Storage.wrap('RESLastToolTip', 0);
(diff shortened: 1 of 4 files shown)


lib/utils/dom.js
```diff
@@ -35,14 +35,17 @@ export function observe(ele, options, callback) {
...
+		if (
+			(subtree && $(ele).find(selector).length) ||
+			(!subtree && Array.from(ele.children).some(child => $(child).is(selector)))
+		) {
 			resolve();
 			return;
 		}
 
-		const observer = observe(ele, { childList: true }, mutation => {
+		const observer = observe(ele, { childList: true, subtree }, mutation => {
```
larsjohnsen: This is probably to expensive (especially if there's many new features). I'm looking for an alternative.

lib/modules/RESTips.js
```diff
@@ -247,83 +296,47 @@ const tips = [{
...
-}];
+},
+...Object.values(featureTips),
+];
 
 const initTips = _.once(() => {
 	$('body').on('click', '#disableDailyTipsCheckbox', e => Options.set(module, 'dailyTip', e.target.checked));
-
-	// create the special "you have never visited the console" guider...
-	createGuider(0, 'console');
```
larsjohnsen: Could find anywhere where this was actually run, so I've removed everything relating to `consoleTip`

lib/modules/RESTips.js
```diff
@@ -211,142 +281,106 @@ const tips = [{
...
-	}
-
-	$description.on('click', 'a', () => { PenaltyBox.alterFeaturePenalty(module.moduleID, 'sectionMenu', -20); });
-
-	guiders.createGuider(guiderObj);
-}
-
-function handleEscapeKey(event: KeyboardEvent) {
-	if (event.which === 27) {
-		PenaltyBox.alterFeaturePenalty(module.moduleID, 'sectionMenu', 10);
```
larsjohnsen: @andytuba `handleEscapeKey` has no equivalent in the refactor, since close is always handled by the guider library (via `onClose`), so this penalty event is removed. However, the `onClose`'s penalty is `15`, so I'm wondering if I have overlooked something.

andytuba: Whoops, I think that is copypasta that didn't get tested and tidied up before it got merged.  `sectionMenu` isn't a thing in `RESTips`..

andytuba: Yeah, it looks like it was appropriate to trash this since `onClose` will handle it, and `sectionMenu` should be replaced with `dailyTip` I guess..

... or maybe we should strip out the penaltybox handling in this module.

PR #3600, opened 2016-11-09
floater and pageNavigator: Use IntersectionObserver instead of checking scrollY

larsjohnsen:
[IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) provides a cheaper way to check whether things should be shown. Note that it is available only in Chrome and Firefox Nightly, so a polyfill is used elsewhere. The old approach checks `scrollY`, which usually causes layout trashing and janking.

Changes tested on Edge, Firefox 52 and Chrome.

Resolves https://www.reddit.com/r/RESissues/comments/5bqca2/bug_initial_scroll_lag_after_comment_page_load/

### DIFF
--- a/firefox/package.json
+++ b/firefox/package.json
@@ -9,7 +9,7 @@
 	"version": "{{prop?version!../package.json}}",
 	"main": "{{./background.entry.js}}",
 	"engines": {
-		"firefox": ">=48.0"
+		"firefox": ">=49.0"
 	},
 	"permissions": {
 		"private-browsing": true,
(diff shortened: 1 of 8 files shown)


erikdesjardins:
We could probably use https://www.npmjs.com/package/intersection-observer; import it in vendor/index.js or something (it's a noop if `'IntersectionObserver' in window`)

larsjohnsen:
@erikdesjardins I considered that, but since it is polling-based, using it would likely be more expensive and slower.

erikdesjardins:
You were likely looking at a different one, this one doesn't poll by default, just uses scroll/resize listeners and a MutationObserver...either way is fine though

larsjohnsen:
The polyfill appears to work fine in Edge/Firefox.

PR #3602, opened 2016-11-10
Add minimum comment count to enable custom comment depth

### LINKED ISSUE #3456: Option to only apply comment depth if # comments > threshold
threshold default of .. 50?
customizable per subreddit
option (enabled) to always apply depth if # comments not immediately available

dhensche:
Resolves #3456

### DIFF
--- a/lib/modules/commentDepth.js
+++ b/lib/modules/commentDepth.js
@@ -21,7 +21,8 @@ module.options = {
 	defaultMinimumComments: {
 		type: 'text',
 		value: '50',
-		description: 'Default minimum number of comments required to apply custom depth',
+		description: 'commentDepthDefaultMinimumCommentsDesc',
+		title: 'commentDepthDefaultMinimumCommentsTitle',
 	},
 	commentPermalinks: {
 		type: 'boolean',
@@ -48,7 +49,7 @@ module.options = {
 			type: 'text',
 			value: '4',
 		}, {
-			name: 'minimumComments',
+			name: 'commentDepthMinimumComments',
 			type: 'text',
 			value: '50',
 		}],
--- a/lib/modules/commentDepth.js
+++ b/lib/modules/commentDepth.js
@@ -14,13 +14,13 @@ module.description = 'commentDepthDesc';
 module.options = {
 	defaultCommentDepth: {
 		type: 'text',
-		value: '4',
+		value: 4,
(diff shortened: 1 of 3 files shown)


lib/modules/commentDepth.js
```diff
@@ -29,7 +34,7 @@ module.options = {
 		value: false,
 		description: 'Set depth on links to particular comments with context.',
 	},
-	subredditCommentDepths: {
+	subredditCommentDepthConfigurations: {
```
dhensche: Totally forgot about possible option migration issues when doing this rename

lib/modules/commentDepth.js
```diff
@@ -39,9 +44,14 @@ module.options = {
 		}, {
 			name: 'commentDepth',
 			type: 'text',
+			value: '4',
```
dhensche: yeah, I can update all the fields to be a plain number if desired

erikdesjardins: I think all text option default values should be strings (until we have a number type) to emphasize that you should use `parseInt`/`parseFloat` since it _could_ be saved as a string

lib/modules/commentDepth.js
```diff
@@ -39,9 +44,14 @@ module.options = {
...
+		}, {
+			name: 'minimumComments',
+			type: 'text',
+			value: '50',
```
erikdesjardins: In a larger module I'd say no because it might be easy to miss, but this module is so small that it's not a big deal

andytuba: Yeah, it's a smallish module and I don't expect it to scope creep too much.

As an academic exercise, I would implement it like:

```javascript
module.loadDynamicOptions = function() {
    module.options.subredditDepthsWhateverItsCalled.fields[0].value = module.options.defaultDepth.value;
    module.options.subredditDepthsWhateverItsCalled.fields[1].value = module.options.defaultMinimum.value;
};

```

andytuba:
Somebody give it a spin and I'd say merge in?

dhensche:
@erikdesjardins left a comment about leaving the numeric text option default values as strings right before I committed the change to plain numbers. @andytuba should I revert those back to strings before merging in?

andytuba:
ah yeah, let's go back to strings, sorry for the waffling!

PR #3604, opened 2016-11-11
Transform standard filters into 'engine' filters

### LINKED ISSUE #2839: Use the customFilters "engine" to run regular filters
Having two different code paths to filter by subreddit, keywords, etc. is redundant.

dhensche:
This is a first cut at removing the custom filtering logic
for the basic filters. There is still a transformation phase
that converts the value input from the user into different
filters used by the custom filter engine (group, currentSub, etc.).
These transformations are cached and then run just like the custom
filters during the filtering phase

fixes #2839

### DIFF
--- a/lib/modules/filteReddit.js
+++ b/lib/modules/filteReddit.js
@@ -434,11 +434,11 @@ module.go = () => {
 const createFilterline = _.once(() => {
 	const filterline = new Filterline();
 
-	function buildBasicFilter(key, customFilterType, { forcePartialMatch = false }: { forcePartialMatch?: boolean } = {}) {
+	function buildBasicFilter(key, customFilterType, { fullMatch = false }: { fullMatch?: boolean } = {}) {
 		return {
 			key,
 			evaluate(thing) {
-				const { filters, sources } = getStringMatchFilters(key, customFilterType, forcePartialMatch);
+				const { filters, sources } = getStringMatchFilters(key, customFilterType, fullMatch);
 				return executeFilters(thing, filters, sources);
 			},
 			clearCache() { getStringMatchFilters.cache.delete(key); },
@@ -447,9 +447,9 @@ const createFilterline = _.once(() => {
 
 	const externalFilters = [
 		buildBasicFilter('keywords', 'postTitle'),
-		buildBasicFilter('domains', 'domain', { forcePartialMatch: true }),
+		buildBasicFilter('domains', 'domain'),
 		{
-			...buildBasicFilter('subreddits', 'subreddit'),
+			...buildBasicFilter('subreddits', 'subreddit', { fullMatch: true }),
 			state: (
 				module.options.filterSubredditsFrom.value === 'everywhere' ||
 				module.options.filterSubredditsFrom.value === 'everywhere-except-subreddit' && !currentSubreddit() ||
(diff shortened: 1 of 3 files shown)


lib/modules/filteReddit.js
```diff
@@ -493,74 +488,60 @@ function executeCustomFilters(thing) {
 	return filters.find(filter => config[filter.body.type].evaluate(thing, filter.body, config));
 }
 
-const regexRegex = /^\/(.*)\/([gim]+)?$/;
```
dhensche: This change is going to break any of the basic filters that use a regex. Might need to refactor how the post filters work to accept regex flags in order to fully implement this change

lib/modules/filteReddit.js
```diff
@@ -493,74 +488,60 @@ function executeCustomFilters(thing) {
...
 		const [matchString, applyTo = 'everywhere', applyList = '', except = ''] = Array.isArray(source) ? source : [source];
+		const matcher = module.options.regexpFilters.value ? matchString : matchString;
+		const filter = {
+			type: 'group',
+			op: 'all',
+			of: [],
+		};
 
-		let whenMatching;
-		if (module.options.regexpFilters.value && regexRegex.test(matchString)) {
```
dhensche: Probably need to keep some of this block around as well to handle regexes

lib/modules/filteReddit.js
```diff
@@ -494,25 +489,31 @@ function executeCustomFilters(thing) {
 }
 
 const regexRegex = /^\/(.*)\/([gim]+)?$/;
-
 const getStringMatchFilters = _.memoize(type => {
 	const sources = module.options[type].value;
+	const filterMap = {
```
dhensche: do you mean adding the filter type (postTitle, linkFlair, etc.) to the option itself like this?

```javascript
keywords: {
    type: 'table',
    addRowText: '+add filter',
    filter: 'postTitle',... // like this for example
```

lib/modules/filteReddit.js
```diff
@@ -382,27 +381,23 @@ function createFilterline() {
 	const filterline = new Filterline();
 
 	function addExternalFilters() {
+		function buildBasicFilter(key) {
```
larsjohnsen: Add a `mapping` parameter here …

lib/modules/filteReddit.js
```diff
@@ -382,27 +381,23 @@ function createFilterline() {
...
+			return {
+				key,
+				evaluate(thing) { return filtersMatchString(key, thing); },
```
larsjohnsen: … and change this to 
```
evaluate(thing) { 
  const filters = getStringMatchFilters(key, mapping);
  return filtersMatchString(filters, thing);
}
```
…

lib/modules/filteReddit.js
```diff
@@ -522,45 +523,55 @@ const getStringMatchFilters = _.memoize(type => {
...
-
-		if (filter.except.length && compareString.includes(filter.except)) return false;
-
-		if (filter.whenMatching instanceof RegExp) return filter.whenMatching.test(compareString);
-		else if (matchFullString) return compareString === filter.whenMatching;
-		else return compareString.includes(filter.whenMatching);
-	});
-
-	return result && result.source;
+function filtersMatchString(type, thing) {
```
larsjohnsen: … and replace `type` with `filters`.

erikdesjardins:
Thanks!

PR #3606, opened 2016-11-12
New feature: Browse by wheel

larsjohnsen:
Inserts a little widget into the floater, which when scrolled upon will selected the next/previous entry or gallery piece.

image

When the selected entry has an expanded gallery:
image

When scrolling in the red-ish area, next/prev will act on the gallery. The area will disappear after reaching the end of the gallery, so the next scroll will select the next entry.

### DIFF
--- a/lib/css/modules/_wheelBrowse.css
+++ b/lib/css/modules/_wheelBrowse.css
@@ -1,0 +1,18 @@
+.res-wheel-media-browse {
+	position: relative;
+	width: 30px;
+	height: 16px;
+	margin-left: 4px;
+	border-radius: 5px;
+	cursor: crosshair;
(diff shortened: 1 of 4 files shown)


larsjohnsen:
I am a little uncertain how well this works with touchpads. AFAIK Chrome emits the non-standard `mousewheel` event, while Firefox does not distinguish touchpad and mouse in that regard. I don't think Edge emits any.

As for intrusiveness, I guess that anyone that find it sufficiently annoying will click on it. That will bring them to the settings console, which will allow them to disable it. Alternatively, a featureGuider (cf. #3592) could include a button "don't show it" which disables it.

erikdesjardins:
Looks great, thanks!

PR #3607, opened 2016-11-12
showImages: Set mediaMaxSize as internal style

larsjohnsen:
Avoids trashing in order to determine viewport(Width|Height).


erikdesjardins:
Neat, thanks!

PR #3611, opened 2016-11-13
Housekeeping

wojtekmaj:
* ~~Define module.options in one expression~~
* Move jquery-tokeninout to node modules
* Update all eslint-related dependencies
* Remove double spaces, unnecessary empty lines
* Fixed mixed spaces and tabs


wojtekmaj:
Interestingly, we've started getting no-useless-escape errors with this update, and some of them look like we could have wrong RegExes.

For instance, submitHelper.js:

````
	const stripUrlRe = /^(?:https?:\/\/)?(?:(?:www|i|m)\.)?(.+?)\/?(?:\.\w+)?(?:#[^\/]*)?$/i;
                            was this supposed to be backslash or escape character? ^
````

File: lib/modules/about.js [hunk 1]
```diff
@@ -10,88 +10,77 @@ module.sort = -10;
 module.alwaysEnabled = true;
 module.description = aboutTemplate();
 
-module.options = {};
```
erikdesjardins: Moving these into a single object will be taken care of by https://github.com/erikdesjardins/Reddit-Enhancement-Suite/commit/348913b062a16ceeff043b62cae5c5e8c751875a#diff-9a6bec128b40a3ba6580d95a3a4f54b0 (which also includes other changes to those options in some places) so let's leave this out for now to avoid causing conflicts. It will also conflict with i18n'ing the option descriptions

package.json
```diff
@@ -61,6 +61,7 @@
     "favico.js": "0.3.10",
     "jquery": "3.1.1",
     "jquery-sortable": "0.9.13",
+    "jquery-tokeninput": "^1.6.0",
```
erikdesjardins: Use an exact version like all the other dependencies

package.json
```diff
@@ -70,7 +71,7 @@
     "autoprefixer": "6.5.1",
     "ava": "0.15.2",
     "babel-core": "6.17.0",
-    "babel-eslint": "7.0.0",
+    "babel-eslint": "^7.1.0",
```
erikdesjardins: Exact version

package.json
```diff
@@ -93,17 +94,17 @@
...
-    "eslint-plugin-ava": "3.1.1",
-    "eslint-plugin-babel": "3.3.0",
-    "eslint-plugin-consistent-return-legacy": "1.0.1",
-    "eslint-plugin-dollar-sign": "1.0.0",
-    "eslint-plugin-filenames": "1.1.0",
-    "eslint-plugin-import": "2.0.1",
-    "eslint-plugin-no-useless-assign": "1.0.2",
-    "eslint-plugin-prefer-bind-operator": "0.0.4",
-    "eslint-plugin-prefer-spread": "1.0.3",
+    "eslint": "^3.10.0",
```
erikdesjardins: Use an exact version; most of these aren't even updated

erikdesjardins:
That regex (probably all of them) work fine. It's escaped because in general you _do_ need to escape slashes, just not inside character classes. Same with `.`, `+`, etc.

So just remove the backslashes from those (or disable the rule).

erikdesjardins:
Please rebase to get rid of the merge commits.

wojtekmaj:
I messed up, will create new one

PR #3612, opened 2016-11-13
Fix graphiq expando

thybag:
It appears the graphiq url's had changed slightly meaning the a lot of the new graphiq urls were no longer expandable. Have added a fix + also added a fallback for when the oembed api doesn't return any data for a url (which appears to happen occasionally)

Tested against: 
https://www.reddit.com/domain/graphiq.com/

### DIFF
--- a/lib/modules/hosts/graphiq.js
+++ b/lib/modules/hosts/graphiq.js
@@ -9,8 +9,7 @@ export default {
 	landingPage: 'https://www.graphiq.com/', // set menually as https url without www didn't work when i tried it
 	detect: ({ pathname }) => (/^\/(?:w|wlp|vlp)\/([A-z0-9]+)/i).exec(pathname),
 	async handleLink(href, [url, id]) {
-
-		const { width, height } = await ajax({
+		let { width, height } = await ajax({
 			url: 'https://oembed.graphiq.com/services/oembed',
 			data: { url },
 			type: 'json',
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/graphiq.js
```diff
@@ -7,15 +7,19 @@ export default {
 	domains: ['graphiq.com'],
 	logo: '//www.graphiq.com/favicon.ico',
 	landingPage: 'https://www.graphiq.com/', // set menually as https url without www didn't work when i tried it
-	detect: ({ pathname }) => (/^\/(?:w|wlp)\/([A-z0-9]+)/i).exec(pathname),
+	detect: ({ pathname }) => (/^\/(?:w|wlp|vlp)\/([A-z0-9]+)/i).exec(pathname),
 	async handleLink(href, [url, id]) {
-		const { width, height } = await ajax({
+		let { width, height } = await ajax({
```
erikdesjardins: The default values can be inline, e.g.

```js
const {
    width = '640',
    height = '360',
} = await ajax({
```

erikdesjardins: I keep forgetting that I can just change this 😅

thybag: Wow, never knew that was possible :o Have updated he PR :)

thybag: Edit: didn't notice you'd already updated it. Will leave as is :p

PR #3626, opened 2016-11-18
commentsOpened filter: newCommentCount criteria [WIP]

larsjohnsen:
Useful in order to hide posts with less than n new comments.

WIP since modifying `trueText`/`falseText` in `parse` is hacky.

### DIFF
--- a/lib/modules/newCommentCount.js
+++ b/lib/modules/newCommentCount.js
@@ -295,7 +295,7 @@ function subscribe(threadid) {
 	drawSubscriptionsTable();
 }
 
-function displayNewCommentCount(thing) {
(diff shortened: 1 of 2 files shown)


lib/modules/newCommentCount.js
```diff
@@ -286,16 +286,19 @@ function subscribe(threadid) {
 	drawSubscriptionsTable();
 }
 
-function displayNewCommentCount(thing) {
+export function getNewCount(thing: Thing): ?number {
```
andytuba: For this purpose, it might be better to surface `export function isPostVisited(thing: Thing): bool` to handle the case where the user has visited the post but there aren't new comments since then.

PR #3627, opened 2016-11-18
Reattach the comment tools and preview after a post/comment has been saved

### LINKED ISSUE #3625: Post style buttons do not appear after second edit.
https://www.reddit.com/r/RESissues/comments/5dltjl/bugedge_trying_to_edit_again_after_saving_first/

Confirmed repro'd on Edge.

dhensche:
This is a PR to make the comment tools and comment preview available when a user edits a comment/post twice on a page without refreshing the page. Currently, when the form is posted the HTML form is completely replaced, which wipes away the attached tools and preview.

fixes #3625

### DIFF
--- a/lib/utils/dom.js
+++ b/lib/utils/dom.js
@@ -38,6 +38,18 @@ export function observe(ele: Node, options: MutationObserverInit, callback: Iter
 	return observer;
 }
 
+export function observeNewThing(ele: Node): Promise<MutationRecord> {
+	return new Promise(resolve => {
+		const observer = observe(ele, { childList: true }, mutation => {
+			if (Array.from(mutation.addedNodes).find(e => e.classList.contains('thing'))) {
+				observer.disconnect();
+				resolve(mutation);
+				return true;
+			}
+		});
+	});
+}
+
 export function waitForChild(ele: Element, selector: string): Promise<void> {
 	return new Promise(resolve => {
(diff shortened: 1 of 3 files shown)


dhensche:
This current solution is gross (sets an interval to check for when the HTML has been replaced after the post and then reattaches the preview and tools). I know there has to be a more RES way to do this but I'm busy at work right now so I didn't have time to investigate ☹️

andytuba:
First thought is a MutationObserver watching for the HTML being replaced.

dhensche:
That's what I was thinking, but saw a bunch of warnings about it not always being available in a browser? Also, my js is a little rusty

lib/utils/dom.js
```diff
@@ -38,6 +38,18 @@ export function observe(ele: Node, options: MutationObserverInit, callback: Iter
 	return observer;
 }
 
+export function observeNewThing(ele: Node): Promise<MutationRecord> {
+	return new Promise(resolve => {
+		const observer = observe(ele, { childList: true }, mutation => {
+			if (Array.from(mutation.addedNodes).find(e => e.classList.contains('thing'))) {
+				observer.disconnect();
+				resolve(mutation);
```
andytuba: How about sending back just the target node? Is there potential benefit in returning the whole mutation record?

dhensche: I'm cool with that. Probably makes the most sense, as the calling code should most likely only need access to the target (Thing container)

PR #3644, opened 2016-11-21
Centralized batching

### LINKED ISSUE #2294: Better handling of watched elements
Mentioned by andytuba [here.](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/1948#issuecomment-68491809)
In many modules we have stuff like [this](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/40c2ced61ec32412c9e4ce9ae8c91d5de394a075/lib/modules/betteReddit.js#L183-L189), which is repetitive and should be simplified.

The solution will also have to take care of batching/chunking.

Perhaps something like:
`RESUtils.watchFor(elementTypes [, pageTypes ], callback)`

### LINKED ISSUE #3009: findAllImages happens before initObservers
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/showImages.js#L352
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/utils/watchers.js#L30
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/core/init.js#L172

Seems that `go` (and within it `findAllImages`) in `showImages` is run before `initObservers`, so elements added in the meanwhile is not captured by the watchers.

A solution could be to run `findAllImages` very early on, at least on linklist pages, so that expandos quickly becomes available, and then later rerun it when the page is ready.

larsjohnsen:
New watcher functions:
 - WatchForThing: Used for enhancement which are Thing-centric
 - WatchForElements (replaces `WatchForElement`): Used when changes doesn't act on Thing

Performance improvements:
- All callbacks are organized and executed based on visibility / distance from viewport on the initial execution, so that only things which can be seen are prioritized.
- KeyboardNavigation move functions are available earlier, and are more responsive

Minor bugfixes:
- SelectedEntry: If several selects occurred during 1 frame, listeners was invoked with the wrong `old` Thing
- scrollType `middle` didn't always scroll when transitioning to/from a collapsed Thing
- userHighlight `highlightFirstCommenter` and `autoColorUsernames` caused severe jank due to chunking
- newCommentCount only indicated that comment page had been opened if there are additional comments (now it only highlights when there are new comments)
- saveComments included `.keyNavAnnotation` and `.expando-button` 

Observers and hideChildComments are refactored.

Resolves #2294
Closes #3009

### DIFF
--- a/lib/modules/menu.js
+++ b/lib/modules/menu.js
@@ -14,11 +14,14 @@ module.hidden = true;
 
 module.beforeLoad = () => {
 	renderConsoleLink();
-	addLegacyStyling();
 };
 
 module.go = () => {
 	addConsoleLink();
+};
+
+module.afterLoad = () => {
+	addLegacyStyling();
 };
 
 let RESPrefsLink;
--- a/lib/modules/menu.js
+++ b/lib/modules/menu.js
@@ -14,15 +14,11 @@ module.hidden = true;
 
 module.beforeLoad = () => {
 	renderConsoleLink();
+	addLegacyStyling();
 };
 
 module.go = () => {
 	addConsoleLink();
-};
-
-
-module.afterLoad = () => {
-	addLegacyStyling();
 };
 
 let RESPrefsLink;
--- a/lib/utils/async.js
+++ b/lib/utils/async.js
@@ -27,10 +27,12 @@ export function waitFor<T>(callback: () => ?T, interval?: number = 1): Promise<T
 export const forEachChunked = (() => {
 	const framerate = 15;
 	const frameTime = 1000 / framerate;
+	const minSlot = frameTime / 2; // In case there's much else going on, don't compromise this too much
 
 	const queues = [];
 
 	const run = frameThrottle((start: number) => {
+		const slotEnd = Math.max(start + frameTime, now() + minSlot);
 		do {
 			_.remove(queues, ({ generator, callback, resolve, reject }): ?boolean => {
 				const { value, done } = generator.next();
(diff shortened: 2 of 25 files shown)


lib/utils/async.js
```diff
@@ -266,6 +268,7 @@ export const nextFrame = (() => {
 /* eslint-disable no-redeclare */
 // rest params can't be placed before non-rest params, so this is unfortunately necessary
 // add another case if you need more params
+declare function frameThrottle<A, B, C>(callback: (a: A, b: B, c: C, start: number) => void): () => void;
```
larsjohnsen: @erikdesjardins I only got this working by putting the case above the others. Is it correct?

erikdesjardins: Flow's intersections are order-dependent (so that it's predictable which case is chosen), so in general it's fine to put a more-specific case first. In this case I'd expect it to work however; maybe it's because the types are missing from the returned function? i.e. it's `() => void` instead of `(a: A, b: B, c: C) => void`.

Don't worry about it though, I'll be merging this next (probably) so I'll take care of it if that's the case.

lib/modules/menu.js
```diff
@@ -14,17 +14,13 @@ module.hidden = true;
 
 module.beforeLoad = () => {
 	renderConsoleLink();
+	addLegacyStyling();
```
erikdesjardins: FYI this needs to be in afterLoad (or at least go) since it's based on subreddit stylesheets, which aren't ready (perhaps not even parsed, if we're fast) in beforeLoad and which could potentially be removed by styleTweaks at some point. I'll change it though, dw about it

erikdesjardins:
Awesome!
ds3buyb

PR #3659, opened 2016-11-23
Use StyleSheet disabled property to disable subreddit styles

erikdesjardins:


erikdesjardins:
Closing as this is, somewhat surprisingly, less effective than removing the element. I'm consistently seeing FOUCs on refresh (~100% of the time) on pages where I _never_ do with the current implementation.

It's very important that we don't regress on performance here.

Issue #2105, opened 2015-03-14
Subreddits whose name is all numbers confuses filteReddit and SubredditManager


andytuba:
specifically /r/195

this is medium difficulty mostly because it entails poking a bunch of different sections of subreddit manager

erikdesjardins:
#3665

PR #3665, opened 2016-11-24
Replace $.fn.data with $.fn.attr for reading data-attributes

### LINKED ISSUE #3662: Don't use jQuery.fn.data to read data-attributes
...because it behaves "cleverly" for `'true'`, `'123'`, etc.

Causing issues related to (but not as reported): https://www.reddit.com/r/Enhancement/comments/5eju98/why_does_filtering_r195_make_reddit_not_load_at/
Namely: you can't filter /r/195 from the subreddit info hover

erikdesjardins:
Fixes #3662


larsjohnsen:
Also fixes #2105

PR #3669, opened 2016-11-25
Filterline related fixes / improvements

larsjohnsen:


erikdesjardins:
Some [stuff I'm working on](https://github.com/erikdesjardins/Reddit-Enhancement-Suite/compare/d544677a4abaf34cb0a300e0dc0e35eb41db8161...403c963536ab9766694f120c0d63b0c8c21ea02d) probably obviates the need for the last commit.

After my changes, if I understand correctly, queuing would only affect async filters, which I think is unimportant enough (async filters are already slow, and an extra call to `isURLVisited` per thing isn't a big deal IMO).

erikdesjardins:
You could keep the changes to `getUpdateRange`->`getFiltersToTest` though, they look good

erikdesjardins:
Looks good, thanks!

PR #3671, opened 2016-11-26
showImage: Reuse common move/resize functions for iframes

larsjohnsen:
Compared to previous solution, this accepts keyboard navigation and moving the iframe.


erikdesjardins:
Great improvement, thanks!

PR #3675, opened 2016-11-29
Fast-path for synchronous filters

erikdesjardins:


erikdesjardins:
Appveyor failure appears spurious

PR #3680, opened 2016-11-29
Filter init optimization

erikdesjardins:
Depends on #3679

Now, regardless of your combination of external filters and filterline filters (and as long as storage isn't _really_ slow), posts will only be processed once on pageload. (Previously they would be processed once by external filters and once again by filterline filters in the next tick)

### DIFF
--- a/lib/modules/filteReddit.js
+++ b/lib/modules/filteReddit.js
@@ -334,6 +334,9 @@ module.loadDynamicOptions = () => {
 module.beforeLoad = async () => {
 	updateNsfwBodyClass(module.options.NSFWfilter.value);
 
+	// start initializing filterline early
+	createFilterline();
+
 	if (await initialFilterlineVisibility().get()) {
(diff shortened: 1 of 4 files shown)


lib/modules/filteReddit.js
```diff
@@ -311,32 +315,29 @@ module.shouldRun = () => !(
...
+const initialFilterlineVisibility = _.once(() => reifyPromise(asyncSome([
+	filterlineVisibilityStorage.get(),
+	Init.loadOptions.then(() => module.options.showFilterline.value),
+], x => x)));
 
+module.loadDynamicOptions = () => {
 	// Initiate async load now so state will be restored asap
-	filterlineState = Storage.get(filterlineKey);
-
-	module.options[filterlineVisibilityOptionKey] = {
```
erikdesjardins: I shouldn't have suggested this, it doesn't work well when we need multiple keys.

Just loading it with `Storage` in `loadDynamicOptions` is fast enough

PR #3686, opened 2016-12-01
Fix for already submitted notification display wrong time

corylulu:
Might need to be checked over, but as is, it's showing me that a post submitted an hour ago as being 7 hours old (I'm PST time). Looking at the JSON, it appears that the create_utc actually gives the local time.


erikdesjardins:
Thanks!

PR #3691, opened 2016-12-01
Don't stage options that haven't changed

erikdesjardins:
So all #3690 callbacks in a module don't fire every time you save.


andytuba:
:confetti: :ok_hand:

PR #3694, opened 2016-12-01
Video captions

erikdesjardins:
https://www.reddit.com/r/RESissues/comments/5fx8vd/captions_not_showing_up_in_imgur_gallery/

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1848,6 +1848,7 @@ function videoAdvanced(options) {
 	element.appendChild(player);
 
 	const vid: HTMLVideoElement = (player.querySelector('video'): any);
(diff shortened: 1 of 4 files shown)


lib/modules/showImages.js
```diff
@@ -2029,9 +2030,9 @@ function videoAdvanced(options) {
 	setMediaMaxSize(vid);
 	makeMediaZoomable(vid);
 	setMediaControls(vid, undefined, options.sources[0].source);
-	makeMediaMovable(player);
-	keepMediaVisible(player);
-	makeMediaIndependentOnResize(element, player);
+	makeMediaMovable(container);
```
erikdesjardins: @larsjohnsen is this correct? (so it doesn't move the captions as well)

larsjohnsen: Looks good.

PR #3703, opened 2016-12-04
Replace filterline "for all" icon with "filter" (upside down beaker)

andytuba:
screen shot 2016-12-04 at 1 12 53 pm

Technically it's an upside-down beaker instead of filter/funnel https://thenounproject.com/term/filter/101890/

Would the "equalizer" Batch icon be more appropriate? a la https://thenounproject.com/term/filter/103807/

### DIFF
--- a/lib/css/modules/_filteReddit.scss
+++ b/lib/css/modules/_filteReddit.scss
@@ -47,6 +47,17 @@ body.hideOver18 {
 	}
 }
 
+.res-filterline-icon {
+	display: inline-block;
+	transform: rotate(180deg);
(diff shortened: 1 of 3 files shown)


lib/css/modules/_filteReddit.scss
```diff
@@ -47,6 +47,17 @@ body.hideOver18 {
 	}
 }
 
+.res-filterline-icon {
```
erikdesjardins: If this selector doesn't match anything, you should use `%res-filterline-icon`

PR #3708, opened 2016-12-05
subredditManager: Do not await manageSubreddits

larsjohnsen:
When the `getMultiCounts` request is not cached, it will have to have to wait for Reddit to respond on the API request, which may take several seconds and holds up the `go` phase.


erikdesjardins:
Holding up the completion of `go` is not so important because `afterLoad` is generally unimportant...but we don't want it to take many seconds, I suppose.

PR #3709, opened 2016-12-05
Support `staticflickr` URLs

alex-seville:
Add support for properly embedding and crediting `staticflickr` URLs.

Right now when people link directly to the CDN copy of a flickr image (i.e. `http://c1.staticflickr.com/9/8345/8186873856_5d36a04f3e_k.jpg`) it doesn't pass the `detect` regexes.  

This code allows the photo oembed data to be retrieved using the photoid embedded within the `staticflickr` URL.  This ensures the image is shown and that the credit is preserved.


alex-seville:
Hmm, actually that endpoint doesn't work with `noembed`. Let me find another way.

PR #3711, opened 2016-12-05
Eroshare host file updated to exclude subdomains.

RandyWild:
Thank you for the suggestions on the previous pull request.


erikdesjardins:
Thanks!

PR #3724, opened 2016-12-05
Move jQuery plugin tokeninput to package.json

wojtekmaj:


erikdesjardins:
FYI I rebased and force-pushed the branch.

Though I noticed that it seems to behave slightly differently than the version in-tree (1.6.1, apparently), so we'll have to do something about that. Possibly I'll publish it on npm under `jquery.tokeninput`

PR #3738, opened 2016-12-06
Emphasize RES should be run from /dist/chrome

roshkins:
For development of the chrome extension. I loaded the wrong folder and got a hard-to-comprehend error.


erikdesjardins:
Thanks!

PR #3740, opened 2016-12-07
use more translated strings

knorssman:
mostly switched descriptions of settings console options to use translated strings


erikdesjardins:
Awesome, thanks!

FYI I set up #3755 so we don't accidentally work on the same thing; just drop a comment in there if you want to work on anything else.

PR #3744, opened 2016-12-07
showImages: Don't force updateRevealedImages on refresh

larsjohnsen:
Running it ignores the possibility that expandos are manually closed.


erikdesjardins:
This fixes reports of "loading a new NER page with 'show all images' reopens all images"?

larsjohnsen:
Correct

PR #3763, opened 2016-12-10
Fix a typo in betteRedditShowTimestampPostsDesc key

wojtekmaj:
That could prevent someone from successfully finding this key later.


erikdesjardins:
Thanks!

Issue #3622, opened 2016-11-16
Notify if subscribed post is edited


andytuba:
If a user has subscribed to new comment notifications on a post, and the post has been edited since the last visit to that page, then a notification should show. This should be optional and probably off by default. Is it worth making this optional by post?

https://www.reddit.com/r/Enhancement/comments/5dc290/does_res_notify_you_if_you_are_subscribed_to_a/

Propheis:
Is this issue closed?

erikdesjardins:
Yep, thanks

PR #3764, opened 2016-12-10
Notify if subscribed post is edited (#3622)

preethamrn:
If a user has subscribed to new comment notifications on a post, and the post has been edited since the last visit to that page, then a notification should show. This is optional and off by default.


preethamrn:
Sure. I've made most of those changes but I'm not sure how to get a Thing object in the updateCommentCount function.

PR #3772, opened 2016-12-12
Use lodash-webpack-plugin

erikdesjardins:
~~This doesn't seem to improve parse or runtime performance (though it does reduce bundle size). -- needs retesting in Webpack 2~~

Reduces bundle size by ~100k and speeds up init by a few frames (a few tens of ms on my machine)

### DIFF
--- a/webpack.config.babel.js
+++ b/webpack.config.babel.js
@@ -3,6 +3,7 @@
 import path from 'path';
 
 import InertEntryPlugin from 'inert-entry-webpack-plugin';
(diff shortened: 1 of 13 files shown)


webpack.config.babel.js
```diff
@@ -161,6 +162,10 @@ export default (env = {}) => {
 					},
 				}),
 				new InertEntryPlugin(),
+				new LodashModuleReplacementPlugin({
+					collections: true,
+					unicode: true, // for _.startCase todo remove when 0.11.1 is released
```
erikdesjardins: Note to self: todo

PR #3773, opened 2016-12-12
[experimental] parallelize build

erikdesjardins:


erikdesjardins:
This makes minimal difference on my machine, even with 8 threads, it's not worth the complexity ATM

PR #3774, opened 2016-12-12
Webpack 2

erikdesjardins:
Will not be merged until Webpack 2 is stable (though it's been in beta a _long_ time).

~~*Note to self**: make sure cheap sourcemaps are fixed~~ (as long as native imports are not used)


wojtekmaj:
@erikdesjardins, how come that you're ending up with manifest.json files without properly "loading" json files? I can't see how these files are being processed, and if I'm trying to do this myself in my test project, I'm ending up with manifest.json file being corrupted webpack-generated JavaScript file. I need to write `entry: \`file-loader?name=[name].[ext]!extricate-loader!interpolate-loader!./${entry}\`,` to get it working, although in that case I'm getting an error that multiple assets emit to the same filename 'manifest.json'... Teach me master :O

erikdesjardins:
Yep, we use [inert-entry-webpack-plugin](https://github.com/erikdesjardins/inert-entry-webpack-plugin) to do that.

It effectively does the same thing as you were trying with `file-loader?name=[name].[ext] ...rest of loaders`, but it also removes the original asset, which should avoid the error about emitting to the same filename.

mc10:
It's [finally stable](https://github.com/webpack/webpack/releases/tag/v2.2.0)!

erikdesjardins:
Finally! Now I just need to fix some issues in `InertEntryPlugin` (it breaks in watch mode, strangely) and this'll be good.

Though I think we should probably wait until after next release to merge this so we aren't making big build system changes right before it

erikdesjardins:
Webpack 2 significantly changes VM startup behaviour by emitting parens around each module (taken as a hint to full-parse up front rather than pre-parsing), so it may regress startup performance; I will revisit/fix this shortly.

I am intentionally not removing the Babel commonjs plugin, because Webpack's handling is unnecessarily verbose with no advantage (no tree-shaking unless we minify, which we can't really do); this will also be revisited.

PR #3776, opened 2016-12-12
[i18n] Option titles, module/option keys, search results

wojtekmaj:
As the translation goes forward, it will be hard for our community to discuss modules and options without something that will be common between languages.

As of this PR, we now can translate option titles.
Because of this, this PR also makes module and option keys way more visible by adding smaller, gray text next to each option name, similar to how it was always added next to title in search results.

Looks like so:

image


andytuba:
> this PR also makes module and option keys way more visible by adding smaller, gray text next to each option name, similar to how it was always added next to title in search results.
👍 👍 👍

lib/modules/search.js
```diff
@@ -166,20 +166,22 @@ function makeOptionSearchResult(moduleKey, optionKey) {
 	const mod = Modules.get(moduleKey);
 	const option = mod.options[optionKey];
 
+	const optionName = i18n(option.title) || _.startCase(option.title || optionKey);
```
andytuba: Could you also add i18n to [rankOption's "get option title"](https://github.com/wojtekmaj/Reddit-Enhancement-Suite/blob/687161bdd5c554f70d7a5f1f0a62baf7f9cc48d3/lib/modules/search.js#L135)

wojtekmaj:
BTW there's `result.breadcrumb = ` twice in search.js, it seems to be some obsolete piece of code, would you like me to remove it while we're at it?

erikdesjardins:
Thanks!

erikdesjardins:
Ah I didn't see that comment, yeah it looks like it's completely unused, might as well get rid of it

erikdesjardins:
FYI I just removed it in 3021af41c0c62eae883fdfb4ea737ba95f0cb0e1

PR #3779, opened 2016-12-12
[i18n] Add toggle on/off to i18n strings

wojtekmaj:


lib/utils/createElement.js
```diff
@@ -9,8 +10,8 @@ export function toggleButton(
 	onClick?: (enabling: boolean) => void = () => {},
 	fieldID: string,
 	enabled?: boolean = false,
-	onText?: string | HTMLElement = 'on',
-	offText?: string | HTMLElement = 'off',
+	onText?: string | HTMLElement = i18n('toggleOn'),
```
andytuba: I feel like the `i18n` should be applied inside the function to `.onText` and `.offText`, instead of only to the default value. Then we can put simply i18n keys into the `onText` and `offText` properties in the code and rely on them being magically translated as late as possible.

erikdesjardins: This is all run synchronously so I think I prefer it this way--if callers who want to provide different text are forced to call `i18n` there will be less implicit behaviour, and unlike module descriptions, etc. there won't be any extra `i18n` calls.

wojtekmaj: The question is whether we even need onText and offText, they are not overwritten anywhere in the code, it seems no feature in the nearest future could use it. Personally, I'd just remove them and call i18n on toggleOn directly in rendering function `$(<span>....)`.

erikdesjardins: They're used here https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/80e5b3403306cc179dc4d33c3dcba4a558943471/lib/modules/nightMode.js#L305

PR #3783, opened 2016-12-13
easterEgg: Add types to a few things

mc10:
Still getting a hang of the typing system.


erikdesjardins:
FWIW I'm not a big fan of changes like these since they don't really increase type safety and they make refactoring (if we change any of the types) marginally harder. Flow's type inference is very powerful, it doesn't need much help.

mc10:
Ahh, I didn't know that. I've worked with TypeScript before, but generally it needs a bit more help specifying types everywhere.

PR #3788, opened 2016-12-14
Use chocolatey to install yarn on Appveyor

erikdesjardins:


erikdesjardins:
This forcibly installs the newest Node version, which we may not always want. Not feasible at the moment.

Eventually Appveyor will ship a newer version of Yarn and we won't need to install it at all.

PR #3791, opened 2016-12-15
Notify if subscribed post is edited

preethamrn:
Requested in Issue #3622 
If a user has subscribed to new comment notifications on a post, and the post has been edited since the last visit to that page, then a notification should show. This is optional and off by default.

NOTE: I've already pull requested this (#3764) but due to some issues with my organization, I'm doing it again.

### DIFF
--- a/lib/modules/newCommentCount.js
+++ b/lib/modules/newCommentCount.js
@@ -46,6 +46,12 @@ module.options = {
 		description: 'newCommentCountShowSubscribeButtonDesc',
 		title: 'newCommentCountShowSubscribeButtonTitle',
 	},
+	notifyEditedPosts: {
+		type: 'boolean',
+		value: false,
+		description: 'Notify if a subscribed post is edited.',
+		advanced: true,
+	},
 	monitorPostsVisited: {
 		type: 'boolean',
 		value: true,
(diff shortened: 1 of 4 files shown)


andytuba:
> NOTE: I've already pull requested this (#3764) but due to some issues with my organization, I'm doing it again.

:+1: on submitting from a named branch instead of master.

andytuba:
Re: [getting a Thing object in the newCommentCount function](https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/3791)

You can call `const post = new Thing(document.querySelector('#siteTable .thing')` to get a `Thing` for the original post inside of [`updateCommentCount`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/66ced58cf2e22ebb980758584b1483bce0474fdc/lib/modules/newCommentCount.js#L390).  While you're in there, could you also change [`thisCount`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/66ced58cf2e22ebb980758584b1483bce0474fdc/lib/modules/newCommentCount.js#L383) to use `post.getCommentsLink()`?

preethamrn:
Any reason you want to get `post.getCommentsLink()` instead of `post.getCommentCount()`? It seems like the latter would be easier since we're using `thisCount` to get the `currentCommentCount` anyway.

andytuba:
`.getCommentCount` sounds great, I just overlooked it.

erikdesjardins:
Sorry about the delay; thanks!

Issue #3789, opened 2016-12-14
loggedInUser should pull from username link, not text content


andytuba:
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/3d23be61b4f10b85993bb6866dfa2a2069b79c27/lib/utils/user.js#L10

Don't pull the text Content, that can get manipulated by other extensions.  Instead, extract the username from the link href

gardnm4:
@andytuba I think this is closed now since Issue #3792 fixed it.

andytuba:
@gardnm4 This issue can be closed as soon as that PR is merged.

PR #3792, opened 2016-12-15
loggedInUser should pull from username link, not text content

preethamrn:
Requested in Issue #3789 
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/3d23be61b4f10b85993bb6866dfa2a2069b79c27/lib/utils/user.js#L10

Extracts username from the link href instead of textContent.


lib/utils/user.js
```diff
@@ -7,7 +7,9 @@ import { MINUTE, string } from './';
 
 export const loggedInUser = _.once((): string | void => {
 	const userLink = document.querySelector('#header-bottom-right > span.user > a');
-	return userLink && !userLink.classList.contains('login-required') && userLink.textContent || undefined;
+	const words = userLink.href.split('/');
```
andytuba: There's actually a handy regex you can use to [extract username from `userLink.href.pathname`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/37511141d7e5571d837df826643f47487d5dc0d9/lib/utils/location.js#L55), sorry I didn't mention it in the original ticket.

preethamrn: Ahh. That simplifies things. There seems to be a Travis failure in the flow job because href isn't a property of userLink? Do you have any idea how to fix this?

mc10: Try something like `(userlink: HTMLAnchorElement).href`?

preethamrn:
Do you know how I can test the travis build locally? Despite the testing I do, I still get errors in the build and I don't want to clutter up the pull request with mostly small bug fixes.

mc10:
If you follow [these instructions](https://github.com/honestbleeps/Reddit-Enhancement-Suite#building-development-versions-of-the-extension), then you should be able to run Flow locally. (I think it's `npm run flow` to run the Flow checker.)

Adding new commits is fine, as we squash all your changes into one commit before merging anyways.

lib/utils/user.js
```diff
@@ -3,11 +3,13 @@
 import _ from 'lodash';
 import { ajax } from '../environment';
 import type { RedditAccount } from '../types/reddit';
-import { MINUTE, string } from './';
+import { MINUTE, downcast, string, regexes } from './';
 
 export const loggedInUser = _.once((): string | void => {
-	const userLink = document.querySelector('#header-bottom-right > span.user > a');
-	return userLink && !userLink.classList.contains('login-required') && userLink.textContent || undefined;
+	const link = document.querySelector('#header-bottom-right > span.user > a');
```
mc10: If you do a `null` check here you might be able to avoid the later `link &&` code? Not entirely sure how Flow checking works though.

PR #3793, opened 2016-12-15
dashboard: Use local ajax over $.ajax; ajax: Create AjaxError

mc10:
I'm not too big of a fan of `ajax`'s error handling; any ideas for improvement?

### DIFF
--- a/lib/modules/dashboard.js
+++ b/lib/modules/dashboard.js
@@ -12,7 +12,7 @@ import {
 	string,
 	newSitetable,
 } from '../utils';
-import { Storage } from '../environment';
+import { Storage, ajax } from '../environment';
 import * as Init from '../core/init';
 import * as Modules from '../core/modules';
 import * as Menu from './menu';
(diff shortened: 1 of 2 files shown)


lib/modules/dashboard.js
```diff
@@ -793,8 +799,8 @@ function WidgetObject(widgetOptions) {
 			newSitetable($widgetContent[0]);
 		}
 	};
-	this.error = xhr => {
-		if (xhr.status === 404) {
+	this.error = e => {
+		if (e.message.startsWith('XHR status 404 ')) {
```
andytuba: Dang, `e` doesn't include the xhr object?

mc10: I don't think so; I think it's just the error object from
```js
throw new Error(`XHR status ${response.status} - url: ${url}`);
```

If there's a nicer approach I'm all ears.

mc10:
Any reason why the Flow checking is failing? I don't think I wrote my `await ajax` call differently than the other call sites...

lib/modules/dashboard.js
erikdesjardins: We can throw a subclass of `Error`:

```js
class AjaxError extends Error {
    status: number;

    constructor(message: string, status: number) {
        super(message);
        this.status = status;
    }
}
```

erikdesjardins:
FYI we're not merging 5.4.0 PRs right now (working towards 5.2.2), I suppose I shouldn't have approved this.

I'll revert if we do a 5.2.2 release (it may not happen).

mc10:
My bad, I didn't notice the tag. Feel free to revert if you'd like.

Issue #3755, opened 2016-12-08
Modules which need i18n


erikdesjardins:
If you want to work on translating any of these modules, just comment on this issue and a maintainer will mark it as taken!

### Basic translation (module descriptions, option descriptions):

- [ ] about.js
- [ ] accountSwitcher.js
- [ ] announcements.js
- [ ] backupAndRestore.js
- [ ] betteReddit.js
- [ ] commandLine.js
- [ ] commentDepth.js
- [ ] commentHidePersistor.js
- [ ] commentNavigator.js
- [ ] commentPreview.js
- [ ] commentStyle.js
- [ ] commentTools.js
- [ ] context.js
- [ ] contribute.js
- [ ] customToggles.js
- [ ] dashboard.js
- [ ] easterEgg.js
- [ ] filteReddit.js
- [ ] floater.js
- [ ] hideChildComments.js
- [ ] hover.js
- [ ] keyboardNav.js
- [ ] localDate.js
- [ ] logoLink.js
- [ ] menu.js
- [ ] messageMenu.js
- [ ] modhelper.js
- [ ] multiredditNavbar.js
- [ ] neverEndingReddit.js
- [ ] newCommentCount.js
- [ ] nightMode.js
- [ ] noParticipation.js
- [ ] notifications.js
- [ ] onboarding.js
- [ ] orangered.js
- [ ] pageNavigator.js
- [ ] presets.js
- [ ] profileNavigator.js
- [ ] quickMessage.js
- [ ] RESTips.js
- [ ] saveComments.js
- [ ] search.js
- [ ] searchHelper.js
- [ ] selectedEntry.js
- [ ] settingsConsole.js
- [ ] settingsNavigation.js
- [ ] showImages.js
- [ ] showKarma.js
- [ ] showParent.js
- [ ] singleClick.js
- [ ] sourceSnudown.js
- [ ] spamButton.js
- [ ] spoilerTags.js
- [ ] stylesheet.js
- [ ] styleTweaks.js
- [ ] submitHelper.js
- [ ] submitIssue.js
- [ ] subredditInfo.js
- [ ] subredditManager.js
- [ ] subredditTagger.js
- [ ] tableTools.js
- [ ] troubleshooter.js
- [ ] userbarHider.js
- [ ] userHighlight.js
- [ ] userInfo.js
- [ ] usernameHider.js
- [ ] userTagger.js
- [ ] version.js
- [ ] voteEnhancements.js
[...]

PR #3802, opened 2016-12-17
[i18n] Add option titles & descriptions in more modules

wojtekmaj:
#3755 

This also fixes the number of strings to translate which is currently at satanic number 666.


erikdesjardins:
Thanks!

(build failure is spurious)

Issue #3797, opened 2016-12-16
Add filterreddit (two R's) to SEO terms for filtereddit


andytuba:
https://www.reddit.com/r/RESAnnouncements/comments/5go2xd/announcement_res_v520_release/dba1rrj/

And for anyone watching, this is RES's internal settings search, not Google SEO.

wojtekmaj:
Consider it done :D

PR #3803, opened 2016-12-17
Searchable module & option keywords

wojtekmaj:
As suggested in #3797, added searchable module & option keywords.

As an example, you can now succesfully search for "filterreddit" and get "filteReddit" module, as well as for the phrase "remember" search will return "keepLoggedIn" option. How cool is that?

Also I've noticed a pattern in building searchable strings so I've split them into separate functions to avoid code duplication and unexpected behavior of search in case someone would change something just in one place.


andytuba:
Looks great! Thanks for quick implementation.

erikdesjardins:
Looks good, thanks!

(build failure is again spurious)

Issue #3800, opened 2016-12-16
Enable hardIgnore by default


andytuba:
https://en.reddit.com/r/Enhancement/comments/5irfcu/what_use_is_an_ignore_feature_if_its_not_going_to/

This complaint is a constant refrain.

PR #3804, opened 2016-12-17
Fixes Issue #3800 and adds support for #3755.

jakewilson:
This should be better

### DIFF
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -50,7 +50,7 @@ module.options = {
 	*/
 	hardIgnore: {
 		type: 'boolean',
-		value: false,
+		value: true,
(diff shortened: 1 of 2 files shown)


lib/modules/userTagger.js
```diff
@@ -50,43 +50,43 @@ module.options = {
 	*/
 	hardIgnore: {
 		type: 'boolean',
-		value: false,
-		description: 'Completely remove links and comments posted by ignored users. If an ignored comment has replies, collapse it and hide its contents instead of removing it.',
+		value: true,
```
andytuba: @erikdesjardins We don't need a migration for this, right?

jakewilson:
I opened another PR using a branch other than master since it's better practice and less stress on my part.

PR #3805, opened 2016-12-17
Add autocomplete for wiki pages of current subreddit

### LINKED ISSUE #3785: /w/ autosuggest wiki pages (a la /r/...)
https://www.reddit.com/r/ideasfortheadmins/comments/5i2i94/comments_with_the_word_wiki_should_be_autosuggest/

https://www.reddit.com/r/Enhancement/comments/5i73k5/feature_request_comments_with_the_word_wiki/

I know I'd love this for the res support subreddits..

wojtekmaj:
I've added a feature (requested in #3785) that enables the users to get autocomplete for wiki pages belonging to the subreddit they are currently on.

image

It works similarly to /r/ and /u/ autocomplete. For example, typing `/wiki/bette` on /r/Enhancement will result in `/index/ui/betteReddit` being suggested. If the user wants to link other's subreddit's wiki, they can write `/r/subredditname/wiki/...` to get the result from this wiki. 

As links to wiki are working in a slightly more complicated way than `/r/...` and `/u/...` links (plain /wiki/ won't work without prepending subreddit link), along with autosuggest item "shortLabel" (the one being displayed), I now pass url in `data-label` attribute.

It is then sent to a function that pastes autosuggestion into an input field. This function, given label, will completely replace autosuggest-triggering phrase, and in other cases will behave like it always used to.

In our case, the final result of this autosuggestion will be: 
`/r/Enhancement/wiki/index/ui/bettereddit`.

Enjoy!

Edit(@erikdesjardins): fixes #3785

### DIFF
--- a/lib/modules/commentTools.js
+++ b/lib/modules/commentTools.js
@@ -10,6 +10,7 @@ import {
 	Thing,
 	checkKeysForEvent,
 	click,
+	currentSubreddit,
 	escapeHTML,
 	isPageType,
 	loggedInUser,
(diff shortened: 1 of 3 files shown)


wojtekmaj:
Phew! It appears everything works as I wanted it to, but it needs quite a bit of testing and I think I'd need some assistance with flow checks.

erikdesjardins:
Great, thanks! Sorry about the delay

PR #3806, opened 2016-12-18
Add automatic Night Mode based on geolocation and suncalc

### LINKED ISSUE #3657: Automatic nightmode based on sunrise/sunset times
Using http://api.sunrise-sunset.org/json and the geolocation API.

See discussion in #2502 (in particular, attribution and usage limits)

wojtekmaj:
As requested in #3657.

It needs some works around descriptions, user preferences migration and error handling. I won't have time to do this so if you want, feel free to finish this :)

fixes #3657

### DIFF
--- a/lib/modules/nightMode.js
+++ b/lib/modules/nightMode.js
@@ -8,6 +8,7 @@ import * as Init from '../core/init';
 import * as Modules from '../core/modules';
 import * as Options from '../core/options';
 import {
+	Alert,
 	BodyClasses,
 	CreateElement,
 	currentSubreddit,
@@ -15,7 +16,7 @@ import {
 	HOUR,
 	MINUTE,
 } from '../utils';
-import { Session, Storage, multicast } from '../environment';
+import { Session, Storage, i18n, multicast } from '../environment';
 import * as CommandLine from './commandLine';
 import * as Menu from './menu';
 import * as StyleTweaks from './styleTweaks';
@@ -274,7 +275,6 @@ function overrideNightMode() {
  * Gets user's latitude and longitude using Geolocation API.
  */
 function getGeolocation() {
-	console.log('getGeolocation');
 	return new Promise((resolve, reject) => {
 		navigator.geolocation.getCurrentPosition(
 			position => resolve(position.coords),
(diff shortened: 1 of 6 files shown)


lib/modules/nightMode.js
```diff
@@ -38,27 +39,34 @@ module.options = {
...
+			name: 'nightModeAutomaticNightModeUser',
+			value: 'user',
+		}],
 		description: 'nightModeAutomaticNightModeDesc',
 	},
 	nightModeStart: {
 		type: 'text',
 		value: '20:00',
 		description: 'nightModeNightModeStartDesc',
-		dependsOn: 'automaticNightMode',
```
wojtekmaj: Would be cool if we could write something like `dependsOn: ['automaticNightMode', (value) => value === 'user']`

lib/modules/nightMode.js
```diff
@@ -242,7 +250,7 @@ const handleAutomaticNightMode = _.once(async function check() {
 });
 
 function overrideNightMode() {
-	if (!module.options.automaticNightMode.value) {
+	if (!module.options.automaticNightMode.value !== 'none') {
```
mc10: You'll need to remove the `!` at the beginning of this line.

lib/modules/nightMode.js
```diff
@@ -38,27 +39,34 @@ module.options = {
 		advanced: true,
 	},
 	automaticNightMode: {
-		type: 'boolean',
+		type: 'enum',
 		value: false,
```
andytuba: `value` needs updating to ... `'user'`?

lib/modules/nightMode.js
```diff
@@ -251,14 +259,34 @@ function overrideNightMode() {
 	nightmodeOverrideStorage.set(Date.now());
 }
 
-function isTimeForNightMode() {
+function getGeolocation() {
+	return new Promise(resolve => {
```
andytuba: To avoid freaking out users, we should show a notification/modal explaining why geolocation is about to be requested before/while asking for permission.

Might need to be "during" to avoid showing explanation if user has already granted permission.

lib/modules/nightMode.js
```diff
@@ -251,14 +259,34 @@ function overrideNightMode() {
...
 	const currentTime = currentHour * 100 + currentMinute;
 
-	const startingTime = convertTimeStringToInt(module.options.nightModeStart.value);
-	const endingTime = convertTimeStringToInt(module.options.nightModeEnd.value);
+	const isAuto = module.options.automaticNightMode.value === 'automatic';
+
+	let startingTimeString, endingTimeString;
+
+	if (isAuto) {
+		const { latitude, longitude } = await getGeolocation();
```
andytuba: It would probably improve UX to optimistically guess at a time (nightModeStart/nightModeEnd by default, or last value) and then in the background fetch a new `getSunriseSunset`

lib/modules/nightMode.js
```diff
@@ -251,14 +259,34 @@ function overrideNightMode() {
 	nightmodeOverrideStorage.set(Date.now());
 }
 
-function isTimeForNightMode() {
+function getGeolocation() {
+	return new Promise(resolve => {
+		navigator.geolocation.getCurrentPosition(position => resolve(position.coords));
```
andytuba: Does `getCurrentPosition` accept a "user denied permission"` callback, so flow can return to the calling function and fail over to `'user'` style?

lib/modules/nightMode.js
andytuba: Additionally, if geolocation is unavailable or denied permission, this function should fail over to `nightModeStart` and `nightModeEnd` options.

andytuba:
Great start on this, thanks!

lib/modules/nightMode.js
wojtekmaj: It used to be false, so I guess it should be 'none', I've fixed that now :)

lib/modules/nightMode.js
wojtekmaj: With the updates I've just pushed - yes :)

lib/modules/nightMode.js
wojtekmaj: I have updated preferences description, would that be alright?

lib/modules/nightMode.js
wojtekmaj: Fallbacks implemented!

erikdesjardins:
Nice, thanks!

PR #3807, opened 2016-12-19
customFilters postAge: Set default comparison operator

larsjohnsen:
Otherwise there will be no operator by default, which when not changed will cause RES to crash when the filter is executed.


erikdesjardins:
CI failure is spurious ([it even says "Build success"](https://ci.appveyor.com/project/honestbleeps/reddit-enhancement-suite/build/3096#L821))

PR #3809, opened 2016-12-20
CommentQuickCollapse

jakewilson:


lib/modules/commentQuickCollapse.js
```diff
@@ -0,0 +1,37 @@
...
+module.go = () => {
+	if (module.options.hideCommentsOnHeaderDoubleClick.value) {
+		hideCommentsOnHeaderDoubleClick();
+	}
+};
+
+function hideCommentsOnHeaderDoubleClick() {
+	$('p.tagline').on('dblclick', (e: Event) => {
+		// the first child of p.tagline is a.expand
+		e.target.children[0].click();
```
andytuba: Could you specifically target `e.target.querySelector('.expand')` instead of `e.target.children[0]`? Just in case the layout shifts around, plus it won't add a significant performance hit.

lib/modules/commentQuickCollapse.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+module.go = () => {
+	if (module.options.hideCommentsOnHeaderDoubleClick.value) {
+		hideCommentsOnHeaderDoubleClick();
+	}
+};
+
+function hideCommentsOnHeaderDoubleClick() {
+	$('p.tagline').on('dblclick', (e: Event) => {
```
andytuba: How about `$(document.body).on('dblclick', '.tagline', ...)` to support comments loaded dynamically?

andytuba:
Looks like you started your branch on top of your previous project work. You should probably...

```
git remote add upstream https://github.com/honestbleeps/Reddit-Enhancement-Suite.git
git fetch upstream
git checkout commentHider
git reset upstream/master
git cherry-pick 78dea26
git cherry-pick c2c9f3a
git push -f
```

And when you're starting a new project,
```
git fetch upstream && git checkout upstream/master
git checkout -b myNewProject
```

lib/modules/commentQuickCollapse.js
jakewilson: Yes, I didn't like doing .children[0] but couldn't think of a more dynamic way to do it. Thanks.

jakewilson:
So, I already closed my other PR is it okay if I leave this one open with all the commits?

jakewilson:
Updated.

lib/modules/commentQuickCollapse.js
jakewilson: I did this, but can you elaborate on this? How do comments get loaded dynamically?

lib/modules/commentQuickCollapse.js
```diff
@@ -0,0 +1,37 @@
...
+
+module.go = () => {
+	if (module.options.hideCommentsOnHeaderDoubleClick.value) {
+		hideCommentsOnHeaderDoubleClick();
+	}
+};
+
+function hideCommentsOnHeaderDoubleClick() {
+	$(document.body).on('dblclick', '.tagline', (e: Event) => {
+		// the first child of p.tagline is a.expand
```
andytuba: Comment is superfluous now.

andytuba:
Looking good, and I'll allow the mashup of projects. Not tested yet, though.

lib/modules/commentQuickCollapse.js
jakewilson: Oops, thanks. Is this too little for one commit?

PR #3812, opened 2016-12-21
 keyboardNav: Use requiresModule where more appropriate than `include`

larsjohnsen:


erikdesjardins:
build failure is spurious

PR #3816, opened 2016-12-22
Adds i18n to some more files.

jakewilson:
Issue #3755 

Adds translated strings to neverEndingReddit.js and multiredditNavbar.js.


erikdesjardins:
Looks like this was already done by #3802 (sorry!)

Make sure you claim modules on #3755 so we don't accidentally duplicate effort.

PR #3818, opened 2016-12-22
try updating appveyor yarn with self-update

erikdesjardins:


erikdesjardins:
(somewhat) amusingly the self-update command was removed between 0.16.1 and 0.18.1, so the build will fail again once Appveyor starts shipping Yarn 0.18.1. But that's fine, because we can just remove this line entirely at that point.

PR #3820, opened 2016-12-23
Issue 3755: Adds more translated strings.

jakewilson:
Adds translated strings to
- presets.js
- quickMessage.js
- RESTips.js
- searchHelper.js


erikdesjardins:
Thanks!

PR #3825, opened 2016-12-24
[experimental] sync the 100 most popular subreddits to native /r/all filters

erikdesjardins:
This somewhat abuses reddit's API, so it may not be a viable solution.

We really need a batch subreddit info endpoint.


erikdesjardins:
Verdict: this will be called infrequently enough (only after a [relatively rare] user action: filtering a subreddit) that it shouldn't be a problem for reddit. If it turns out to be, we can revert it.

Once a better alternative is available (batch subreddit endpoint), it will be replaced.

PR #3826, opened 2016-12-24
accountSwitcher: clear `token` cookie for new modmail

### LINKED ISSUE #3822: Account switcher doesn't switch new modmail
Iirc @creesch recommended trashing the token cookie when switching accounts, and the  loading  new modmail will start a new session under the  current account.

erikdesjardins:
Fixes #3822

As #3822 predicted, all we need to do is remove the cookie and new modmail will start a new session on next pageload.


erikdesjardins:
Test failure is spurious

PR #3828, opened 2016-12-24
Update streamable favicon, old CDN offline.

BenMcGarry:
Appears to use fastly with tokens to pull the images now. API docs are out of date.


erikdesjardins:
Test failure is spurious (#3829)

PR #3834, opened 2016-12-25
Implement search as you type, search performance improvements

wojtekmaj:
Start searching for options immediately after the user starts typing in the search box.
Mad progrmming skillz.

Also vastly improved search drawing performance by switching to pure JavaScript:
image

### DIFF
--- a/lib/modules/search.js
+++ b/lib/modules/search.js
@@ -283,11 +283,6 @@ function drawSearchResults(query, results) {
 	$resultsContainer.show();
 	$resultsContainer.find('#SearchRES-query').text(query);
 
-	const searchInput = document.getElementById('SearchRES-input');
-	if (searchInput.value !== query) {
-		searchInput.value = query;
-	}
-
 	if (advancedResults) {
 		$resultsContainer.find('#SearchRES-results-hidden').addClass('advancedResults');
 		$('#SearchRES-results-hidden a').off('click').on('click', () => {
--- a/lib/modules/search.js
+++ b/lib/modules/search.js
@@ -300,32 +300,37 @@ function drawSearchResults(query, results) {
 	} else {
 		$resultsContainer.find('#SearchRES-results-hidden').removeClass('advancedResults');
 	}
+
 	if (!results || !results.length) {
 		$resultsContainer.find('#SearchRES-results').hide();
 	} else {
-		const resultsList = $('#SearchRES-results', $resultsContainer).show();
+		$resultsContainer.find('#SearchRES-results').show();
 
-		resultsList.empty();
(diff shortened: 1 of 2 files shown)


lib/modules/search.js
```diff
@@ -62,6 +62,27 @@ function showSearch() {
 
 const PRESERVE_SPACES = true;
 
+const searchDomain = _.once(() => {
```
andytuba: May as well leave this where it was and preserve the blame a little?

lib/modules/search.js
```diff
@@ -300,32 +300,37 @@ function drawSearchResults(query, results) {
...
+	element.classList.add('SearchRES-result-item');
+	if (result.advanced) {
+		element.classList.add('advanced');
+	}
 
 	const details = searchResultTemplate(result);
-	element.append(details);
+	element.innerHTML = details;
+
+	const copybutton = document.createElement('span');
```
andytuba: You can also use the helper function `CreateElement.icon('F159', 'span', 'SearchRES-result-copybutton', i18n('searchCopyResultForComment'));`

and in `en.json` add `searchCopyResultForComment`

lib/modules/search.js
```diff
@@ -248,7 +248,7 @@ function onSearchResultSelected(result) {
 export function renderSearchForm(container: HTMLElement) {
 	$(container).find('#SearchRES-input-container').replaceWith(searchFormTemplate());
 
-	$(container).on('submit', '#SearchRES-input-container', e => {
+	$(container).on('keyup submit', '#SearchRES-input-container', e => {
```
andytuba: Could you add a throttle around the event handler to avoid thrashing lower-performance computers?

andytuba:
Neat! Live search is way better than having to actually click the button.

wojtekmaj:
@andytuba I believe I've done everything you requested, correct me if I'm wrong :)

lib/modules/search.js
```diff
@@ -54,6 +54,8 @@ export function search(query?: string) {
 	SettingsNavigation.setUrlHash(module.moduleID, query);
 }
 
+const debouncedSearch = _.debounce(search, 250);
```
larsjohnsen: I think `frameThrottle` is a more natural choice here since it won't use any more resources than available. In addition, debouncing prevents it from being "search as you type" since it will only search after stopping typing.

wojtekmaj: Oh, I didn't know about that one :) Thank you!

erikdesjardins:
Neat, thanks!

I also threw in some optimization on the match-finding side; we were doing a lot more work than necessary (localizing each string every time, etc.).

PR #2877, opened 2016-04-10
Liveleak host

erikdesjardins:
https://www.reddit.com/r/Enhancement/comments/4e8416/feature_request_a_dropdown_video_player_for/

~~Alas, Liveleak has zero HTTPS support, so this doesn't work. As soon as they do, it should.~~

Closes #3837 (superseded)

PR #3837, opened 2016-12-27
LiveLeak embed

wojtekmaj:
I am not entirely sure about this solution as we take MP4 files directly from them instead of using their player, as their player won't work on HTTPS pages... But hey, it works! What do you think?

### DIFF
--- a/lib/modules/hosts/liveleak.js
+++ b/lib/modules/hosts/liveleak.js
@@ -1,0 +1,30 @@
+/* @flow */
+
+import { Host } from '../../core/host';
+import { ajax } from '../../environment';
+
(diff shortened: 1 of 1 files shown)


BenMcGarry:
ref #2877

erikdesjardins:
Using the mp4 files directly is fine IMO; it's what the embed would be doing anyways.

However, fetching the page and scraping the URL out of it is less than ideal. If there's some other way to get the source URL (an API or constructing it manually) then I'd probably be fine with that, but otherwise I don't think there's enough need for Liveleak embeds for us to resort to scraping.

PR #3838, opened 2016-12-28
Issue 3755: More i18n.

jakewilson:
Adds translated strings to
- selectedEntry.js
- settingsNavigation.js

Also a note that `settingsConsole.js` has no options, so it can be marked as finished in #3755.


locales/locales/en.json
```diff
@@ -3063,5 +3063,61 @@
...
+    "selectedEntryFocusFGColorNightDesc": {
+        "message": "The text color while using Night Mode"
+    },
+
+    "selectedEntryAddFocusBorderDesc": {
+        "message": "Use a border"
+    },
+
+    "selectedEntryFocusBorderDesc": {
+        "message": "Border appearance. E.g. <code>1px dashed gray</code> (CSS)"
```
wojtekmaj: That might not work as you expect it to

jakewilson: I checked it in my browser before committing, shows up as expected. Good eye!

erikdesjardins: It does work but we want to avoid it because eventually we will switch module/option descriptions to markdown (I should write that down somewhere).

In the case of something like a `<table>` you can just leave it untranslated for now and we'll replace all of those at the end; in this case I changed it to ````1px dashed gray```` because it looks fine both raw: \`1px dashed gray\` and in markdown: `1px dashed gray`.

PR #3839, opened 2016-12-28
Add option to page navigator to scroll to comment box

### LINKED ISSUE #67: [Req] Hide/move comment box
Most users don't comment right away when opening the comments. It would be nice to have the option to collapse/hide the 
user comment text box (usertext.cloneable) and/or to move it to the bottom.

dhensche:
fixes #67

### DIFF
--- a/lib/modules/pageNavigator.js
+++ b/lib/modules/pageNavigator.js
@@ -64,7 +64,7 @@ function backToTop() {
 }
 
 function backToNewCommentArea() {
-	const commentArea = document.querySelector('.commentarea > form.usertext');
+	const commentArea = document.querySelector('.commentarea > form.usertext textarea');
 	if (!commentArea) return;
 	const $backToNewCommentArea = $(`<a class="pageNavigator res-icon" data-id="addComment" href="#comments" title="${i18n('pageNavToCommentTitle')}">&#xF139;</a>`);
 	$backToNewCommentArea.on('click', (e: Event) => {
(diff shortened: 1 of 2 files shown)


lib/modules/pageNavigator.js
```diff
@@ -39,20 +45,32 @@ module.afterLoad = () => {
 	if (module.options.toTop.value) {
 		backToTop();
 	}
+	if (module.options.toComment.value) {
```
andytuba: Can this also test if the comment area is available? In case user is logged out or banned. (Probably should be a utility function, although dunno which section that should go in.)

lib/modules/pageNavigator.js
```diff
@@ -39,20 +45,32 @@ module.afterLoad = () => {
...
-	$backToTop.on('click', '[data-id="top"]', (e: Event) => {
+	$backToTop.on('click', (e: Event) => {
 		e.preventDefault();
 		scrollTo(0, 0);
 	});
 	Floater.addElement($backToTop);
 }
 
+function backToAddComment() {
+	const $backToAddComment = $('<a class="pageNavigator res-icon" data-id="addComment" href="#comments" title="back to add comment box">&#xF139;</a>');
```
andytuba: Can you i18n the title too?

lib/modules/pageNavigator.js
dhensche: can I just use the selector `.usertext-edit textarea` to check for existence of the comment area? I don't know of a better way using existing utility functions

PR #3840, opened 2016-12-28
showImages YouTube: Remove iframe when collapsing

larsjohnsen:
Using the `stop` command resets the playback position to 0 (not the position specified in the URL), which often is more of a pain than having to load the iframe content anew.

### DIFF
--- a/lib/modules/hosts/youtube.js
+++ b/lib/modules/hosts/youtube.js
@@ -56,8 +56,6 @@ export default new Host('youtube', {
 			type: 'IFRAME',
 			embed: src,
 			embedAutoplay: `${src}&autoplay=1`,
-			pause: '{"event":"command","func":"stopVideo","args":""}',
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/youtube.js
```diff
@@ -56,7 +56,6 @@ export default new Host('youtube', {
 			type: 'IFRAME',
 			embed: src,
 			embedAutoplay: `${src}&autoplay=1`,
-			pause: '{"event":"command","func":"stopVideo","args":""}',
 			play: '{"event":"command","func":"playVideo","args":""}',
```
erikdesjardins: Does `play` have any use without `pause`?

larsjohnsen: No. It's redundant since there `embedAutoplay` exists.

erikdesjardins:
Thanks!

PR #3845, opened 2017-01-01
selectedEntry: Fix onNewComments on outermost "load more comments"

larsjohnsen:
The previously implementation verified that the new comment should be selected by matching a parent Thing. This matches `.sitetable` instead.

Fixes https://www.reddit.com/r/RESissues/comments/5le5xq/active_selection_resets_after_using_the_enter/


erikdesjardins:
Nice, thanks!

PR #3854, opened 2017-01-05
travis: Cache yarn (built-in) and node_modules

mc10:
This should be possible now that travis-ci/travis-ci#6720 and yarnpkg/yarn#1576 are both fixed. See also the [official Travis blog post about yarn](https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ addons:
     secure: hHF06xHY6pLdxscwYhUHdSRuUH2CPA61AUrvPiW54lnMwwkDvCQakR+aY5HbjptFIfOwkWsS4xoqXH4pZTfhMOji32S/+ELT+AqLdkIgnLYDfBj4RcsZoXyvuiLx29i4PVCnGMikctBp4dM6wORgfd1tu3uwhcoQrBjxCMW1qmY=
 cache:
   yarn: true
+  directories:
+    - node_modules
 matrix:
   exclude:
     - node_js: 7 # exclude default job https://github.com/travis-ci/travis-ci/issues/4681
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -23,12 +23,9 @@ addons:
     # SAUCE_ACCESS_KEY=<secure>
     secure: hHF06xHY6pLdxscwYhUHdSRuUH2CPA61AUrvPiW54lnMwwkDvCQakR+aY5HbjptFIfOwkWsS4xoqXH4pZTfhMOji32S/+ELT+AqLdkIgnLYDfBj4RcsZoXyvuiLx29i4PVCnGMikctBp4dM6wORgfd1tu3uwhcoQrBjxCMW1qmY=
 cache:
+  yarn: true
```
erikdesjardins: The cache dir was explicit because Travis caches the wrong directory for `yarn: true` (`~/.yarn-cache` or something), though that may be fixed

mc10: I sent a [PR to Travis](https://github.com/travis-ci/travis-build/pull/920), hopefully they fix the issue.

mc10:
That's fair. We can wait to merge this in until Travis adds yarn to their images. I'll add the install commands in at that time.

.travis.yml
mc10: This should be fixed on Travis's end.

mc10:
I removed the use of Travis's own `yarn`, as it prevents us from having a custom `install` phase. This PR now only modifies the `cache` section.

Issue #1828, opened 2014-11-29
Account switcher dropdown links to profile


andytuba:
https://www.reddit.com/comments/2nn963/slug/cmg6bun

erikdesjardins:
https://www.reddit.com/r/Enhancement/comments/31wbnr/feature_request_add_a_couple_of_simple_but_useful/
(and an option to log in to that account at the same time)

PR #3855, opened 2017-01-05
Add link in account switcher dropdown to the accounts profile page

thybag:
Fixes https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/1828

### DIFF
--- a/lib/modules/accountSwitcher.js
+++ b/lib/modules/accountSwitcher.js
@@ -252,7 +252,7 @@ async function populateUser(container) {
 		}
 
 		// Append link to profile page
-		$contents.append($('<a>', { href: `/u/${username}`, class: 'res-icon linkIcon' }).click(e => e.stopPropagation()));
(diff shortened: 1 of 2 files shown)


lib/modules/accountSwitcher.js
```diff
@@ -250,6 +250,9 @@ async function populateUser(container) {
 
 			$contents.append($gilded);
 		}
+
+		// Append link to profile page
+		$contents.append($('<a>', { href: `/u/${username}`, class: 'res-icon linkIcon' }).click(e => e.stopPropagation()));
```
andytuba: Could you link to `/user/${username}/` to avoid a redirect?

andytuba: and swap `.click()` for `.on('click', e => ...)` just to clarify "this is an event listener"?

thybag: Cool, both done 👍

PR #3858, opened 2017-01-08
Integrate updates css

roshkins:
Some of the css refers to features that needs to be implemented.

### DIFF
--- a/lib/css/modules/_neverEndingReddit.scss
+++ b/lib/css/modules/_neverEndingReddit.scss
@@ -115,6 +115,10 @@
 #NREPause {
 	cursor: pointer;
 
+	.res-neverEndingReddit-hideFloatingPauseButton & {
+	    display: none;
+	}
+
 	&::after {
 		font: normal 100% Batch;
 		color: #888;
--- a/lib/css/modules/_neverEndingReddit.scss
+++ b/lib/css/modules/_neverEndingReddit.scss
@@ -115,10 +115,6 @@
 #NREPause {
 	cursor: pointer;
 
-	.res-neverEndingReddit-hideFloatingPauseButton & {
-	    display: none;
-	}
-
 	&::after {
 		font: normal 100% Batch;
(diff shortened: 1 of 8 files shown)


lib/css/modules/_nightMode.scss
```diff
@@ -1382,3 +1382,19 @@
 		}
 	}
 }
+
+
+.res-nightmode:not(.res-nightMode-coloredLinks){
```
andytuba: Does nightmode options include a colored links boolean body class option yet? Should probably be enabled by default.

lib/css/res.scss
```diff
@@ -43,6 +43,9 @@
 
 body.res-console-open {
 	overflow: hidden;
+	&.front-page.res-hide-tagline-frontpage .tagline {
```
andytuba: Looks like a merge conflict resolution issue here, this probably shouldn't be inside the console open block.

lib/css/modules/_newCommentCount.scss
```diff
@@ -3,6 +3,10 @@
 	color: orangered;
 }
 
+.new-comment .usertext-body {
```
andytuba: This seems like it should move inside a "new comments count" block.

lib/css/modules/_neverEndingReddit.scss
```diff
@@ -115,6 +115,10 @@
 #NREPause {
 	cursor: pointer;
 
+	.res-neverEndingReddit-hideFloatingPauseButton & {
```
andytuba: This should become a disabled boolean body class option, with a migration to enable it if found in a stylesheet module's option  body class value,  and filter out the row containing it.

andytuba: *there should be added

erikdesjardins:
Thanks!

<sup>(so the build passed but it's marked as errored...hmmm)</sup>

PR #3863, opened 2017-01-09
Redditmin

### LINKED ISSUE #3847: Make comment collapse buttons 100% height
- [ ] https://github.com/KetillG/RedditMin
- [ ] In PR, please `Closes #3847`
- [ ] revert https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/8472f16382791660aae6047ef711bb963ee23cae
- [ ] maybe add migration to disable ^ that option and enable the module

KetillG:
Closes #3847


locales/locales/en.json
```diff
@@ -3267,5 +3267,33 @@
...
+    "toggleCommentsLeftEdgeCollapsedColorDesc": {
+        "message": "Color of the bar when hovered."
+    },
+
+    "toggleCommentsLeftEdgeWidthDesc": {
+        "message": "Width of the bar."
+    },
+
+    "toggleCommentsLeftEdgeHideButtonDesc": {
+        "message": "Hides the default button ([-]) to collaps comments."
```
andytuba: collapse*

lib/css/modules/_commentQuickCollapse.scss
```diff
@@ -0,0 +1,44 @@
+.res-commentQuickCollapse-toggleCommentsOnClickLeftEdge #siteTable ~ .commentarea {
+	.comment  {
+	/*
```
andytuba: Is this commented-out block useful? Since RES already has the commentBoxes styleTweaks, it feels superfluous here.

KetillG: No it's commented-out anyway so it's not doing anything. Removed.

lib/css/modules/_commentQuickCollapse.scss
```diff
@@ -1,13 +1,11 @@
 .res-commentQuickCollapse-toggleCommentsOnClickLeftEdge #siteTable ~ .commentarea {
 	.comment  {
-	/*
-		box-shadow: 0px 1px 5px rgba(0,0,0,.16) !important;
-		margin-bottom: 10px !important;
-		margin-left: 0px !important;
-	*/
+		margin-left: 2px !important;
```
KetillG: /r/pokemon looked strange without this.

lib/modules/commentQuickCollapse.js
```diff
@@ -11,14 +16,78 @@ module.category = 'commentsCategory';
...
+	},
+	hideCollapseButton: {
+		dependsOn: 'toggleCommentsOnClickLeftEdge',
+		type: 'boolean',
+		value: true,
+		bodyClass: true,
+		description: 'toggleCommentsLeftEdgeHideButtonDesc',
+	},
+	scrollOnCollapse: {
+		dependsOn: 'toggleCommentsOnClickLeftEdge',
```
erikdesjardins: If this is intended to be dependent on `toggleCommentsOnClickLeftEdge`, it should be moved inside the `if (module.options.toggleCommentsOnClickLeftEdge.value) {` block in `module.go`, so it doesn't take effect when the former is disabled.

erikdesjardins: fca902719e007c00bf56cdb98a96257a36fe4d06

lib/modules/commentQuickCollapse.js
```diff
@@ -58,7 +58,7 @@ module.options = {
 	hideCollapseButton: {
 		dependsOn: 'toggleCommentsOnClickLeftEdge',
 		type: 'boolean',
-		value: true,
+		value: false,
```
KetillG: Since you added a check to see if the whole module was disabled I think this should be on by default now. The feature doesn't look that good with this enabled.

andytuba:
https://www.reddit.com/r/Enhancement/comments/5naw9r/feature_request_neat_subreddit_css_feature/ :grin:

lib/modules/commentQuickCollapse.js
```diff
@@ -35,3 +107,40 @@ function hideCommentsOnHeaderDoubleClick() {
...
+}
+// scrolls on collapse
+function scrollOnCollapse() {
+	$(document.body).on('click', '.expand', (event: MouseEvent) => {
+		// Stops event if condition not met
+		if (click.isProgrammaticEvent(event) || Thing.checkedFrom(event.target).thing.classList.contains('noncollapsed')) {
+			return;
+		}
+		// Scrolls by a default value if top of comment is out of window
+		if ($(event.target).offset().top < $(window).scrollTop()) {
```
larsjohnsen: This and below can probably be replaced with

```
			KeyboardNav.move(
				thing => thing.getClosest(thing.getNextSibling, { direction: 'down' }),
				{ scrollStyle: 'top' }
			);
```

though the `move` function (https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/737b617c4de860ef322a54e6a6267ff66f321d16/lib/modules/keyboardNav.js#L966) needs to be exported first.

andytuba: We really ought to refactor all the KeyboardNav "move around in the comment tree" into its own library instead of calling an options callback like that.

KetillG: That sounds like a great idea. Would probably be simple to substitute the code.

larsjohnsen:
@KetillG The code seems correct, but it in order to get `scrollStyle: 'middle'` working you need to merge the latest master.

PR #3871, opened 2017-01-09
Fix command line settings search

### LINKED ISSUE #3860: RES Command line `settings some search query` doesn't search for query
at least it loads the search page

erikdesjardins:
Fixes #3860

### DIFF
--- a/tests/commandLine.js
+++ b/tests/commandLine.js
@@ -1,5 +1,11 @@
 module.exports = {
 	'basic functionality': browser => {
+		if (browser.options.desiredCapabilities.browserName === 'firefox') {
+			// geckodriver doesn't support elementSendKeys https://github.com/mozilla/geckodriver/issues/159
+			browser.end();
+			return;
+		}
+
 		browser
 			.url('https://en.reddit.com/wiki/pages/')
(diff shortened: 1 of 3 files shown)


tests/commandLine.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+	'basic functionality': browser => {
+		browser
+			.url('https://www.reddit.com/wiki/pages/')
+			.waitForElementVisible('#RESSettingsButton')
+			.keys([browser.keys['.']])
+			.assert.visible('#keyCommandLineWidget')
+			.keys([browser.keys.SHIFT, browser.keys.SLASH]) // "?"
+			.assert.containsText('#keyCommandForm', 'navigates to subreddit')
```
andytuba: That'll fail when it gets i18n'd.  How about some command name, or bullet points or something?

erikdesjardins:
```
erik [6:27 PM] I'll explicitly make it `en.reddit.com` so it's not i18n'd if that sounds good, since there's not much else to look for (classless <li>s, but they could be empty if it bugs out)
andytuba [6:29 PM] +1
```

PR #3878, opened 2017-01-11
[experimental] even more test workers

erikdesjardins:


erikdesjardins:
Okay, no change as expected. Sauce Labs limits us to 5 concurrent test runners and other requests block until one is available.

Issue #1234, opened 2014-06-11
Add a way to disable res temporary (mostly for dev purpose)


matheod:
In troubleshooter -> Reload page an disable RES for this tab

=> Reload the page and set a flag in sessionStorage => init.js check this flag at loading

>00:48:05	andytuba  oh yeah okay here's an idea
00:48:12	andytuba  add an option to the RES menu
00:48:20	andytuba  "reload without RES"
00:48:28	andytuba  that dumps a flag in RESstorage or sesionstorage or something
00:48:33	matheod__  yea nice idea
00:48:34	andytuba  and when RES loads it can check for that flag
00:48:37	 matheod_ has quit: Ping timeout: 121 seconds.
00:49:06	andytuba  i think sessionstorage is better because you can just close the tab to go back to normal
00:49:10	matheod__  yea
00:49:11	andytuba  session stoarge is per tab, right?
00:49:13	matheod__  yes
00:49:39	matheod__  This option should be in troubleshooter I think
00:49:46	andytuba  that sounds good

erikdesjardins:
Fixed by #1235

PR #3884, opened 2017-01-11
Added .res-orangered-hideEmptyModmailButton to orangered css

### LINKED ISSUE #3883: Orangered options to hide envelope and modmail buttons if no mail
https://en.reddit.com/r/Enhancement/comments/5nb5in/feature_request_hide_the_old_modmail_button_if/

The new modmail is excellent and all of the subreddits that I mod have switched over to it. I assume most moderation teams have done the same. While replies to conversations from prior to the switch (>1 month ago) still go to the old modmail, this is a very rare occurrence. There's thus no need to show the old modmail button in the vast majority of situations, and it would be nice if it was hidden.

kevinliu6102:
Fixes #3883

### DIFF
--- a/lib/css/modules/_orangered.scss
+++ b/lib/css/modules/_orangered.scss
@@ -25,6 +25,13 @@
 	}
 }
 
+.res-orangered-hideEmptyModmailButton {
+	#modmail.nohavemail,
+	#modmail.nohavemail + .separator {
+	    display: none;
+	}
+}
+
 .res-orangered-retroUnreadCount {
 	.message-count {
 		background: none;
--- a/lib/css/modules/_orangered.scss
+++ b/lib/css/modules/_orangered.scss
@@ -33,15 +33,15 @@
 }
 
 .res-orangered-hideEmptyMail {
-	a#mail.nohavemail,
-	a#mail.nohavemail + .separator {
+	#mail.nohavemail,
+	#mail.nohavemail + .separator {
 		display: none;
 	}
 }
 
 .res-orangered-hideEmptyNewModMail {
-	a#new_modmail.nohavemail,
-	a#new_modmail.nohavemail + .separator {
+	#new_modmail.nohavemail,
+	#new_modmail.nohavemail + .separator {
 		display: none;
 	}
 }
--- a/lib/css/modules/_orangered.scss
+++ b/lib/css/modules/_orangered.scss
@@ -19,7 +19,7 @@
 }
 
 .res-orangered-hideModMail #modmail,
-.res-orangered-hideNewModMail #new_modmail,
+.res-orangered-hideNewModMail #new_modmail
 {
 	&,
 	& + .separator {
--- a/lib/css/modules/_orangered.scss
+++ b/lib/css/modules/_orangered.scss
@@ -25,7 +25,7 @@
 	}
 }
 
-.res-orangered-hideEmptyModmailButton {
+.res-orangered-hideEmptyModMailButton {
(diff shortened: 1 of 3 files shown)


lib/css/modules/_orangered.scss
```diff
@@ -25,6 +25,13 @@
 	}
 }
 
+.res-orangered-hideEmptyModmailButton {
```
andytuba: Can you add an option to orangered.js to activate this rule? You can model it on the [`hideModmail` option](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/e70d7963907dd8e2ec18db079fecf398de13cd6b/lib/modules/orangered.js#L75).

kevinliu6102: Hopefully the update works. I'm a beginner so a lot of this is actually flying over my head; it's not clear to me what you're asking for in request 1. I may have bitten off more than I can chew. Sorry about that.

kevinliu6102: Also realized #1234 was fixed already?

andytuba:
To clarify, you should put "Fixes #1234" in the PR _description_ (not the title. change the "Addition for issue.."). and you shouldn't use literally 1234, you should use the issue number that you linked in the first place.

kevinliu6102:
Like so? 

Is there something I'm missing in orangered.js?

WilliamChiu:
Might want to replace "Fixes #1234" with "Fixes #3883"

lib/css/modules/_orangered.scss
```diff
@@ -25,7 +25,7 @@
 	}
 }
 
-.res-orangered-hideEmptyModmailButton {
+.res-orangered-hideEmptyModMailButton {
 	#modmail.nohavemail,
 	#modmail.nohavemail + .separator {
 	    display: none;
```
WilliamChiu: Before 'display: none;', use a tab instead of four spaces in order to pass the lint check

kevinliu6102: Thanks for the advice.

andytuba:
@kevinliu6102 👍  for the option you added to orangered.js.

It would also be good to add two more options / CSS rules which do the same thing for the main envelope icon `a#mail` and the new modmail icon. `a#new_modmail`

kevinliu6102:
@andytuba I think this should do it?

andytuba:
@kevinliu6102 Yay! Just one more thing -- adding the titles/descriptions (e.g. `orangeredHideEmptyModMailDesc `) into `en.json`.

It would also be nice to order the options "mail, modmail, newmodmail".

lib/css/modules/_orangered.scss
```diff
@@ -25,6 +25,27 @@
...
+	}
+}
+
+.res-orangered-hideEmptyNewModMail {
+	#new_modmail.nohavemail,
+	#new_modmail.nohavemail + .separator {
+		display: none;
+	}
+}
+
```
larsjohnsen: These may be more readable as

```
.res-orangered-hideEmptyMail #mail,
.res-orangered-hideEmptyModMail #modmail,
.res-orangered-hideEmptyNewModMail #new_modmail {
	&.nohavemail,
	&.nohavemail + .separator {
		display: none;
	}
}

kevinliu6102:
@andytuba I actually went ahead and added a general option for hiding the new modmail button since I saw the original modmail button had such an option. If it's not needed, just let me know and I'll get rid of it.
@larsjohnsen Thanks for the advice.

PR #3888, opened 2017-01-13
Thing: cache getSubreddit result

larsjohnsen:
Since `Thing.getSubreddit` may be invoked on every `Thing` by all subreddit filters, caching the result significantly reduces the time spent on filtering. (In my case 350 ms → 120 ms)


erikdesjardins:
Damn, I didn't consider that side-effect of moving legacy filters to custom filters. Thanks.

PR #3890, opened 2017-01-13
commentNavigator: Fix sortType popular

larsjohnsen:
It did not work nicely with `onScroll` since that assumes that the posts are ordered from top to bottom; `popular` sorts posts by score.

### DIFF
--- a/lib/modules/commentNavigator.js
+++ b/lib/modules/commentNavigator.js
@@ -149,6 +149,7 @@ const sortTypes = {
 				map(([, thing]) => thing.element)
 			)();
 		},
+		nonlinear: true,
 	},
 	new: {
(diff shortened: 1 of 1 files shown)


lib/modules/commentNavigator.js
```diff
@@ -198,6 +199,10 @@ let currentCategory = '';
 function onScroll() {
 	const category = currentCategory;
 	if (category) {
+		// Non-linear comment categories are not sorted by distance from top,
+		// so the most matching comment cannot determined by the algorithm below
+		if (sortTypes[category].nonlinear) return;
```
erikdesjardins: Would `if (currently selected post is on screen) return;` also work? I think I'd prefer that so that there's still _some_ degree of functionality for `popular`

PR #3892, opened 2017-01-14
Accept function for option dependsOn

andytuba:
Enable complex dependsOn behavior like values of an enum

attn: @KetillG

### DIFF
--- a/lib/core/module.js
+++ b/lib/core/module.js
@@ -53,7 +53,7 @@ export type BooleanOption = {|
 	keywords?: Array<string>,
 	value: boolean,
 	bodyClass?: boolean | string,
-	dependsOn?: string,
+	dependsOn?: string | function,
 	advanced?: boolean,
 	noconfig?: boolean,
 	onChange?: () => void,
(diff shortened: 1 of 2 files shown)


lib/core/module.js
```diff
@@ -53,7 +53,7 @@ export type BooleanOption = {|
 	keywords?: Array<string>,
 	value: boolean,
 	bodyClass?: boolean | string,
-	dependsOn?: string,
+	dependsOn?: string | function,
```
erikdesjardins: `dependsOn?: string | () => boolean,`

erikdesjardins:
I want to do a bit more with this: require dependsOn to be a function (so we catch misspellings etc.) and possibly subsume noConfig, but I'll just push to this branch to take care of that (unless you want to)

lib/core/module.js
andytuba: oh duh i have to specify the function signature

andytuba:
It would be nice to preserve the simple legacy behavior of `dependsOn` as "option whose value is truthy". The settingsConsole framework would throw an error either way, if it's a plain string indirection or directly referenced within a function.  ... Would flow catch a misspelled direct reference to an option?

Adding dependsOn to bodyClass sounds great.

Go ahead with pushing to this branch.

andytuba:
:warning: TODO: `settingsConsole` should call `onOptionChanged` for more option types, e.g. enum's, and expand `onOptionChanged` to also handle the function type.

PR #3899, opened 2017-01-18
Add "warning" candystripe overlay to spoiler expando buttons

andytuba:
screen shot 2017-01-18 at 9 12 16 am

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -24,7 +24,7 @@ img {
 
 	.res-showImages-highlightSpoilerButton .thing.spoiler &::before {
 		@extend %expando-button-overlay;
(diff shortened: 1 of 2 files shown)


lib/css/modules/_showImages.scss
```diff
@@ -10,16 +10,25 @@ img {
...
+	pointer-events: none;
+}
+
 .expando-button {
 	cursor: pointer;
 	user-select: none;
 
+	.res-showImages-highlightSpoilerButton .thing.spoiler &::before {
+		@extend %expando-button-overlay;
+		background-image: linear-gradient(45deg, rgba(255, 130, 0, .30) 25%, transparent 25%, transparent 50%, rgba(255, 130, 0, .30) 50%, rgba(255, 130, 0, .30) 75%, transparent 75%, transparent);
```
erikdesjardins: Maybe switch the transparent and colored sections here so it doesn't overlap with NSFW stripes?

andytuba: NSFW styling should override spoiler IMO. Or should we add a combo rule to show both stripes at once?

PR #3905, opened 2017-01-19
Escape the lookup url during reverse image search

### LINKED ISSUE #3904: Image reverse searches fail on redditupload images
1) Find static redditupload image on frontpage
2) Hover over and select the Eye icon to search for it on Google
3) "The URL doesn't refer to an image or the image is not publicly accessible."

In action: https://gfycat.com/AmbitiousInconsequentialGalapagosdove

WilliamChiu:
Fixes #3904 
The lookup url should be escaped because of the query string parameters.


erikdesjardins:
Thanks!

WilliamChiu:
Thanks! Didn't notice the string encoding util

PR #3911, opened 2017-01-20
Adds /r/all to the "Go Mode" navigation

corylulu:
Adds /r/all to the "Go Mode" navigation. 
Includes changes from commit #3909 as well to prevent the 'A' hotkey for /r/all also firing the "Upvote" default hotkey.

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -700,7 +700,7 @@ module.options = {
 	},
 	slashAll: {
 		type: 'keycode',
-		value: [65, false, false, false, false], // a
+		value: [65, true, false, false, false], // alt-a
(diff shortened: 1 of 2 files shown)


erikdesjardins:
Thanks!

I changed the default to alt-A, because even with #3909 it's possible to disable the G prefix for goMode (the `useGoMode` option), so anyone with that off would navigate to /r/all when they try to upvote.

(test failure expected as I manually cancelled the integration test build)

corylulu:
Umm, maybe an alternative default could be set-up for when goMode is / isn't enabled, or simply strip the modifier keys out of the GoMode hotkeys when enabled. Think having to include a modifier kinda defeats the purpose of using GoMode.

PR #3913, opened 2017-01-20
Merge translations and edge ver bump

BenMcGarry:


erikdesjardins:
Integration test build manually cancelled cause we're about to release (and I don't want to wait); I'll wait for all builds to pass on the changelog commit

PR #3920, opened 2017-01-23
Fix highlightFirstCommenter highlighting all repliers instead of only top level

erikdesjardins:
image


andytuba:
Is that all repliers or usernames that start with the same test?

erikdesjardins:
highlightFirstCommenter doesn't know about usernames. The bug is it thinks "Reply 1" is a top level comment, and therefore that "Reply 3" is a reply from the first commenter.

PR #3931, opened 2017-01-25
filteReddit commentsOpened: Ignore query

larsjohnsen:
In some instances the comment link has a query, which can vary between page loads.

Related https://www.reddit.com/r/RESissues/comments/5pzfva/bug_filtering_by_comments_opened_does_not/

-- not actually an fix


larsjohnsen:
Closing this since it didn't work as expected.

Issue can be partially resolved with #3626 since that connects `commentsOpened` with `newCommentCount`, which separately keeps track of visited comment pages.

Issue #949, opened 2014-04-16
Keyboard shortcut settings option needs (X) "clear shortcut" button


jewel-andraia:
it's practically impossible to clear a hotkey on macros or keyboard nav in settings

PR #3957, opened 2017-01-30
keyboardNav: prompt on ambiguous shortcut

### LINKED ISSUE #3753: Filterline F vs "go to frontpage" F
1. If goMode is enabled and active, F shouldn't show filterline command line
2. If goMode is disabled, ... does filterline or frontpage win?

http://redd.it/5h2mt7

@larsjohnsen

larsjohnsen:
This will show a prompt when encountering a shortcut which has a goMode callback in addition to some other callback.

image

Closes #3753 
Lays groundwork for #949

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -892,40 +892,41 @@ const _commandLookup = _.once(() => {
 	for (const option of [...getActiveCommandOptions(), ...getCommentsLinkKeys()]) {
 		const hash = hashKeyArray(option.value);
 		if (!lookup[hash]) {
-			lookup[hash] = $.Callbacks();
+			lookup[hash] = [];
 		}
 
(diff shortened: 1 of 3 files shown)


lib/modules/keyboardNav.js
```diff
@@ -890,42 +893,85 @@ const getActiveCommandOptions = _.once(() =>
...
+	const hash = hashKeyEvent(e);
+	const options = (_commandLookup()[hash] || [])
+		.filter(v => {
+			if (v.goMode) return module.options.useGoMode.value === goModeActive;
+			else return !goModeActive;
+		});
+
+	if (!options.length) {
+		return false;
+	} else if (options.length > 1 && options.some(v => v.goMode)) {
```
larsjohnsen: The reason why this is limited to `goMode` is that there exist shortcuts coexist on the same key, e.g. `saveLink` and `saveComment`.

PR #3966, opened 2017-02-02
Create new i.redd.it mediahost

### LINKED ISSUE #3882: Add an option to use redditmedia.com instead of redd.it while expanding images
<div class="md">
<p>Let's see if RES can help work around overzealous firewalls by adding an option and an error failover for i.redd.it image posts which asks the reddit api for the post's redditmedia.com preview urls. !GitHub</p>
</div>
<br><br>

https://www.reddit.com/r/Enhancement/comments/5ndixx/feature_request_add_an_option_to_use/dcamhwe/
Even though redd.it domain is great to share images, because the URL is shortened, domains outside the usual .com, .net, .org are often blocked on some public networks, as they are misused by malware disseminators.
Resources to implement the idea:
Post with redd.it image: https://www.reddit.com/r/photoshopbattles/comments/5nc0l2/psbattle_dog_running_in_the_snow/
The preview:images array in post's .json provides the redditmedia.com URL's.
A second expand button option would be really great for this, without changing the redd.it link, that still would be used on sharing the image.
Thanks in advance! Even if the feature is not achievable, be it because it goes against some RES' project future ideas or another reason. But it would be really good to see it implemented. Thanks again.

### LINKED ISSUE #3941: Play i.redd.it/<id>.gif as gifv
RES can ask the reddit api for preview URLs on a post

https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/3882#issuecomment-275475787

:warning: **NSFL** https://www.reddit.com/r/WTF/comments/5qae5z/basketball_players_eye_gets_gouged_out/dcxmm9u/ RES is naively trying to load the OP, i.redd.it/....gif, as an ``

WilliamChiu:
closes #3882 
fixes #3941

### DIFF
--- a/lib/modules/hosts/ireddit.js
+++ b/lib/modules/hosts/ireddit.js
@@ -1,0 +1,36 @@
+/* @flow */
+
+import { Host } from '../../core/host';
+import { DAY, string, Thing, getPostMetadata } from '../../utils';
+import { ajax } from '../../environment';
+
+export default new Host('ireddit', {
+	name: 'ireddit',
+	domains: ['i.redd.it'],
+	options: {
+		useiReddit: {
+			title: 'iredditUseIReddItTitle',
+			description: 'iredditUseIReddItDesc',
+			value: false,
+			type: 'boolean',
+		},
+	},
+	detect: function(mediaurl, element) { //not entirely sure what I'm supposed to return :p
+		const { pathname } = element, thing = Thing.from(element);
+		if (!thing.isPost()) return; 
+		let id = thing.$thing.data('fullname').replace('t3_',''); //To replace with thing.id()
+		let postMetadata = getPostMetadata({ id });
+		postMetadata.then(function (data) {
+			let newmediaurl = data.preview.images[0].source.url; //got the url!
+			console.log(newmediaurl); //this also prints
+			return true; //this does nothing
+		});
+	},
+	async handleLink(href) {
+		console.log(href); //doesn't get logged
+		return {
+			type: 'IMAGE',
+			src: href,
+		};
+	},
+});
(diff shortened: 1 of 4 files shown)


lib/modules/hosts/ireddit.js
```diff
@@ -0,0 +1,27 @@
...
+			title: 'iredditUseIReddItTitle',
+			description: 'iredditUseIReddItDesc',
+			value: false,
+			type: 'boolean',
+		},
+	},
+	detect: (url, thing) => thing && thing.isPost() && thing.id(),
+	async handleLink(href, id) {
+		let postMetadata = await getPostMetadata({ id: id.replace('t3_', '') });
+		newhref = postMetadata.preview.images[0].source.url; //got the url!
```
erikdesjardins: There needs to be two branches here, for static images and gifs (which will give us the .mp4). Something like:

```js
const preview = postMetadata.preview.images[0];
if (preview.variants.mp4) {
    return {
        type: 'VIDEO',
        controls: false,
        loop: true,
        muted: true,
        fallback: preview.variants.gif && preview.variants.gif.source.url,
        sources: [{
            source: preview.variants.mp4.source.url,
            type: 'video/mp4',
        }],
    };
} else {
    return {
        type: 'IMAGE',
        src: preview.source.url,
    };
}
```

lib/modules/hosts/ireddit.js
```diff
@@ -0,0 +1,33 @@
...
+import { getPostMetadata } from '../../utils';
+
+export default new Host('ireddit', {
+	name: 'ireddit',
+	domains: ['i.redd.it'],
+	detect: (url, thing) => thing && thing.isPost() && thing.id(),
+	async handleLink(href, id) {
+		const postMetadata = await getPostMetadata({ id: id.replace('t3_', ''),  });
+		const preview = postMetadata.preview.images[0];
+		console.log(preview);
```
erikdesjardins: Stray `console.log`

PR #3978, opened 2017-02-04
filteReddit: notification when everything is filtered out

larsjohnsen:
It's probably not too uncommon among users to add a filter which causes every post to be hidden, without realizing what they've done or knowing how to remove it.

peek 2017-02-04 14-00

### DIFF
--- a/lib/modules/filteReddit/Filterline.js
+++ b/lib/modules/filteReddit/Filterline.js
@@ -13,6 +13,7 @@ import {
 	unhide,
 	keyedMutex,
 } from '../../utils';
+import { i18n } from '../../environment';
 import filterlineTemplate from '../../templates/filterline.mustache';
 import * as CommandLine from '../commandLine';
 import * as Notifications from '../notifications';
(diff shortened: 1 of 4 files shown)


lib/modules/filteReddit/Filterline.js
```diff
@@ -409,6 +409,41 @@ export default class Filterline {
...
+				notificationID: 'everyThingHidden',
+				header: i18n('filteRedditEmptyNotificationHeader'),
+				message: $('<div>').append(info).append(toggle).get(0),
+				closeDelay: Infinity,
+			});
+		};
+
+		return _.throttle(() => {
+			if (this.getFiltered().length !== this.things.size) {
+				if (Number.isInteger(showId)) clearTimeout(showId);
```
erikdesjardins: Could use `debouncedFunction.cancel()` for this to avoid tracking the timeout id

andytuba:
I wonder if we could combine or also give this treatment to filteReddit's existing "80% of posts are filtered for such-and-such reasons" notification. 

Another tangent: it could be neat to embed a summary of filtered posts just above the NER banner.

larsjohnsen:
@andytuba Filter notifications were removed with the Filterline PR. My thinking was that it should be evident that if posts are not showing that means they are filtered out. I do however think it is worthwhile to show a summary in order to make filtering harder for the user to mess up.

PR #3995, opened 2017-02-07
Set `.visited` on Thing

larsjohnsen:
Fixes http://redd.it/5sljll


lib/css/res.scss
```diff
@@ -1419,7 +1419,7 @@ body.RESScrollLock {
 	z-index: $zindex-autocomplete-dropdown;
 }
 
-.thing .title.visited {
+.thing.visited .title {
```
erikdesjardins: We can possibly remove this rule now, since reddit should have the same thing

PR #3996, opened 2017-02-07
travis: Install yarn in before_install

mc10:
This probably won't make very much of a difference, since yarn packages are being cached, but it'll help with intermittent network issues for new packages.


erikdesjardins:
Yarn internally retries failed requests, so this really shouldn't make a difference. The other install changes are good though

PR #3999, opened 2017-02-07
Feature/single click opener self post

### LINKED ISSUE #3969: Single-Click Opener optionally opens subreddit frontpage on self-posts
https://www.reddit.com/r/Enhancement/comments/5rqfim/feature_request_lc_allow_lc_link_on_self_posts/

EXAMPLE: You see a self post on /r/technology, you click [l+c], two tabs open. One with the self post, and one with the front page of /r/technology.

The [l+c] link when added to self posts would open the selfpost/comments in one tab and the front page of the subreddit that the self post originated from in a separate tab.

This should be an option which is disabled by default.

alexvanolst:
Resolves #3969

~~The unshift instead of push is to maintain the correct order of link or frontpage / comments~~

### DIFF
--- a/lib/modules/singleClick.js
+++ b/lib/modules/singleClick.js
@@ -69,12 +69,16 @@ module.go = () => {
 
 export function openTabs(thing: Thing, focused: boolean) {
 	const link = thing.getPostLink().href;
-	const comments = thing.getCommentsLink().href;
-	const frontpage = thing.getSubredditLink().href;
 	const urls = [link];
 
-	if (thing.isLinkPost()) urls.push(comments);
-	else if (module.options.openFrontpage.value) urls.unshift(frontpage);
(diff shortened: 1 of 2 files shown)


lib/modules/singleClick.js
```diff
@@ -62,9 +67,11 @@ module.go = () => {
 export function openTabs(thing: Thing, focused: boolean) {
 	const link = thing.getPostLink().href;
 	const comments = thing.getCommentsLink().href;
+	const frontpage = thing.getSubredditLink().href;
 	const urls = [link];
 
 	if (thing.isLinkPost()) urls.push(comments);
+	else if (module.options.openFrontpage) urls.unshift(frontpage);
```
andytuba: Should be `module.options.openFrontpage.value`

PR #4008, opened 2017-02-08
Fix disabling modules with no options

erikdesjardins:
https://www.reddit.com/r/RESissues/comments/5sveiv/unable_to_turn_off_browse_by_wheel_wheelbrowse/


lib/modules/settingsConsole.js
```diff
@@ -419,8 +419,8 @@ const drawSettingsConsole = _.once(() => {
 		const enable = !enabled;
 		$(thisToggle).add($moduleButton).add(this.classList)
 			.toggleClass('enabled', enable);
-		updateSaveButton(enable && !_.isEmpty(Modules.get(moduleID).options));
```
erikdesjardins: `updateSaveButton` takes no arguments, it just checks the stage

PR #4012, opened 2017-02-09
ResizeObserver

larsjohnsen:
The API is only available for Chrome under a flag (chrome://flags/#enable-experimental-web-platform-features), and isn't yet implemented for the other browsers. It does however perform significantly better than the current solution.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1,7 +1,6 @@
 /* @flow */
 
 import _ from 'lodash';
-import elementResizeDetectorMaker from 'element-resize-detector';
 import { filter, flow, keyBy, map, sortBy } from 'lodash/fp';
 import { $ } from '../vendor';
 import type {
@@ -505,8 +504,6 @@ function getSelector(isSelfText) {
 	}
 }
 
-const elementResizeDetector = _.once(() => elementResizeDetectorMaker({ strategy: 'object' }));
-
 module.go = () => {
 	createImageButtons();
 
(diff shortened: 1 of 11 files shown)


larsjohnsen:
Chrome 64+ has ResizeObserver enabled by default.

lib/modules/showImages.js
```diff
@@ -1818,73 +1791,45 @@ function makeMediaMovable(media, element, dragInitiater = element) {
...
-			});
-		}
-
-		window.addEventListener('resize', debouncedResize);
-
-		if (prevExpand) return prevExpand();
+	const updateWrapper = ({ width, height }) => {
+		wrapper.style.height = `${horizontalIsVertical ? width : height}px`;
+		// Force immediate reflow in order to get resizeObserver to react faster
+		window._trash_updateWrapperRefowProvocation = wrapper.getBoundingClientRect();
```
erikdesjardins: Debugging?

larsjohnsen: This made the whole process go a bit faster, but that's not the seemingly not case anymore.

larsjohnsen:
@erikdesjardins Were you able to get `resize-observer-polyfill` working? It apparently breaks when run through `transform-dead-code-elimination`. Even after that plugin is removed, it for some reason does not trigger. `resize-observer-lite` however works fine as it simply wraps `element-resize-detector` when ResizeObserver is not available.

erikdesjardins:
Hmm, yeah, I thought it was working at some point but apparently not.

erikdesjardins:
Test failure is spurious

PR #4018, opened 2017-02-11
frameThrottle: Run as ordinary throttle (via setTimeout) if document is hidden

larsjohnsen:
This so that `watchForThing` / `watchForElement` listeners will run even when the page is opened in the background.


erikdesjardins:
I was initially concerned that this could make opening a new tab more expensive, but it doesn't seem to have a noticeable affect on the current tab. In any case, browsers can/do delay timers in background tabs, so it shouldn't be a problem.

PR #4024, opened 2017-02-14
Popular in subreddit manager

andytuba:
haven't actually tested it, but the implementation it's working against is:

If user is logged out, show:

    [Popular](https://www.reddit.com/r/popular) | All | Random

If user is logged in, show:

    [Front](https://www.reddit.com) | [Popular](https://www.reddit.com/r/popular) | All | Random


erikdesjardins:
Ran integration tests locally

PR #4028, opened 2017-02-15
Treat /r/popular like /r/all

erikdesjardins:
i.e. filter subreddits in /r/popular


erikdesjardins:
Ran integration tests locally

PR #4039, opened 2017-02-18
Lots of cleanup

erikdesjardins:
See commit messages for more details.

https://www.reddit.com/r/Enhancement/wiki/support#wiki_safari


lib/modules/pageNavigator.js
```diff
@@ -133,5 +133,5 @@ function showLinkTitle() {
 		}, { rootMargin: '100px 0px 0px 0px' }).observe($submission.get(0));
 	});
 
-	window.addEventListener('wheel', updateWidget, process.env.BUILD_TARGET !== 'safari' ? { passive: true } : false);
+	window.addEventListener('wheel', updateWidget, process.env.BUILD_TARGET !== 'edge' ? { passive: true } : false);
```
BenMcGarry: For my own knowledge, what's this doing?

erikdesjardins: Edge doesn't support [passive event listeners](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md), but it treats `{ passive: true }` as `true`[1], which makes the listener capturing, which does something we don't want (makes the listener capturing)

[1]: this might sound weird but it's fairly logical behaviour

lib/core/profiling.js
```diff
@@ -52,108 +50,23 @@ function setupPerf() {
 	});
 }
 
-type TimingInfo = { [key: string]: { [key: string]: {
-	start: number,
-	end: number,
-} } };
-
 let recordTiming = false;
-let usePerformanceApi = false;
```
andytuba: Oh, only Safari didn't support nice performance timings.

PR #4040, opened 2017-02-18
Accountswitcheri18n

roshkins:
Added extra i18n with snazzy snippet
```
'.source.js':
  'RES-i18n':
    'prefix': 'i18n'
    'body': """
            ${i18n("$2")}${3:
            "$2": {
              "message": "$1"
            \\},
            }"""
```

part of #3755

### DIFF
--- a/lib/modules/accountSwitcher.js
+++ b/lib/modules/accountSwitcher.js
@@ -4,7 +4,7 @@ import _ from 'lodash';
 import { $ } from '../vendor';
 import { Module } from '../core/module';
 import { Alert, formatDate, formatDateDiff, getUserInfo, loggedInUser } from '../utils';
-import { ajax, multicast } from '../environment';
+import { ajax, multicast, i18n } from '../environment';
 import * as BetteReddit from './betteReddit';
 import * as CommandLine from './commandLine';
 import * as Notifications from './notifications';
@@ -185,18 +185,18 @@ module.go = () => {
 	}
 
 
-	CommandLine.registerCommand(/^sw$/, 'sw [username] - switch users to [username]',
+	CommandLine.registerCommand(/^sw$/, 'sw [username] - ${i18n("accountSwitcherCliHelp")}',
 		(command, val) => {
 			const found = findMatchingAccount(val, true);
 			if (val && found) {
-				return `Switch to username: ${found[0]}`;
+				return `${i18n("accountSwitcherCliSwitchToUsernamePrompt")} ${found[0]}`;
 			} else {
-				return 'Switch to username: ...';
+				return '${i18n("accountSwitcherCliSwitchToUsernamePrompt")} ...';
 			}
 		},
 		(command, val) => {
 			if (!val.length) {
-				return 'No username specified.';
+				return '${i18n("accountSwitcherCliNoUsername")}';
 			}
 
 			const found = findMatchingAccount(val, true);
@@ -232,7 +232,7 @@ async function populateUser(container) {
 
 	if (module.options.showUserDetails.value) {
 		const { data: userInfo } = await getUserInfo(username).catch(e => {
-			console.error('Error loading userinfo for', username, e);
+			console.error('${i18n("accountSwitcherCliUserinfoError")}', username, e);
 			return { data: {} };
 		});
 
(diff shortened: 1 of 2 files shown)


lib/modules/accountSwitcher.js
```diff
@@ -248,7 +248,7 @@ async function populateUser(container) {
 			if (userInfo.gold_expiration) {
 				const today = new Date();
 				const expDate = new Date(userInfo.gold_expiration * 1000); // s -> ms
-				$gilded.attr('title', `Until ${formatDate(expDate)} (${formatDateDiff(today, expDate)})`);
+				$gilded.attr('title', `${i18n("accountSwitcherGoldUntil")} ${formatDate(expDate)} (${formatDateDiff(today, expDate)})`);
```
erikdesjardins: Nit: interpolation here too; in this case translators will have no idea what `"Until "` means. `"Until $1 ($2)"` is a bit better.

You can also add [descriptions](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/26703c80c01e27cd888e0bc180514ba3e5511172/locales/locales/en.json#L2835), which I would do here because it's fairly ambiguous (though they're certainly not required for most translations)

lib/modules/accountSwitcher.js
```diff
@@ -341,23 +341,20 @@ async function switchTo(user) {
 		Notifications.showNotification({
 			type: 'error',
 			moduleID: 'accountSwitcher',
-			message: 'Could not switch accounts. Reddit may be under heavy load. Please try again in a few moments.',
+			message: '${i18n("accountSwitcherAccountSwitchError")}',
```
erikdesjardins: Ditto, no need to interpolate the whole string

lib/modules/accountSwitcher.js
```diff
@@ -341,23 +341,20 @@ async function switchTo(user) {
...
 		});
 		throw new Error(response);
 	}
 
 	if (response.includes('WRONG_PASSWORD')) {
 		Alert.open(`
-			Could not log in as <b>${username}</b> because either the username or password is wrong.
-			<p>Check your settings?</p>
+			${i18n("accountSwitcherLoginError", username)}
```
erikdesjardins: Ditto

lib/modules/accountSwitcher.js
```diff
@@ -341,23 +341,20 @@ async function switchTo(user) {
...
 		`, { cancelable: true }).then(manageAccounts);
 
 		switchedAccount(false);
 	} else if (response.includes('RATELIMIT')) {
 		Alert.open(`
-			Could not log in as <b>${username}</b> because reddit is seeing too many requests from you too quickly.
-			Perhaps you've been trying to log in using the wrong username or password?
-			<p>Check your settings?</p>
+			${i18n("accountSwitcherRateLimitError", username)}
```
erikdesjardins: Ditto

lib/modules/accountSwitcher.js
```diff
@@ -417,13 +414,13 @@ function _notifySwitchedAccountElsewhere() {
 	}
 
 	let message = username ?
-		`You switched to /u/${username}.` :
+		`${i18n("accountSwitcherUserSwitched",username)}` :
```
erikdesjardins: Ditto

lib/modules/accountSwitcher.js
```diff
@@ -417,13 +414,13 @@ function _notifySwitchedAccountElsewhere() {
...
 		'You have been logged out.';
 
 	if (hasDraft) {
-		message += ` However, you haven't finished posting on this page as /u/${loggedInUser() || ''}.`;
+		message += ` ${i18n("accountSwitcherDraft", loggedInUser() || '')}`;
```
erikdesjardins: FYI interpolation is fine here because prepending a space is awkward for translating (your code is correct)

locales/locales/en.json
```diff
@@ -189,6 +189,49 @@
...
+    "accountSwitcherGoldUntil": {
+      "message": "Until"
+    },
+
+    "accountSwitcherAccountSwitchError": {
+      "message": "Could not switch accounts. Reddit may be under heavy load. Please try again in a few moments."
+    },
+
+    "accountSwitcherLoginError": {
+      "message": "Could not log in as <b>$1</b> because either the username or password is wrong.
```
erikdesjardins: We want to avoid HTML in translations because it's awkward to translate (and because it's unsafe). In this case, since these are rare error messages, we can go without the extra formatting and just use `\n`.

Also, JSON strings can't be multiline, so this (and below) is invalid syntax.

locales/locales/en.json
```diff
@@ -189,6 +189,49 @@
...
+      <p>Check your settings?</p>"
+    },
+
+    "accountSwitcherRateLimitError": {
+      "message": "Could not log in as <b>$1</b> because reddit is seeing too many requests from you too quickly.
```
erikdesjardins: Ditto

locales/locales/en.json
```diff
@@ -3556,5 +3603,7 @@
     "penaltyBoxFeaturesPenalty": {
         "message": "penalty",
         "description": "How heavily the feature is penalized (0-100)"
-    }
+    },
```
erikdesjardins: JSON doesn't support trailing commas

PR #4043, opened 2017-02-20
Prevent RES Diagnostics re-dumping

### LINKED ISSUE #4038: Prevent RES Diagnostics re-dumping if a slight change is made
Currently when RES dumps the diagnostic data into a post it keeps checking if its there, if you add a space after any of the settings or lines directly above `Night mode:` it will re-dump it causing it to appear multiple times.

Example: https://www.reddit.com/r/Enhancement/comments/5unllm/feature_request_autofindinpage/

https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/submitIssue.js#L77

lhofmann:
Fixes #4038


BenMcGarry:
Thanks!

PR #4052, opened 2017-02-21
Update imgur.js

ATMarcks:
Added '.jpeg' inside the _handleimage function; previously expanding ".jpeg" file types would also show the original filename above the image

before
after

### DIFF
--- a/lib/modules/hosts/imgur.js
+++ b/lib/modules/hosts/imgur.js
@@ -137,7 +137,7 @@ export default new Host('imgur', {
 		function _handleImage(hash, url) {
 			const [,, extension] = hashRe.exec(url) || [];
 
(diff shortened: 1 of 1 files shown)


erikdesjardins:
That's not quite true; the filename is displayed (correctly) because it's [the imgur title of the image](http://imgur.com/YbRJDj2). Though for consistency with other direct links, we shouldn't hit the API for `.jpeg`, so this patch still makes sense.

ATMarcks:
Thanks for the clarification; I was a bit perplexed at the filename bit because I use a third party program that uploads files to imgur and automatically copy direct links to the clipboard (apparently it automatically titles them after the filename as well--I now realize that). Anyway, I noticed the inconsistency and decided to dig in a bit.

Hooray for my first pull request 😄😎

PR #4056, opened 2017-02-22
Replace mustache templates with template strings

erikdesjardins:
For easier i18n, typechecking, and (ostensibly) simplicity.

~~Not sure if I'm happy with this yet.~~


lib/modules/filteReddit/Filter.js
```diff
@@ -56,7 +55,11 @@ export default class Filter<Parsed: { [key: any]: any }> {
 	}
 
 	createElement() {
-		this.element = $(filterlineFilterTemplate(this))[0];
+		this.element = $(`
```
andytuba: Maybe we should have a `module.utils.html(markup: string)` that abstracts away `$(markup)[0]`? We could probably even skip jQuery (unless it adds security) with something like `x = document.createDocumentFragment(); x.innerHTML = markup; return x`

andytuba:
I'd be in favor of replacing mustache templates with native templates. Handlebars certainly provides  some niceties, but sometimes the syntax is more confusing than `${vanillaJavascript}`.

erikdesjardins:
Yeah, it'll default to escaping interpolated strings by default, then you can mark strings as safe (like Handlebars' triple brace) with `string.safe`

PR #4059, opened 2017-02-22
[very experimental] webpack-rollup-loader

erikdesjardins:
This unholy abomination sandwiches Rollup scope hoisting and tree shaking between Webpack bundling and module resolution. But it makes bundles smaller and init faster.


erikdesjardins:
This probably won't be feasible for a while (if ever), so I'm closing this for now.

PR #4060, opened 2017-02-22
Move translations to background page; cache on foreground in localStorage

erikdesjardins:
This eliminates the effect of translations on bundle size (and hence parse/compile performance), at the cost of slow init on the first (ever) pageload on each domain or when switching languages.

Each dictionary is 64-100k, so we have quite a bit of localStorage headroom.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -12,7 +12,7 @@
 [options]
 module.use_strict=true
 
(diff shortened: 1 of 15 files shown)


erikdesjardins:
Build failure is due to coverage reporting, presumably due to AWS and recent (possibly unrelated) Travis issues

PR #4061, opened 2017-02-22
filter expando: Update reference to getEntryExpando

larsjohnsen:


erikdesjardins:
Sorry about that, should have looked for more usages when updating the old branch :/

PR #4066, opened 2017-02-23
keyboardNav: Don't restrict linkAnnotations to comments	

larsjohnsen:
Tested on all page types that SelectedEntry runs on.


erikdesjardins:
Another coverage-only failure due to Travis/Coveralls infra issues (still s3?)

PR #4067, opened 2017-02-23
fixHideLinks: Option for hiding link instantaneously

larsjohnsen:


andytuba:
Even faster than reddit itself! :rocket:

PR #4074, opened 2017-02-26
Wikipedia: Fix fragment and schemeless links

larsjohnsen:
Fixes

https://www.reddit.com/r/RESissues/comments/5vjshc/bug_links_to_anchors_in_wikipedia_expandos/
https://www.reddit.com/r/RESissues/comments/5uk8y8/bug_geohack_links_broken_in_wikipedia_expandos/

### DIFF
--- a/lib/modules/hosts/wikipedia.js
+++ b/lib/modules/hosts/wikipedia.js
@@ -55,8 +55,8 @@ export default new Host('wikipedia', {
 		const $wikiData = $('<div>', { html: data.parse.text['*'] });
 		// Remove unwanted sections
 		$wikiData.find('.metadata, .hatnote, .mw-editsection, .reference, .references').remove();
-		// Make all relative links direct
-		$wikiData.find('a[href^="/"]').attr('href', (i, old) => `https://${language}.wikipedia.org${old}`);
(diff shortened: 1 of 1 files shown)


lib/modules/hosts/wikipedia.js
```diff
@@ -56,7 +56,12 @@ export default new Host('wikipedia', {
 		// Remove unwanted sections
 		$wikiData.find('.metadata, .hatnote, .mw-editsection, .reference, .references').remove();
 		// Make all relative links direct
-		$wikiData.find('a[href^="/"]').attr('href', (i, old) => `https://${language}.wikipedia.org${old}`);
+		$wikiData.find('a')
```
erikdesjardins: The [URL constructor](https://developer.mozilla.org/en/docs/Web/API/URL/URL) is likely what we need here, e.g. ```new URL(old, `https://${language}.wikipedia.org/wiki/${article}`).href```; it can resolve any url relative to a base URL

larsjohnsen: Appears to be a much better solution, thanks.

erikdesjardins:
Thanks!

PR #4078, opened 2017-02-27
voteEnhancements: remove misleading post vote estimation

erikdesjardins:
image

~~Since the uncertainty becomes extreme as the upvote percentage approaches 50%.~~

Edit: Actually, that still appears to have an unwarranted amount of precision.

We know that the one post had ~12k points and 51% upvotes indicated, but only ~50k votes overall. This means either upvotes are weighted ~1.5x higher than downvotes or upvote percentage is skewed by ~10% (probably a combination of both).

Either way, trying to calculate vote totals could definitely be called "misleading", even for posts with a more usual vote percentage. (They're so far from the real vote counts that no remotely accurate conclusions can be drawn.)

So we're just going to remove them.

### DIFF
--- a/locales/locales/en.json
+++ b/locales/locales/en.json
@@ -2420,24 +2420,6 @@
 		"message": "your votes for $1: $2",
 		"description": "As in `your votes for username: 42`"
 	},
-	"voteEnhancementsUpvote": {
-		"message": "upvote"
-	},
-	"voteEnhancementsUpvotes": {
-		"message": "upvotes"
-	},
-	"voteEnhancementsDownvote": {
-		"message": "downvote"
-	},
-	"voteEnhancementsDownvotes": {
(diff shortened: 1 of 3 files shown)


locales/locales/en.json
```diff
@@ -2390,6 +2390,9 @@
 		"message": "your votes for $1: $2",
 		"description": "As in `your votes for username: 42`"
 	},
+	"voteEnhancementsEstimated": {
+		"message": "±$1 (estimated from fuzzed score and upvote percentage)"
```
erikdesjardins: This could imply greater precision than there actually is, should I change it to `± >1234 ...` or similar?

andytuba: I'd opt for the numeral approach.

andytuba:
👍

PR #4080, opened 2017-02-28
Beam Mediahost

BenMcGarry:


lib/modules/hosts/beam.js
```diff
@@ -0,0 +1,19 @@
+/* @flow */
+
+import { Host } from '../../core/host';
+
+export default new Host('beam', {
+	name: 'Beam',
+	domains: ['beam.pro'],
+	logo: 'https://beam.pro/_latest/assets/favicons/favicon-32x32.png',
+	detect: ({ pathname }) => (/(\w+)$/).exec(pathname),
```
erikdesjardins: Can be `/^\/(\w+)$/` since the pathname includes the leading slash (`'/stream_name'`)

BenMcGarry:
No autoplay param at this time but coming soon.

PR #4081, opened 2017-03-01
Addiere bildgur Unterstuetzung

dasJ:


erikdesjardins:
What is the use-case for bildgur.de instead of imgur.com?

AFAICT it redirects to imgur without proxying--in which case it's merely a link obfuscator, which doesn't seem like something we want to support.

dasJ:
Yes, it already works, sorry. I was mistaken

Issue #3919, opened 2017-01-22
Edge bugs with arrow function destructuring


BenMcGarry:
Revisit at Creators to see if issue has been resolved to reverse commit ~~#3918~~ #4102

External Reference: https://github.com/Microsoft/ChakraCore/issues/2401 and https://github.com/Microsoft/ChakraCore/issues/2606

BenMcGarry:
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/93fec936a1da5cca5403586b1f0229c1cf17ec60/lib/utils/alert.js#L43 Use before Dec

BenMcGarry:
@erikdesjardins Looks like this has been resolved on 16237

erikdesjardins:
Alright, I'll add it to a milestone and we'll keep pushing it back until Edge 16 is stable

BenMcGarry:
Edge 16 has October 17th release.

BenMcGarry:
#4102 can now be reversed.

PR #4102, opened 2017-03-07
Switch to a targeted workaround for Chakra "use before declaration" bug

erikdesjardins:
Part of #3919, reverts #3918 (in practice, but not in spirit)


BenMcGarry:
Ref Microsoft/ChakraCore#2606

PR #4107, opened 2017-03-08
userHighlight: change highlight color when collapsed

lhofmann:
This fixes a bug in userHighlight, where collapsed posts were still showing the highlighted color instead of gray. 
Originally, the selector used was `.tagline .collapsed`, when it should be `.collapsed .tagline`.

I also refactored a small part of the code.


erikdesjardins:
Looks good, thanks!

PR #4130, opened 2017-03-21
Fix shortcut bar left scrolling by using integer value

mikeparas:
Relevant issue: With long shortcut bars, the left scroll button appeared to be unresponsive. The math being done to find the shift distance could create a possible fractional value (0 < marginLeft < 1). Since marginLeft would have to be <= 0 to trigger the shift, it would fail in this case.

Tested in browser: Chrome 57.0.2987.110 (Arch Linux)

### DIFF
--- a/lib/modules/subredditManager.js
+++ b/lib/modules/subredditManager.js
@@ -978,14 +978,9 @@ function redrawSubredditBar() {
 		shortCutsLeft.textContent = '<';
 		shortCutsLeft.addEventListener('click', () => {
 			const firstChild: HTMLElement = (shortCutsContainer.firstChild: any);
-			let marginLeft = firstChild.style.marginLeft;
-			marginLeft = parseInt(marginLeft.replace('px', ''), 10);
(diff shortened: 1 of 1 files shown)


lib/modules/subredditManager.js
```diff
@@ -901,6 +901,7 @@ function redrawSubredditBar() {
 
 			const shiftWidth = $('#RESShortcutsViewport').width() - 80;
 			marginLeft += shiftWidth;
+			marginLeft = parseInt(marginLeft, 10);
```
andytuba: How about `Math.floor(marginLeft)` instead of abusing `parseInt` to truncate?

mikeparas: That does seem better. Thanks for the suggestion!

erikdesjardins:
Integration test failure is spurious (/due to test machine starvation); ran tests locally

erikdesjardins:
Thanks!

PR #4131, opened 2017-03-21
add "rel=0" to end of youtube embeds to get rid of related videos

magicwizard8472:
[https://redd.it/60jq6g](https://redd.it/60jq6g)

just appends "rel=0" to end of youtube embed links so the related videos bar doesn't come up when the video gets paused.

### DIFF
--- a/lib/modules/hosts/youtube.js
+++ b/lib/modules/hosts/youtube.js
@@ -22,7 +22,7 @@ export default new Host('youtube', {
 			(/watch|embed|v/i).exec(split[0]) && split[1]; // watch/embed/v
 	},
 	handleLink(href, path) {
-		let src = insertParams(`https://www.youtube.com/embed/${path}`, { version: 3 });
+		let src = insertParams(`https://www.youtube.com/embed/${path}`, { version: 3, rel: 0 });
 
 		const params = getUrlParams(href);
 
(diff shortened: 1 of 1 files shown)


andytuba:
Thanks for implementing [this request]( https://www.reddit.com/r/Enhancement/comments/60jq6g/feature_requesttoggle_for_youtubes_related_videos/)! Looks fine to me, probably not worth adding in an option to show related videos -- people can click through for that.

lib/modules/hosts/youtube.js
```diff
@@ -53,7 +53,7 @@ export default new Host('youtube', {
 		return {
 			type: 'IFRAME',
 			embed: src,
-			embedAutoplay: insertParams(src, { autoplay: 1 }),
+			embedAutoplay: insertParams(src, { autoplay: 1, rel: 0 }),
```
erikdesjardins: This needs to be done on line 25; this will only affect videos with autoplay (unless yt only shows related videos when you enable autoplay? but I don't think that's the case)

magicwizard8472: Ah sorry, must've missed that when looking for a place to insert it.

magicwizard8472:
Welp, not sure what went wrong. Was that change enough to make Travis' check fail?

erikdesjardins:
Thanks!

PR #4145, opened 2017-03-30
remove environment/alert.js

erikdesjardins:
Tested in browser: firefox 54.0a2

This was only used by the old Firefox backend, so is unnecessary after #4116


erikdesjardins:
Integration test failure is spurious; ran tests locally

PR #4147, opened 2017-03-30
Test in Firefox beta for (presumably) more stability

erikdesjardins:
All recent test builds are failing (crashing) in Firefox; I can't tell if it's due to instability in Sauce Labs infra or Firefox dev, so let's see if this changes anything.


erikdesjardins:
Right, you can't sideload extensions in beta. Who knows if the extension even works until you "test" in production? Thanks Firefox! 😠

PR #4167, opened 2017-04-04
Remove explicit yarn installation

### LINKED ISSUE #3624: Use Travis CI native yarn support when available
[Travis PR](https://git.io/vXQPe)

Edit: ^ won't work for us because we need the install stage to run `--pure-lockfile` (or `--frozen-lockfile`). Eventually there will be a Travis image with Yarn preinstalled which we'll be able to use.

erikdesjardins:
Closes #3624

It appears that Travis now installs yarn before the `install`/`beforeinstall` phases, so we aren't overwriting it anymore.


mc10:
Looks like https://github.com/travis-ci/travis-build/commit/0f4b9f68dcfe1c215827b188dd77c747c797eb5a and https://github.com/travis-ci/travis-build/commit/438167e509b5b00fb3cb6b414adc7ee8756a1e9f are the relevant changes.

Issue #4128, opened 2017-03-20
Post to Profile / desktop2x compatibility


andytuba:
https://www.reddit.com/r/modnews/comments/60i60u/tomorrow_well_be_launching_a_new_posttoprofile/

I'll fill in technical details soon.

PR #4174, opened 2017-04-09
Support r2-rendered post-to-profile comments page

andytuba:
Relevant issue: #4128 
Tested in browser:  Chrome

Treats post-to-profile comments pages as regular comments pages, and some incidental cleanup.

### DIFF
--- a/lib/modules/commentDepth.js
+++ b/lib/modules/commentDepth.js
@@ -2,7 +2,7 @@
 
 import { $ } from '../vendor';
 import { Module } from '../core/module';
-import { insertParams, regexes, Thing } from '../utils';
+import { execRegexes, insertParams, regexes, Thing } from '../utils';
 
 export const module: Module<*> = new Module('commentDepth');
 
--- a/lib/modules/commentDepth.js
+++ b/lib/modules/commentDepth.js
@@ -70,12 +70,12 @@ module.go = () => {
 		// no need to proceed if depth already exists in the query string
 		if (search.match(/[?&]depth=/)) return;
 
-		if (pathname.match(regexes.commentPermalink)) {
+		if (regexes.commentPermalink.test(pathname)) {
 			if (!module.options.commentPermalinks.value) return;
 			if (!module.options.commentPermalinksContext.value && search.match(/[?&]context=/)) return;
 		}
 
-		const matches = pathname.match(regexes.comments);
+		const matches = execRegexes.comments(pathname);
 		if (!matches) return;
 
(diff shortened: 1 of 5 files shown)


lib/modules/commentDepth.js
```diff
@@ -70,12 +70,12 @@ module.go = () => {
...
 		if (search.match(/[?&]depth=/)) return;
 
-		if (pathname.match(regexes.commentPermalink)) {
+		if (regexes.commentPermalink.test(pathname)) {
 			if (!module.options.commentPermalinks.value) return;
 			if (!module.options.commentPermalinksContext.value && search.match(/[?&]context=/)) return;
 		}
 
-		const matches = pathname.match(regexes.comments);
+		const matches = execRegexes.comments(pathname);
```
mc10: `execRegexes` should be imported.

lib/utils/location.js
```diff
@@ -67,11 +67,20 @@ export const regexes: { [key: string]: RegExp } = {
 };
 /* eslint-enable key-spacing */
 
+export const execRegexes: { [k: string]: (path: string) => undefined | string[] } = {
```
mc10: I think you need `?string[]` instead of `undefined | string[]`.

lib/utils/location.js
```diff
@@ -67,11 +67,20 @@ export const regexes: { [key: string]: RegExp } = {
...
+		return match;
+	},
+};
+
 export type PageType = 'wiki' | 'search' | 'profile' | 'stylesheet' | 'modqueue' | 'subredditAbout' | 'comments' | 'commentsLinklist' | 'liveThread' | 'inbox' | 'submit' | 'account' | 'prefs' | 'linklist';
 
 export const pageType = _.once((): PageType => {
 	const defaultPageType = 'linklist';
-	const pageTypes = ['wiki', 'search', 'profile', 'stylesheet', 'modqueue', 'subredditAbout', 'comments', 'commentsLinklist', 'liveThread', 'inbox', 'submit', 'account', 'prefs'];
+	const pageTypes = ['wiki', 'search', 'stylesheet', 'modqueue', 'subredditAbout', 'comments', 'commentsLinklist', 'profile', 'liveThread', 'inbox', 'submit', 'account', 'prefs'];
```
erikdesjardins: ```andytuba [7:17] i’m only a little worried about changing up the order of pagetype checks```

Yeah as long as no page types are added or removed it shouldn't really change anything (but we do rely heavily on the fallback to `linklist`)

PR #4185, opened 2017-04-21
Move IntersectionObserver polyfill to firefox-only

erikdesjardins:

...since Chrome and Edge support it.

### DIFF
--- a/firefox/manifest.json
+++ b/firefox/manifest.json
@@ -23,6 +23,7 @@
 	},
 	"background": {
 		"scripts": [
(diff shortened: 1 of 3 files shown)


firefox/manifest.json
```diff
@@ -23,6 +23,7 @@
 	},
 	"background": {
 		"scripts": [
+			"{{./firefox.entry.js}}",
```
erikdesjardins: Technically this doesn't need to be on the background, but its tiny, and this is consistent with `edge.entry.js`

PR #4190, opened 2017-04-23
Disable opera beta deployment

erikdesjardins:
It turns out there's no way to submit an Opera extension without review (unlike Firefox), and we now know that release automation works, so disable this for the moment to not inundate the reviewers.


erikdesjardins:
skipping integration tests as this doesn't change any runtime code...

PR #4191, opened 2017-04-23
Filterline refresh

### LINKED ISSUE #2977: Custom filter:  DOM attribute value
"Attribute: thing/entry's attribute ___ [default:class] [equal | contains | does not equal | does not contain ] ___"

for future features or workarounds to broken things like nsfw

### LINKED ISSUE #3895: "Post thing has class..." custom filter

### LINKED ISSUE #4148: (RES for Chrome) The order of the fIlters in filterline is not maintained (they are sorted alphabetically)
- My browser: Chrome 57 in win10
- My RES version: 5.4.3

Order of the fIlters in filterline is not maintained after you close a reddit.com tab

STR
- open https://www.reddit.com/r/CrackWatch/
- add these flair filters: a `Humor`, `Repack`, `Discussion`, `NFO`
Notice that their order is like this:

- Now close the reddit.com tab, open another tab and load reddit.com. Here is how the filters are displayed now:

The issue doesn't occur in Firefox 52 with RES 5.4.3

larsjohnsen:
- The various filters are split into classes (inherits `Case`).
- Filterline `Filter` points to the case instead of absorbing it
- Allows async CustomFilter: CustomFilter now supports `expando`, `isVisited`, and `commentsOpened`
- `ondemand` CustomFilter which are addable to the Filterline via the dropdown
- Improved conditions validations (invalid filters won't be saved)
- ExternalFilter activation toggles in the dropdown
- Remove option for whether to parse string as RegEx — most text inputs support RegEx now; inputs not written as a RegEx will be escaped
- New cases "userVoteWeight" and "selector", misc improvements on other filters
- Adding new / hovering filters open an interactive prompt
- Adds button to temporarily disable all filtering

Infocard when hovering a filter:
image

Optimizations:
- Some cases accepts multiple values into `evaluate` in order to avoid overhead
- Cases are ordered by estimated execution speed
- Group cases are resolved by evaluating browseContext cases, extracting single cases, and merging parts.

Performance is about the same, but some of the load is shifted into `beforeLoad` due to precomputation.

Closes #4148 
Closes #3895 
Closes #2977

Tested in browser: Chrome 61, Firefox 54, Edge 15


erikdesjardins:
Deferring to 5.8.0 because the 5.6.0 release is on the horizon, but this sounds really good.

larsjohnsen:
I'm working on a infocard which appears when hovering on a Filter:
image

For moment is the builder read-only, but I hope to get it interactive fairly soon.

lib/css/modules/_filteReddit.scss
```diff
@@ -135,31 +173,28 @@ body.hideOver18 {
...
-			float: right;
-			content: attr(matches-count);
+			right: 0;
 		}
 	}
 
 	&-other {
 		margin: 2px 4px;
 		line-height: 16px;
+		display: grid;
```
erikdesjardins: NTS: grid layout may not work quite right in Edge, and/or may need more Autoprefixer configuration to get it to add `-ms-` prefixes, I'll have to check

larsjohnsen:
Interactive conditions:
peek 2017-06-15 18-00

larsjohnsen:
@erikdesjardins Could you review this?

It would be nice to have Flow associate a given `{ type }` with its respective `Case` class, but I have not been able to figure that out. The lack of this is most apparent when manipulating `Group`, and the `any` annotation is used more often than I would like.

larsjohnsen:
Namechange since the scope of this PR has steadily increased.

darkred:
First of all, thanks a lot for your work.

I just tried 5.9.5beta in Firefox and I find the use of the regex itself as filter titles, confusing:


**What about offering an option to use the string inside the regex as filter title, like in the previous versions?**
_For reference, the full regex can already be displayed inline the filtered posts, via toggling the "Show filter reason" option, whenever the user wants to check the applied filters._
example  : `link Flair NFO` 
instead of `link Flair /^NFO$/i`
e6f341a4-159b-11e7-87ec-798eaa470690

darkred:
First of all, thanks a lot for your work.

I just tried 5.9.5beta in Firefox and I find the use of the regex itself as filter titles, confusing, i.e. I find it difficult to distinguish the filters now:


**What about displaying just the _string_ , when the filter is just a string, like in the previous version?**
In the following example, I have only entered 'NFO' in the filter box ([screenshot](https://user-images.githubusercontent.com/723651/34642285-4d74ba6a-f319-11e7-8d32-33821b97e3ce.jpg)) :
example  : `link Flair NFO` 
instead of `link Flair /^NFO$/i`
e6f341a4-159b-11e7-87ec-798eaa470690
_For reference, the full regex can already be displayed next to each filtered post, via toggling the "Show filter reason" option, whenever the user wants to check the applied filters._

darkred:
Thanks a lot @larsjohnsen for implementing my suggestion!

PR #4194, opened 2017-04-23
Improve featureTip—dailyTips integration 

larsjohnsen:
FeatureTips are now seamlessly included with dailyTips / tips & tricks.

Tested in Chrome 58


erikdesjardins:
This is better, thanks!

PR #4201, opened 2017-04-27
showImages: Reset certain parent styles for expandos

### LINKED ISSUE #4199: Strikethrough and italic formatting of links falls through to expando contents
- My browser (copy from https://www.whatismybrowser.com): Chrome 57 on Windows 10
- My RES version: v5.4.3
- Am I in reddit beta (yes/no): Yes

If a link to a wikipedia article is italicised or struck through, all the body text in the expando will also have that formatting.

Eg, `~~[Struckthrough](https://en.wikipedia.org/wiki/Software_bug)~~` will produce:

image

and `*[Italicised](https://en.wikipedia.org/wiki/Software_bug)*` will produce:

image

This does not occur if the formatting is entirely contained in the link text, as in `[~~Internal strikethrough~~](https://en.wikipedia.org/wiki/Software_bug)`, and does not appear to occur for bold (**), superscript (^) or headlined (#) links.

*Edit:* On further investigation, formatting appears to fall through to the contents of any expando, such as flickr & deviantart titles and imgur album text.

mc10:
Fixes #4199. Need some confirmation that changing to `inline-block` won't cause undesired layout issues.

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -194,9 +194,18 @@ img {
 	}
 }
 
+.res-expando-box-wrapper {
+	display: block; // Allow res-expando-box to be inline-block
+}
+
 .res-expando-box {
-	display: block;
+	display: inline-block; // inline-block resets parent strikethrough
 	clear: left; // Necessary on some custom stylesheets where .entry .flat-list has no height
+
+	// Reset a few possible parent styles
+	font-style: normal; // Italics
+	font-weight: normal; // Bold
+	white-space: normal; // Whitespace nowrap
 
 	.usertext-body {
 		user-select: text; // Allow selection if is a text expando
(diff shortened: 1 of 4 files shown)


larsjohnsen:
You are correct in that `inline-block` on `.res-expando-block` is undesirable:
image

To avoid that, wrapping `.res-expando-block` in a `div` will probably suffice.

larsjohnsen:
Additional style that needs to be reset: https://www.reddit.com/r/RESissues/comments/69l6v0/popups_are_cut_off_when_using_search/

erikdesjardins:
I changed it to add an inner element instead without changing the existing `.res-expando-box`, which makes me much less nervous about potential unintended consequences. (There were a few things subtly broken before: forced break after expando icons, slightly different spacing above, etc.)

PR #4213, opened 2017-05-06
Make integration tests more resilient to update notifications

erikdesjardins:

PR #4215, opened 2017-05-06
break up i18n test into multiple test cases, fix show images button test

erikdesjardins:
It seems to have a really hard time loading each successive page for some reason, so try breaking it up.

And show images button seems to have trouble ref #4213


erikdesjardins:
Remaining test failure is spurious, still...

PR #4220, opened 2017-05-07
new feature "hide comment buttons" for styletweaks module

br4nnigan:
Will simply hide the "permalink source embed..." buttons until you highlight the post. So you can focus on the essentials, the comments. The repetative button bar is a huge waste of space imo. If you want to use a button you'll have to interact with the post anyway, so it's not really a nuisance. Subtle transition. Does not interfere with any subreddit styles that I'm aware of.

Tested in browser: Firefox 53.0.2


honestbleeps:
Hey, thanks for contributing! I want to pre-emptively say we really do appreciate anyone who contributes code / idas to RES.

I do have a couple of concerns here:

- this is a pretty niche need that could be easily satisfied with a CSS snippet, so I'm not sure how I feel about adding it to our ever growing list of toggles/options. A couple of years ago this wouldn't have been a real concern, but RES has grown really huge and we have to be careful about "bloat"...

- if we did decide to approve it, we should definitely not have the default value of this be `true` for 3 million users who likely do not want this on by default.

I'll defer to other team members opinions if there's disagreement on whether or not this should be included in RES rather than making an executive decision...

br4nnigan:
> this is a pretty niche need 

personally I find it a major UI improvement

> that could be easily satisfied with a CSS snippet 

I used this with the Stylish addon for quite some time, but how many users are even aware of the possibility? The module just takes this snippet and adds an option.

> default value of this 

Yeah true, I also had that concern. 

Anyway, take it or leave it ;)

erikdesjardins:
I'm going to have to agree with honestbleeps that this is a bit too niche, but thanks for the suggestion!

FWIW RES does have its own CSS snippet loader (search the settings console for "snippets"), so users wouldn't need to install stylish.

PR #4221, opened 2017-05-07
Revenge of the Firefox Storage APIs 3: Tokyo Drift (tag sharding)

erikdesjardins:
Relevant issue: https://www.reddit.com/r/RESissues/comments/69ljqu/firefox_reddit_sluggish_and_unstable_since_res/
Tested in browser: Chrome 60, Firefox 54

`chrome.storage.local.set` is _absurdly_ slow in Firefox WebExtensions (the profile below would take just 30ms with `JSON.stringify`)

image

To mitigate this, store tags under individual keys, so all tags don't need to be saved at once.

As a bonus, we also don't need to _load_ all tags at once, bringing init improvements to all browsers.

### DIFF
--- a/browser/foreground.js
+++ b/browser/foreground.js
@@ -40,10 +40,16 @@ const set = (key, value) => _set({ [key]: value });
 const _get = apiToPromise((keys, callback) => chrome.storage.local.get(keys, callback));
 const get = async (key, defaultValue = null) => (await _get({ [key]: defaultValue }))[key];
 
+const _delete = apiToPromise((keys, callback) => chrome.storage.local.remove(keys, callback));
+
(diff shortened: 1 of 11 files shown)


BenMcGarry:
I think someone needs to lay off the films a bit...

Ajedi32:
Is there a Firefox issue open for this on [Bugzilla](https://bugzilla.mozilla.org/)?

erikdesjardins:
Should be covered by
https://bugzilla.mozilla.org/show_bug.cgi?id=1320186 and
https://bugzilla.mozilla.org/show_bug.cgi?id=1319474

and may be (somewhat) remedied by
https://bugzilla.mozilla.org/show_bug.cgi?id=1277612

PR #4223, opened 2017-05-07
Migrator to remove old tag object

erikdesjardins:
Split off from #4221 so that the latter is non-destructive (in case something goes wrong with migration).


erikdesjardins:
I've seen enough people running 5.6.1, who were either affected by the slowdown (meaning they have many tags) without widespread reports of deleted tags, or quite obviously still have tags ([e.g.](https://www.reddit.com/r/RESissues/comments/69v2x3/my_user_tags_dashboard_has_ignored_symbols/)), so I'm convinced that #4221 is safe; going ahead with this deletion for 5.6.2

PR #4228, opened 2017-05-08
Avoid concurrent installs warning for updates in Firefox

erikdesjardins:
Tested in browser: Chrome 60, Firefox 54


erikdesjardins:
Who knows what coveralls is doing ¯\\\_(ツ)_/¯

PR #4232, opened 2017-05-10
strawpoll.com mediahost

magicwizard8472:
* code needs checking. was just a copy of strawpoll.me's mediahost, but changed to fit this site.
* the embed's actual embed bit keeps changing its height depending on whether you're on the vote or results page, so that might need some adjusting
* wasn't sure what to call this, because a strawpoll mediahost already exists (strawpoll.me). I just went with "strawpoll.com"


erikdesjardins:
Thanks!

PR #4233, opened 2017-05-12
Settings gear icon click action change for consistency

corylulu:
Changed the Settings gear icon's click action to open the settings panel, not toggle the hover state of the menu. No other hover menu appears to have this behavior and I find it unintuitive and find myself accidentally closing the hover menu trying to click the cog.

Relevant issue: UX consistency 
Tested in browser: Chrome

### DIFF
--- a/lib/core/migrate/migrate.js
+++ b/lib/core/migrate/migrate.js
@@ -524,6 +524,13 @@ const migrations = [
 			await Storage.setMultiple(remappedEntries);
 			await Storage.delete('RESmodules.newCommentCount.counts');
 		},
+	}, {
+		versionNumber: '5.7.3-commandLine-menu-opening',
+		async go() {
+			await Migrators.moveOption('commandLine', 'launchFromMenuButton', 'RESMenu', 'gearIconClickAction', launchFromMenu =>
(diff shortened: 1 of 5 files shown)


Ajedi32:
This could potentially be a problem for mobile (e.g. Firefox) users, could it not? With this change, is it still possible for mobile users to open the hover menu without opening the settings panel?

andytuba:
People are actually using RES on mobile Firefox?

This has been discussed off and on. IMO if we're going to change this, it should be:

* option for clicking gear toggles menu (default), opens settings panel, or opens command line
* option for hovering gear opens menu (disabled by default)

corylulu:
@Ajedi32 Well if that's the case, there are plenty of other features that wouldn't work as well. Doubtful mobile users are using RES, though... Am I wrong in thinking that? 

@andytuba Okay, well I'll see if I can get some time on Monday to write in those options... but yeah, I'm not a big fan of the current system. I originally though to make it a click toggle menu, but it just didn't match the other similar elements... This change makes it identical to how the Mail button works.

corylulu:
@andytuba Updated to add those options. Lmk if I need to tweak anything. All worked from my tests.

corylulu:
Test failure due to behavior change of #RESConsoleContainer, but working as intended.

PR #4235, opened 2017-05-12
Fixed referenced object on reused hover element

corylulu:
**Relevant issue:** The hover menu for the subreddit manager gets reused after it's first creation, but doesn't update it's 'click' action references to the edit and delete button which tells it which subreddit to modify. So if you are trying to edit or delete subreddits, it will always populate the Edit window with the first subreddit that was hovered. 
**Tested in browser:** Chrome 59

### DIFF
--- a/lib/modules/subredditManager.js
+++ b/lib/modules/subredditManager.js
@@ -185,7 +185,7 @@ module.options = {
 	},
 };
 
-let shortCutsContainer, $subredditGroupDropdown, subredditGroupDropdownUL, $srList,
+let shortCutsContainer, $subredditGroupDropdown, subredditGroupDropdownUL, subredditGroupDropdownRefItem, $srList,
 	$editShortcutDialog, deleteButton, $sortMenu, gettingSubreddits,
 	subredditsLastViewed, sortShortcutsButton;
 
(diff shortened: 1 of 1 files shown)


corylulu:
@erikdesjardins Any way this can get pushed to the next update? It's a pretty simple fix that caused me quite a few problems. It's particularly problematic for anyone using the Multi-Reddits RES feature because those menu's immediately show, so it's very easy for the first one that gets hovered to be the incorrect one.

corylulu:
https://gfycat.com/VastUnluckyAnemone

Here's the bug in question. I have accidently deleted my multi's several times from this :( 
@andytuba Can you take a look at this?

lib/modules/subredditManager.js
```diff
@@ -438,6 +438,8 @@ function showSubredditGroupDropdown(obj) {
 }
 
 function _showSubredditGroupDropdown(obj, subreddits, suffix) {
+	// Updates source object referenced on the 'click' events
+	subredditGroupDropdownRefItem = obj;
```
erikdesjardins: I don't really like this, but the existing implementation already has a ton of global state so we can't really do better

erikdesjardins:
Sorry, I let this slip through the cracks

lib/modules/subredditManager.js
corylulu: I didn't like it either, but I didn't see an easy solution, so I went with essentially how everything else was working and just copied that.

corylulu: I think I had originally just added a variable to the `$subredditGroupDropdown` containing the object. Essentially: `$subredditGroupDropdown.selectedItem = obj;`, but it didn't match the existing workflow, so I went with that.

erikdesjardins: Yeah nothing on you, this whole module kinda needs to be refactored

PR #4237, opened 2017-05-13
keyboardNav move(Down|Up): Fix handling when nothing is selected / no target

larsjohnsen:
When nothing is currently selected, auto select the top-most visible thing. When something is selected, and there are no further entries, load the next page.


erikdesjardins:
(integration test failure is spurious)

Thanks!

PR #4240, opened 2017-05-16
NER: avoid clobbering non-NER URL hashes

erikdesjardins:
...like RES or Toolbox settings links.

Tested in browser: Chrome 60


creesch:
<3

PR #4256, opened 2017-05-21
keyboardNav: reduce coupling with showImages and selectedEntry

### LINKED ISSUE #4246: keyboardNav can't scroll to bottom with directional scroll style
- My browser: Chrome 58.0.3029.110 (64-bit) on Linux
- My RES version: 5.6.2

Problem: I use "directional" scrolling without "never ending reddit" or other features that auto load posts. Prior to [this change](https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/02155d5ccbc3b845053d41c3abf572e62157342c) I could move to the bottom of a link page by holding "j" until it stopped. Now, it's impossible to get to the bottom of the page without being very careful, since it automatically goes to the next page. I tried using the move bottom shortcut, but that seems to force a scroll style that puts the last link at the top of the window and leaves a huge blank area below. There is no longer an easy way to put the bottom link at the bottom of the window.

I would prefer that the moveDown action does not load the next page, since there is already a shortcut for that. Alternatively, the moveBottom action should follow the configured scroll style.

*Are there any related posts in https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues or /r/RESissues?* no

larsjohnsen:
Moves `move` functions to selectedEntry
Moves mediaBrowse completely to showImages (as both wheelBrowse and keyboardNav invokes it)

Relevant issue: Closes #4246
Tested in browser:  Chrome 59

### DIFF
--- a/lib/core/migrate/migrate.js
+++ b/lib/core/migrate/migrate.js
@@ -511,6 +511,11 @@ const migrations = [
 		async go() {
 			await Storage.delete('RESmodules.userTagger.tags');
 		},
+	}, {
+		versionNumber: '5.7.0-keyboardNav-showImages-decoupling',
+		async go() {
+			await Migrators.moveOption('keyboardNav', 'mediaBrowseMode', 'showImages', 'mediaBrowse');
+		},
 	},
 
 ];  // ^^ Add new migrations ^^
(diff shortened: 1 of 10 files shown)


lib/core/migrate/migrate.js
```diff
@@ -513,6 +513,12 @@ const migrations = [
 		},
 	},
 
+	{ // FIXME Place correctly in migrator list
```
erikdesjardins: I'm not sure what you mean here; this PR looks good so it can be probably be merged for the first 5.7.0 beta, so the migrator can probably go right here

PR #4257, opened 2017-05-22
Make author and options manifest keys as consistent as possible across browsers

erikdesjardins:


BenMcGarry:
options_page on edge supports open in new tab as standard, ill reach out to get info on options_ui.

Issue #2278, opened 2015-06-12
Show settings console in separate tab


andytuba:
Some subreddit styles make the settings console very hard to read. If the settings console were built ~~inside an iframe~~ in its own tab, then it would be immune from vanilla Reddit or subreddit styles.

---

Also send a notification to all Reddit tabs that "settings have changed, reload?" and/or option to autoreload

andytuba:
Justification: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/2431

PR #4259, opened 2017-05-22
Prepare to move options to standalone page

erikdesjardins:
Relevant issue: #2278
Tested in browser: -


erikdesjardins:
Implemented in #4859

PR #4264, opened 2017-05-30
Subreddit manager features

corylulu:
So this does three things; two feature additions, of which I've been doing for a while with a hack and a stylesheet tweak, and one bug fix. 

1. Option to highlight a multi on the subreddit manager bar if at any of it's nested subreddits. So if I have a `Gaming` multi the target of `overwatch+leagueoflegends+dota2+heroesofthestorm`, if I'm on any of those subreddits, `Gaming` will get highlighted as if it were the current sub. For people that utilize the multi's mainly for the dropdown, this should make it more intuitive for them if they want to enable it. Also makes the next thing more useful. Example: https://gfycat.com/HatefulFalseFlicker

2. Allows for an alternative use of the dropdown to act as a normal, single subreddit with "related subs" in the dropdown. In use, it looks like this: 
 
    **Subreddit:** `leagueoflegends?+leagueoflegends/new+leagueofmeta+loleventvods`
    **Display Name:** `leagueoflegends ▼`
    **Results:** https://gfycat.com/MenacingWastefulCorydorascatfish 
 
The inclusion of the `?` on the first subreddit is all you need. This is how I made it work before because the URL would ignore everything after that point. Shouldn't negatively affect anything else, can be used along side normal multi links, and is pretty simplistic in how to make use of it.

3. Also fixes a bug where reddit redirects url multi's to the ~~alphabetized~~ *randomized* version of the same link. So `/r/c+b+a` ~~would~~ *might* navigate to `/r/a+b+c` *or `/r/b+a+c`*, which is fine, but if it's on your subreddit bar, it wouldn't show as highlighted.

### DIFF
--- a/lib/modules/subredditManager.js
+++ b/lib/modules/subredditManager.js
@@ -551,7 +551,7 @@ function editSubredditShortcut(ele, event) {
 					<input type="text" name="subreddit" value="${subreddit}" id="shortcut-subreddit" class="${unsortable ? 'unsortable' : ''}"><!-- no whitespace
 					--><button type="submit" id="sortButton" title="Sort subreddits">A-Z</button>
 
-					<div class="RESDescription">Put a + between subreddits to make a drop-down menu.</div>
+					<div class="RESDescription">Put a + between subreddits to make a drop-down menu.<br/>Put ?+ to make subreddits after it only show in dropdown.</div>
 				</div>
 			</div>
 			<div class="RESFormItem">
--- a/lib/modules/subredditManager.js
+++ b/lib/modules/subredditManager.js
@@ -345,8 +345,8 @@ function redrawShortcuts() {
 
 		if (shortcut.subreddit.includes('+')) {
 			// Reddit seems to redirect to randomly reorder multi links, so need to sort and compare
-			let currentSub = (currentSubreddit() || '').toLowerCase().split('+').sort().join('+');
-			let sortedSubs = shortcut.subreddit.toLowerCase().replace(/\?\+/g, '+').split('+').sort();
+			const currentSub = (currentSubreddit() || '').toLowerCase().split('+').sort().join('+');
+			const sortedSubs = shortcut.subreddit.toLowerCase().replace(/\?\+/g, '+').split('+').sort();
 			if (module.options.showMultiAsCurrentIfAtChild.value) {
 				for (const sub of sortedSubs) {
 					if (isCurrentSubreddit(sub)) {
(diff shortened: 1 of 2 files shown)


corylulu:
@erikdesjardins If there is anything else you'd want me to modify to merge these changes, lmk. This allows for a lot of clutter reduction on the subreddit bar and I use it everyday. If I get some time, maybe I can rework the subreddit manager and clean some stuff up and add some polish and usability.

lib/modules/subredditManager.js
```diff
@@ -337,11 +343,26 @@ function redrawShortcuts() {
 		thisShortCut.setAttribute('data-subreddit', shortcut.subreddit);
 		thisShortCut.classList.add('subbarlink');
 
-		if (isCurrentSubreddit(shortcut.subreddit)) {
+		if(shortcut.subreddit.includes('+')) {
+			// Reddit seems to redirect to randomly reorder multi links, so need to sort and compare
+			var currentSub = (currentSubreddit() || '').toLowerCase().split('+').sort().join('+');
+			var sortedSubs = shortcut.subreddit.toLowerCase().replace(/\?\+/g, '+').split('+').sort();
+			if (module.options.showMultiAsCurrentIfAtChild.value) {
+				for (var i in sortedSubs) {
```
erikdesjardins: `for (const sub of sortedSubs) {`

erikdesjardins:
Also if you're gonna clean up subreddit manager, it'd be great if it changed the subreddits while keeping reddit's css classes and (roughly) DOM structure, instead of replacing the entire bar with `RESShortcutsViewport` and `srLeftContainer`. (should let us get rid of a bunch of CSS and might work a bit better with subreddit styles)

corylulu:
@erikdesjardins One issue is it won't be transferable to the new profiles that way (see https://www.reddit.com/user/corylulu). For whatever reason, reddit felt the need to totally change the DOM structure throughout. 

So maybe it's not such a bad idea to keep completely redone elements independent of Reddit's resources considering it's all likely to change in the near future and would probably require two separate implementations to make it work. 

I can write it to use the old DOM structure, but I think it would still be preferred to rebuild to entire element and use an independent stylesheet (perhaps loaded on demand) so it will still work on the new layout pages. 

I won't have time this week to work on it, but I'll try if I have the time in the next couple weeks. If any of you guys have any insight on any of the new layouts, lmk as well...

erikdesjardins:
Hmmm, I would still like to do that eventually, but you're right it's probably not worth doing until the current pages are completely replaced.

PR #4265, opened 2017-05-30
Update streamable pathname regex

magicwizard8472:
To support links like these:

* https://streamable.com/s/b416n
* https://streamable.com/s/b416n/zxmwtq


erikdesjardins:
👍

PR #4266, opened 2017-05-31
Backup to Google Drive

larsjohnsen:
Interfaces with Google Drive to store on the user's Drive:
image

- [ ] ~~Establish a Google Drive app with the necessary redirection URLs~~ #4277
- [ ] Evaluate safety aspects, policy notice (afaics the only way for anyone to view/modify the data is via this script run on the user's behalf)


lib/modules/backupAndRestore.js
```diff
@@ -44,38 +96,71 @@ async function backup() {
...
+	});
 }
 
-function restoreFromFile(link) {
-	const file = link.files[0];
+const provider = module.options.cloudProvider.value;
 
-	const reader = new FileReader();
+const setupCloud = _.once(() => {
+	const iframe = downcast(string.html`<iframe hidden sandbox="allow-same-origin allow-scripts allow-popups allow-same-origin allow-forms" src="https://larsjohnsen.github.io/${provider}.html"></iframe>`, HTMLIFrameElement);
```
creesch: This probably needs changing

larsjohnsen: Yes — must be adopted

corylulu:
Any reason this couldn't be made to be automatic on a periodic basis? Would be great to have automatic backups as well as automatic syncing across computers.

erikdesjardins:
I would also like automatic syncing, but I think the initial release should be as simple as possible.

Also, if/when it's automatic, it will probably just be naive, with a corresponding warning about lost data if you use two machines at once. If RES Pro was any indication, proper reconciliation would be nightmarish.

corylulu:
Oh, I kinda just figured there was a reason native syncing couldn't be implemented since I remember it being a problem in the past. Isn't the native sync fairly limited in terms of storage?

But i don't think there would be that many issues if the frequently changing data is kept separate and things like settings options and subreddit manager stuff is kept isolated. Would also be good to keep auto-backups around if possible, even if though Google Drive, so in the case of a bad sync, it could easily be reverted.

erikdesjardins:
Ah, I said na**i**ve, not na**t**ive, sorry 😓 , i.e. no individual key syncing. You're right, native `chrome.storage.sync` is far too limited.

Separating it is exactly the problem I'm talking about, cause we'd have to track writes on every key (or some subset of keys), which is (probably) not gonna happen. We can get >90% of the functionality with <10% of the complexity by just polling for the modified date every 5 mins and replacing everything.

There'll probably be an option (enabled by default), though, to popup an alert to the effect of "new backup at \<date\>, restore?"

erikdesjardins:
I'll switch the application to a proper one in #4277

PR #4269, opened 2017-06-01
Use launchWebAuthFlow to get OAuth tokens

erikdesjardins:
Relevant issue: for #4266

### DIFF
--- a/chrome/manifest.json
+++ b/chrome/manifest.json
@@ -40,6 +40,7 @@
 	}],
 	"permissions": [
 		"cookies",
+		"identity",
 		"tabs",
 		"history",
 		"storage",
--- a/browser/background.js
+++ b/browser/background.js
@@ -45,39 +45,6 @@ addListener('ajax', async ({ method, url, headers, data }) => {
 	};
 });
 
-const waiting = new Map();
-addListener('authFlow', ({ operation, id, token }) => {
-	switch (operation) {
-		case 'start':
-			if (waiting.has(id)) {
-				throw new Error(`Auth handler for id: ${id} already exists.`);
-			}
-			return new Promise((resolve, reject) => waiting.set(id, { resolve, reject }));
-		case 'complete': {
-			const handler = waiting.get(id);
-			if (!handler) {
-				console.error(`No auth handler for id: ${id} (sent token: ${token}).`);
-				return false;
-			}
-			waiting.delete(id);
-			handler.resolve(token);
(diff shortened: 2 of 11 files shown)


chrome/manifest.json
```diff
@@ -40,6 +40,7 @@
 	}],
 	"permissions": [
 		"cookies",
+		"identity",
```
erikdesjardins: CHECK IF THIS CAUSES NEW PERMISSIONS WARNINGS

erikdesjardins:
```js
testAuthFlow: {
	title: 'test auth flow',
	type: 'button',
	text: 'test',
	callback() {
		launchAuthFlow({
			domain: 'https://accounts.google.com/o/oauth2/v2/auth',
			clientId: '237880911304-a4040773adrek8uq6dpbg2p7308kf03u.apps.googleusercontent.com',
			scope: 'https://www.googleapis.com/auth/drive.appdata',
		}).then(alert, e => alert(e.message));
	},
	description: '',
},
```

erikdesjardins:
I can confirm that the identity permission does not cause a permissions warning. (if I made a mistake, the beta release will be a final test)

Notes:
- it's not mentioned (positively or negatively) on the [permission warnings](https://developer.chrome.com/extensions/permission_warnings) page
- when installing from scratch, it's not listed in the warning popup
- running your own autoupdate server ([as suggested](https://developer.chrome.com/extensions/permission_warnings#test)) _does not work_, since Chrome doesn't let you enable packaged extensions that weren't signed by the Chrome Web Store (ugh)--you can even add wide-reaching permissions like `<all_urls>` without a warning
- ...so instead I used a [real extension](https://github.com/erikdesjardins/global-mediakeys/pull/51)

larsjohnsen:
I'm not sure about how big an issue this is, but by using `identity` the native oauth flow can be a little mysterious:
image instead of, as with Edge: 
image. Using this also makes development somewhat harder, since `redirect_url` must be specified to match the build's generated id.

BenMcGarry:
@larsjohnsen is there no way to change it from `chromiumapp.org` on `identity`? I reckon we would see a lot of people asking what it is.

erikdesjardins:
Wait a second...Firefox `chrome.identity.launchWebAuthFlow` works even when redirecting to res.com instead of the special `"mozilla".reverse()` URL, and Chrome also appears to (but needs more testing)...wat

PR #4278, opened 2017-06-04
Automatic backups

erikdesjardins:

re #4266

### DIFF
--- a/locales/locales/en.json
+++ b/locales/locales/en.json
@@ -1632,6 +1632,21 @@
 	"accountSwitcherDropDownStyleDesc": {
 		"message": "Use the \"snoo\" icon, or older style dropdown?"
 	},
+	"backupAndRestoreReloadWarningTitle": {
+		"message": "Reload Warning"
+	},
+	"backupAndRestoreReloadWarningDesc": {
+		"message": "When restoring a backup, either manually or automatically, how should other tabs be reloaded? Tabs must be reloaded for restored settings to take effect."
+	},
+	"backupAndRestoreReloadWarningNone": {
+		"message": "Do nothing."
+	},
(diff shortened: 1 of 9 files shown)


locales/locales/en.json
```diff
@@ -1632,6 +1632,33 @@
...
+		"message": "Open a prompt to reload other tabs."
+	},
+	"backupAndRestoreReloadWarningAuto": {
+		"message": "Automatically reload other tabs."
+	},
+	"backupAndRestoreAutomaticBackupsTitle": {
+		"message": "Automatic Backups"
+	},
+	"backupAndRestoreAutomaticBackupsDesc": {
+		"message": "WARNING: THIS OPTION IS EXPERIMENTAL AND MAY DELETE ALL OF YOUR SETTINGS, TAGS, ETC. AT ANY TIME.\n\nSync your backups to the specified provider.\n\nOn pageload, every five minutes, if a new backup is detected, it will be imported, replacing all settings. Otherwise, a new backup will be created. Changes aren't (and will not be) reconciled, so if you quickly switch between two computers you may lose some changes."
```
erikdesjardins: This ought to do it. Yes, definitely. 😉

erikdesjardins:
@corylulu so I guess we can just mark it experimental, why not. ¯\\\_(ツ)_/¯

corylulu:
@erikdesjardins Nice, might be a good idea to store an md5 of the data to help safe guard against a bad backup/restore if it's not already there.

corylulu:
@erikdesjardins What do I need to do to make this work if I'm building it myself? Should I need a specific chrome.pem or anything while packing the extension or will I need to supply my own client ID? 

I can't get it to not show `Error in target's "authFlow" handler: Authorization page could not be loaded.`

PR #4279, opened 2017-06-04
Fix bigEditor not submitting when opened by keyboard shortcut

larsjohnsen:
Fixes https://www.reddit.com/r/RESissues/comments/6f6q42/res_big_editor_shortcut_is_broken_deletes_posts/
https://www.reddit.com/r/Enhancement/comments/5nssy5/not_sure_if_bug_or_not_i_just_typed_a_nice_long/
Tested in browser: Chrome 59

### DIFF
--- a/lib/modules/commentPreview.js
+++ b/lib/modules/commentPreview.js
@@ -119,7 +119,7 @@ module.beforeLoad = () => {
 module.go = () => {
 	if (module.options.enableBigEditor.value) {
 		// Install the 2 column editor
-		addBigEditorListener();
+		$('body').on('click', '.RESBigEditorPop', showBigEditor);
 
 		if (Modules.isRunning(KeyboardNav)) {
 			$('body').on('keydown', '.usertext-edit textarea, #wiki_page_content', (e: KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


lib/modules/commentPreview.js
```diff
@@ -119,13 +119,11 @@ module.beforeLoad = () => {
...
-		addBigEditorListener();
+		$('body').on('click', '.RESBigEditorPop', showBigEditor);
 
 		if (Modules.isRunning(KeyboardNav)) {
-			$('body').on('keydown', '.usertext-edit textarea, #wiki_page_content', (e: KeyboardEvent) => {
-				if (checkKeysForEvent(e, module.options.openBigEditor.value)) {
-					showBigEditor(e);
-				}
+			$('body').on('keydown', '.usertext-edit textarea, #wiki_page_content', function(e: KeyboardEvent) {
+				if (checkKeysForEvent(e, module.options.openBigEditor.value)) showBigEditor.bind(this)(e);
```
erikdesjardins: Can we change it to use `e.target`? Looks like it should be the same as `this` in both listeners (if it isn't, we find the textarea with `.parents(...).find()` so it should still work).

erikdesjardins:
Thanks!

PR #4289, opened 2017-06-07
Added Adult Swim as a Host to enable embedding video player

jhumbug:


lib/modules/hosts/adultswim.js
```diff
@@ -0,0 +1,22 @@
+/* @flow */
+
+import { Host } from '../../core/host';
+
+export default new Host('adultswim', {
+	name: 'Adult Swim',
+	domains: ['adultswim.com'],
+	logo: 'http://www.adultswim.com/favicon.ico',
+	detect: ({ href }) => (/^(?:https?)?:\/\/(?:(?:www)\.)?adultswim.com[\w\-\/:#]+videos\/([\S]+)/i).exec(href),
```
magicwizard8472: Hm, might be better to use the following here:

`detect: ({ pathname }) => (/^\/videos\/([a-z0-9-#\/]+)$/i).exec(pathname)`

erikdesjardins:
Thanks!

PR #4295, opened 2017-06-10
Slight change to help speedup user tag loading

corylulu:
The async operation is long and yields empty results far more often than not, so it makes more sense to load the user tag icon by default since adding them in often shifts around the whole page layout, then swapping it out for a tag if it exists.

The user tag icon appears anywhere between 500ms - 1500ms faster doing it this way from my tests. Makes the page feel fully loaded much more quickly.

Tested in Chrome 60

### DIFF
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -819,7 +819,7 @@ export async function applyTagToAuthor(thisAuthorObj: HTMLAnchorElement, noEdit:
 	const tagObject = await getTagBatched(thisAuthor);
 
 	if (!_.isEmpty(tagObject)) {
-		if (tagElement && tagElement.parentElement) tagElement.parentElement.removeChild(tagElement);
+		if (tagElement) tagElement.remove();
 		addAuthorTag(thisAuthorObj, noEdit, thisAuthor, tagObject);
 	}
 
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -883,7 +883,7 @@ function addAuthorTag(thisAuthorObj: HTMLAnchorElement, noEdit: boolean, thisAut
 	}
 
 	if (noEdit) {
-		if (tag) addUserTag();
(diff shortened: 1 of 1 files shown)


lib/modules/userTagger.js
```diff
@@ -811,9 +811,15 @@ export async function applyTagToAuthor(thisAuthorObj: HTMLAnchorElement, noEdit:
...
+	let tagElement;
+	if (module.options.showTaggingIcon.value) {
+		tagElement = addAuthorTag(thisAuthorObj, noEdit, thisAuthor);
+	}
+
 	const tagObject = await getTagBatched(thisAuthor);
 
-	if (module.options.showTaggingIcon.value || !_.isEmpty(tagObject)) {
+	if (!_.isEmpty(tagObject)) {
+		if (tagElement && tagElement.parentElement) tagElement.parentElement.removeChild(tagElement);
```
erikdesjardins: `if (tagElement) tagElement.remove();`

corylulu: Is that using the jQuery definition or predefined? I try to avoid jQuery if I can since it's not included in the new profile build, but if you'd prefer jQuery extensions, I'll just use those.

erikdesjardins: It's [part of the DOM 4 spec](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove), which also has a few other jQuery-like things (e.g. [closest()](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)).

We ship a separate version of jQuery because we can't use reddit's (due to sandboxing), so technically it would still work on profile pages. I plan to get rid of it eventually, though, so avoiding it is a good idea. (the main issue is we use a few [jQuery plugins](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/vendor/jqueryPlugins.js) which would have to be replaced)

PR #4298, opened 2017-06-10
newCommentCount: Clean-up and concurrency fix

larsjohnsen:
This should prevent that one instance overwrites another, as sometimes can be seen by `updateCommentCountFromMyComment` sometimes being ineffective.

### DIFF
--- a/lib/core/migrate/migrate.js
+++ b/lib/core/migrate/migrate.js
@@ -516,6 +516,14 @@ const migrations = [
 		async go() {
 			await Migrators.moveOption('keyboardNav', 'mediaBrowseMode', 'showImages', 'mediaBrowse');
 		},
+	}, {
+		versionNumber: '5.7.1-shard-newCommentCount',
+		async go() {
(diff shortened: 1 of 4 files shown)


lib/core/migrate/migrate.js
```diff
@@ -516,6 +516,13 @@ const migrations = [
 		async go() {
 			await Migrators.moveOption('keyboardNav', 'mediaBrowseMode', 'showImages', 'mediaBrowse');
 		},
+	}, {
+		versionNumber: '5.7.1-shard-newCommentCount',
+		async go() {
+			const tags = await Storage.get('RESmodules.newCommentCount.counts') || {};
+			const remappedTags = _.mapKeys(tags, (v, k) => `newCommentCount.${k.toLowerCase()}`);
+			await Storage.setMultiple(remappedTags);
```
erikdesjardins: You can also `await Storage.delete('RESmodules.newCommentCount.counts');` here, since they're temporary anyways, doesn't really matter if the migration goes wrong

PR #4307, opened 2017-06-14
commentHidePersistor: Simplify logic, convert to watchForThing

larsjohnsen:
Like in #4298, fixes multiple instances overwriting each other.

### DIFF
--- a/lib/modules/commentHidePersistor.js
+++ b/lib/modules/commentHidePersistor.js
@@ -3,7 +3,7 @@
 import { $ } from '../vendor';
 import { Module } from '../core/module';
 import { Storage } from '../environment';
-import { click } from '../utils';
+import { DAY, click, execRegexes, watchForThings, Thing } from '../utils';
 
 export const module: Module<*> = new Module('commentHidePersistor');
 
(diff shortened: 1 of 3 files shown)


lib/modules/commentHidePersistor.js
```diff
@@ -16,115 +16,66 @@ module.include = [
 	'inbox',
 ];
 
-const hidePersistorStorage = Storage.wrap('RESmodules.commentHidePersistor.hidePersistor', {
-	hiddenThings: ({}: { [key: string]: string[] }),
-	hiddenKeys: ([]: string[]),
-});
+const expireAfterDays = 7;
```
larsjohnsen: Changed from number of entries (max 100 things / 50 comment pages)

lib/modules/commentHidePersistor.js
```diff
@@ -16,115 +16,66 @@ module.include = [
...
 
-function removeHiddenThing(thingId) {
-	const i = hiddenThings.indexOf(thingId);
-	if (i !== -1) {
-		hiddenThings.splice(i, 1);
+async function restoreCommentCollapse() {
+	const { collapsedThings } = await entryStorage.get(currentId);
+	if (collapsedThings) {
+		watchForThings(['comment'], thing => {
+			if (Reflect.has(collapsedThings, thing.getFullname())) {
```
erikdesjardins: FWIW you can use `in` here if you want, I don't really care about enforcing the use of `Reflect` (mainly enforce `Reflect.apply` just so it stands out more than `fn.call`). If you prefer it this way though then feel free to keep it.

PR #4315, opened 2017-06-17
Allow markdown for module descriptions

larsjohnsen:
From #4191

Tested in browser: Chrome 60

### DIFF
--- a/lib/modules/settingsConsole.js
+++ b/lib/modules/settingsConsole.js
@@ -189,7 +189,8 @@ function renderModulesSelector() {
 			modules: modules
 				.sort(compareModules)
 				.map(mod => {
-					const description = $(`<p>${i18n(mod.description)}</p>`).text().replace(/\s+/g, ' ');
+					const description = $(`<p>${mod.descriptionRaw ? mod.description : markdown(i18n(mod.description))}</p>`).text().replace(/\s+/g, ' ');
+
 					return {
 						moduleID: mod.moduleID,
 						translatedName: i18n(mod.moduleName),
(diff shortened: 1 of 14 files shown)


lib/environment/i18n.js
```diff
@@ -56,3 +57,13 @@ export function i18n(messageName: string, ...substitutions: Array<string | numbe
 	// but Transifex does not create the `placeholders` field in exported JSON
 	// https://developer.chrome.com/extensions/i18n#examples-getMessage
 }
+
+export function i18nMarkdown(messageName: string, ...substitutions: Array<string | number>): string {
+	const message = i18n(messageName, ...substitutions);
+	// Some messages are not i18n-ized and may still be raw html
+	const isHTML = (/<[a-z][\s\S]*>/i).test(message);
+	if (process.env.NODE_ENV === 'development' && isHTML && messages.hasOwnProperty(messageName)) {
```
larsjohnsen: Since this isn't very robust, this should probably be changed to always run through the markdown function when `message` is found in `messages`.

lib/modules/settingsConsole.js
```diff
@@ -189,7 +188,7 @@ function renderModulesSelector() {
 			modules: modules
 				.sort(compareModules)
 				.map(mod => {
-					const description = $(`<p>${i18n(mod.description)}</p>`).text().replace(/\s+/g, ' ');
+					const description = $(`<p>${i18nMarkdown(mod.description)}</p>`).text().replace(/\s+/g, ' ');
```
larsjohnsen: TODO No need to wrap i18Markdown

larsjohnsen:
Changed this to simply flag descriptions that should be converted.

Issue #4319, opened 2017-06-20
Wikipedia expando shows LaTeX code for math formulas


jkayani:
- My browser (copy from https://www.whatismybrowser.com): Chrome 59 on Linux
- My RES version: 5.6.4
- Am I in reddit beta (yes/no): no

*What's the problem?*
Whenever a Wikipedia article with math formulas is rendered in the Wikipedia expando, the math formulas are accompanied by the LaTeX code used to render them on Wikipedia, and what appears to be a Unicode rendering of some of the symbols. Since the rendering step doesn't happen in the expandos, the LaTeX gets in the way and makes things difficult to read. Also, there are "fallback images" that come with the article text, and therefore neither the Unicode renderings or LaTeX code are necessary in the expando. 

This is what the current Wikipedia expando looks like.
![screenshot_20170615_211849](https://user-images.githubusercontent.com/7196382/27311957-865b7006-5532-11e7-801c-1dd7b5bac7a0.png)

*Are there any related posts in https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues or /r/RESissues?*
No.

*What other browser extensions are installed?*
UBlock Origin, LastPass

PR #4320, opened 2017-06-20
Fixes Wikipedia expando issue with LaTeX for math formulas 

jkayani:
The Unicode renderings and LaTeX code for math symbols in Wikipedia articles are removed from the HTML that is rendered by the Wikipedia expando, leaving just the fall-back images. This makes it easier to read and is less cluttered.

This is what it looked like before:
screenshot_20170615_211849

This is what it looks like with the patch applied:
screenshot_20170619_211220

Relevant issue: #4319 
Tested in browser: Chrome


erikdesjardins:
Thanks!

PR #4321, opened 2017-06-20
Check for visited class in visited filter (#3571)

### LINKED ISSUE #3571: Is url visited should check for reddit sync visited
Ie scan the current page for a post linking to that same URL where its thing has reddit's .visited class

sargon2:
Relevant issue: fixes #3571
Tested in browser: Chrome 59.0.3071.104 (Official Build) (64-bit)


erikdesjardins:
Thanks!

PR #4326, opened 2017-06-22
Expandos are no longer direct children of .entry

erikdesjardins:
Relevant issue: https://www.reddit.com/r/RESissues/comments/6iw03n/two_media_expandos_per_post/
Tested in browser: Chrome 60

Only the two selector changes will be backported for the hotfix.

### DIFF
--- a/lib/modules/showImages/expando.js
+++ b/lib/modules/showImages/expando.js
@@ -10,12 +10,10 @@ import {
 } from '../../utils';
 import type { Thing } from '../../utils';
 
-// @type {WeakMap.<expandoButton, Expando>}
 export const expandos: WeakMap<Element, Expando> = new WeakMap();
 
 // Used for specifying primary expando for any given link
-// @type {Map.<string, Expando>} - key: href
-export const primaryExpandos: Map<string, Expando> = new Map();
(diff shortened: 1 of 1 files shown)


lib/modules/showImages/expando.js
```diff
@@ -54,13 +52,13 @@ export class Expando {
 	static getEntryExpandoFrom(thing: ?Thing): ?Expando {
 		if (!thing) return null;
 
-		const button = $(thing.entry).find('> .expando-button')[0];
+		const button = thing.entry.querySelector('.expando-button');
```
erikdesjardins: This is safe because if there are multiple expando-buttons, the first one will be the parent (self post)

lib/modules/showImages/expando.js
```diff
@@ -54,13 +52,13 @@ export class Expando {
...
 		if (!button) return null;
 
 		let expando = expandos.get(button);
 
 		if (!expando) {
-			const box = $(thing.entry).find('> .expando')[0];
+			const box = thing.entry.querySelector('.expando');
```
erikdesjardins: This is safe because the entry doesn't include child comments, etc., so there can only be one `.expando`

erikdesjardins:
Integration test failure appears unrelated

PR #4327, opened 2017-06-23
settingsConsole builder: Style cleanup

larsjohnsen:
Changes made in order to allow easier insertion of its components to other contexts, such as infocards.

From #4191 

Tested in browser: Chrome 61, Firefox 55

### DIFF
--- a/lib/css/res.scss
+++ b/lib/css/res.scss
@@ -413,17 +413,9 @@ div.usertext-edit {
 	}
 
 	ul {
-		margin-bottom: 10px;
-		margin-top: 10px;
-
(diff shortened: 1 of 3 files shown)


lib/css/res.scss
```diff
@@ -413,18 +413,10 @@ div.usertext-edit {
...
 	ul {
-		margin-bottom: 10px;
-		margin-top: 10px;
-
 		&.token-input-list-facebook {
 			margin: 0;
 		}
 	}
-
-	li {
```
larsjohnsen: Very generic rule — did not see any place where it applied.

PR #4329, opened 2017-06-25
Add Search Keybinding & RES-Console Commands

### LINKED ISSUE #2211: Add key bindings for searching contents
There should be a shortcut for quickly add focus to the search bar, so that it's possible to search content using keyboard.

Also, the RES Command console should include a command for searching in the whole reddit (or subreddit), and not only to search in RES settings.

Propheis:
Closes #2211 
Tested in browser: Firefox 54.0, Chrome 59.0.3071.109

This PR adds:

1. A "/" keybinding to focus the cursor on the search box anywhere on Reddit.
2. Two new commands on the RES Command Console

The two new commands are

1. `s[earch]: <query>` - Search current subreddit (if any) or all of Reddit for "query"
2. `sr: <query>` - Search all of Reddit for "query"

Any thoughts?

### DIFF
--- a/lib/modules/commandLine.js
+++ b/lib/modules/commandLine.js
@@ -572,7 +572,7 @@ registerCommand(/^\/([nthrcgp])?/, '/n, /t, /h, /r, /c, /g, or /p - goes to new,
 	}
 );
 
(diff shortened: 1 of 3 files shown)


erikdesjardins:
We can't use `/` by default because it breaks native functionality in Firefox; changed it to `alt-/`
(you can of course change it to `/` if you want to)

erikdesjardins:
Thanks!

PR #4330, opened 2017-06-29
Added a shortcut to show/hide child comment trees on the comment page.

### LINKED ISSUE #1274: keyboard shortcut for show / hide child comments
http://www.reddit.com/r/RESissues/comments/28drjj/using_keyboard_shortcut_to_show_child_comments_of/

maybe shift-enter or alt-enter?

decadent:
"C" by default.

Fixes #1274.

Tested in Chrome and Firefox.

I have the feeling that one shortcut should serve to expand comment trees, collapse them and load more comments—instead of two shortcuts (C and Enter) plus navigation to "load more comments."

With "show/hide child comments" for nested comments enabled, the present change allows to use one shortcut to expand top-level comments (if collapsed by default) and collapse trees of child comments.

However, I also think that child comment trees that were deliberately collapsed should be highlighted visually, like it happens now with Enter—because I use this regularly to find comments that I just read but didn't think to save or leave a reply.  Not sure how this is best done, yet.

### DIFF
--- a/lib/modules/hideChildComments.js
+++ b/lib/modules/hideChildComments.js
@@ -118,3 +118,9 @@ function addToggleChildrenButton(comment) {
 	li.appendChild(a);
 	menu.appendChild(li);
 }
+
+export function toggle(thing: *) {
+	const button = thing.entry.querySelector('a.toggleChildren');
+	if (!button) throw new Error('Toggle button not found');
+	button.click();
+}
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -584,6 +584,14 @@ module.options = {
 			click(getFirstElementInThingByQuery('.entry .edit-usertext', ASSERT, getSelected()));
 		},
 	},
+	showChildComments: {
+		type: 'keycode',
+		include: ['comments', 'commentsLinklist'],
+		value: [67, false, false, false, false], // c
+		description: 'keyboardNavShowChildCommentsDesc',
+		title: 'keyboardNavShowChildCommentsTitle',
+		callback: showChildComments,
+	},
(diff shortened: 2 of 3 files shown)


lib/modules/keyboardNav.js
```diff
@@ -1207,6 +1215,16 @@ function reply() {
 	}
 }
 
+function showChildComments() {
+	const selected = SelectedEntry.selectedThing;
+	if (!selected) return;
+
+	const button = selected.entry.querySelector('a.toggleChildren');
```
andytuba: Instead of clicking the button, could you call the "toggle children on this entry" method directly?

decadent:
Huh, this seems to move faster than I do =)

Regarding andytuba's request to avoid faking clicks on the links and use RES' functions instead: afaik RES only adds 'hide child comments' links on nested comments, whereas links on the root nodes are Reddit's―which means the root links potentially can have new event handlers on them in the future, such as statistics. So personally I wouldn't override 'native' clicks with custom functions, but I don't know if there are other considerations at play.

andytuba:
"hide child comments" is only added by RES.

lib/modules/hideChildComments.js
```diff
@@ -118,3 +118,9 @@ function addToggleChildrenButton(comment) {
 	li.appendChild(a);
 	menu.appendChild(li);
 }
+
+export function toggle(thing: *) {
+	const button = thing.entry.querySelector('a.toggleChildren');
+	if (!button) throw new Error('Toggle button not found');
+	button.click();
```
andytuba: It would be nice to see this module refactored to avoid the click, but this is a reasonable compromise.

PR #4335, opened 2017-07-07
Add host file for supgif.com

jrogozen:
Hey there,

Added host file supgif.com domain which plays an iframe.

Is there a specific protocol for adding hosts?

Thanks!


andytuba:
Why did you name the file `supgifjs`?

jrogozen:
whoops, wrong file. sorry

PR #4342, opened 2017-07-16
Remove loadDynamicOptions

### LINKED ISSUE #4262: Can't hide subreddit style checkbox or toolbar button
- My browser: Firefox 53 on Windows 7
- My RES version: 5.6.4

*What's the problem?*

Neither the subreddit style toolbar button, nor the Use "subreddit style" checkbox respond to settings. "subredditStyleBrowserToolbarButton" is off, and "styleTweaks" is off, but both still appear. The checkbox is on all pages (duplicating Reddit's own "Show this subreddit's theme"), and the toolbar button is on pages where I've previously toggled the style off.

A similar bug specific to the toolbar button was reported for 5.6.0, and a mod responded that it was fixed, but I still experience the bug. https://www.reddit.com/r/RESissues/comments/69n6ub/subreddit_css_toolbar_button_cant_be_properly/

Stylish (reddit smaller (normal) font size fix), Click to Play per-element, YesScript

erikdesjardins:
Relevant issue: fixes #4262; #4422 may also be fixed
Tested in browser: Chrome 61, Firefox 55, ~~Edge 15~~ (Edge fails)

On my machine, these special-cases are no longer necessary to avoid FOUC (yay, Firefox is fast now).

But this is pretty risky, so it needs a beta cycle.

The main benefit of this, other than fixing the above issue, is that it becomes easier to reason about module state, because it becomes harder to have code run before options are loaded. (Except for a function getting called immediately at the top level, which should be more obviously wrong.)

### DIFF
--- a/lib/core/init.js
+++ b/lib/core/init.js
@@ -9,82 +9,92 @@ import {
 	waitForChild,
 	waitForEvent,
 	newSitetable,
+	markStart,
+	markEnd,
+	markCheckpoint,
 } from '../utils';
+import { _logPerfSummary } from '../utils/profiling';
 import { _loadModuleOptions } from './options/options';
 import { _loadModulePrefs, _runModuleStage } from './modules/modules';
 import { _addModuleBodyClasses } from './modules/bodyClasses';
 import { migrate } from './migrate';
 
-let start;
-
-export function init() {
+export async function init() {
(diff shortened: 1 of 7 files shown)


erikdesjardins:
No longer needed, I don't think, see linked issue resolutions

PR #4347, opened 2017-07-16
Open module settings by cli `mod xyz`

larsjohnsen:
Tested in browser: Chrome 60


andytuba:
Neato! Looks like it even supports `xyz` for modules named `fooxyzbar`?

larsjohnsen:
Yep—I see no reason not to.

PR #4348, opened 2017-07-16
Reduce newCommentCount slowdown

larsjohnsen:
Attempts to fix https://www.reddit.com/r/RESissues/comments/6nly3t/constant_color_wheel_after_580_update/ by running `checkSubscriptions` at most once every 5 minutes.

### DIFF
--- a/lib/modules/newCommentCount.js
+++ b/lib/modules/newCommentCount.js
@@ -77,8 +77,8 @@ module.options = {
 };
 
 const currentId: ?string = (execRegexes.comments(location.pathname) || [])[2];
-const lastCleanStorage = Storage.wrap('RESmodules.newCommentCount.lastClean', (null: null | number));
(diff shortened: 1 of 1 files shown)


lib/modules/newCommentCount.js
```diff
@@ -266,20 +272,19 @@ function stopTracking(id) {
...
 			if ((now - subscriptionDate) > (DAY * parseInt(module.options.subscriptionLength.value, 10))) {
 				unsubscribe(id);
 			}
-			// if we haven't checked this subscription in 5 minutes, try it again...
-			if ((now - lastCheck) > 300000) {
-				entryStorage.patch(id, { lastCheck: now });
-				checkThread(id, subscription);
-			}
+
+			entryStorage.patch(id, { lastCheck: now });
```
erikdesjardins: I don't think `lastCheck` is read anymore, so if that's the case might as well remove it from here and the type on line 89

PR #4351, opened 2017-07-17
newCommentCount: use sharded storage only for comment counts, not subscriptions

erikdesjardins:
Relevant issue: https://www.reddit.com/r/Enhancement/comments/6nn2kb/does_580_feel_slower_to_you/; related to #4348 (this is a more aggressive implementation)
Tested in browser: Chrome 60

~~Note: not yet tested, at all~~

### DIFF
--- a/lib/modules/newCommentCount.js
+++ b/lib/modules/newCommentCount.js
@@ -78,15 +78,23 @@ module.options = {
 
 const currentId: ?string = (execRegexes.comments(location.pathname) || [])[2];
 const lastCleanStorage = Storage.wrap('RESmodules.newCommentCount.lastClean', 0);
-const lastSubscriptionCheckStorage = Storage.wrap('RESmodules.newCommentCount.lastSubscriptionCheck', 0);
-const entryStorage = Storage.wrapPrefix('newCommentCount.', (): {
-	subscriptionDate?: number,
+const entryStorage = Storage.wrapPrefix('newCommentCount.', (): {|
 	count: number,
-	editedTime: number,
-	url: string,
-	title: string,
 	updateTime: number,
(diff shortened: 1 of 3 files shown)


lib/modules/newCommentCount.js
```diff
@@ -133,15 +134,19 @@ function setEntry(id: string, newCommentCount, newEditedTime) {
...
 	if (!module.options.monitorPostsVisitedIncognito.value && isPrivateBrowsing()) return false;
 
-	const patch = {
+	const now = Date.now();
+
+	entryStorage.set(id, {
+		count: newCommentCount,
+		updateTime: now,
+	});
+	subscriptionStorage.patch({ [id]: {
```
larsjohnsen: Seems like this can be moved into the `subscribe` function

lib/modules/newCommentCount.js
```diff
@@ -215,11 +220,13 @@ const getToggleSubscriptionButton = _.once(() =>
...
-	if (!currentId) throw new Error('No currentId');
+	const listingThing = getListingThing();
+	if (!listingThing) throw new Error('No current post');
 
 	const $button = getToggleSubscriptionButton();
-	const data = await entryStorage.getNullable(currentId);
-	if (data && data.subscriptionDate) {
+	const currentId = listingThing.getShortId();
+	const subscriptions = await subscriptionStorage.get();
+	if (currentId in subscriptions) {
```
larsjohnsen: otherwise this will not work correctly

lib/modules/newCommentCount.js
erikdesjardins: Done, thanks

PR #4355, opened 2017-07-18
selectedEntry: Set always enabled, add style "line".

larsjohnsen:
Since several modules now depend on `selectedEntry`, it's probably best to set it to `alwaysEnabled` in order to reduce complexity.

A new style `line` is added, as a less obtrusive alternative to outline and highlighting:
image

Users that earlier have disabled the module will be migrated to use it. I'm also considering changing to the default style to it, as I believe most users don't need the whole entry highlighted.

### DIFF
--- a/lib/modules/selectedEntry.js
+++ b/lib/modules/selectedEntry.js
@@ -17,7 +17,6 @@ import {
 	idleThrottle,
 } from '../utils';
 import type { ScrollStyle } from '../utils/dom';
-import * as Hover from './hover';
 
 export const module: Module<*> = new Module('selectedEntry');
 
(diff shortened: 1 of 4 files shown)


lib/modules/selectedEntry.js
```diff
@@ -375,50 +345,49 @@ export function move(direction: $Keys<typeof movers>, options?: SelectOptionsWit
...
 // why !important on .RES-keyNav-activeElement?  Because some subreddits are unfortunately using !important for no good reason on .entry divs...
 
-function addFocusBGColor() {
-	const focusBGColor = module.options.focusBGColor.value || module.options.focusBGColor.default;
-	const focusBGColorNight = module.options.focusBGColorNight.value || module.options.focusBGColorNight.default;
-	const focusFGColorNight = module.options.focusFGColorNight.value || module.options.focusFGColorNight.default;
-
+function styleLine() {
 	addCSS(`
+		.RES-keyNav-activeElement { box-shadow: 3px 0 0 -1px #c2d2e0; !important; }
```
larsjohnsen: Using `box-shadow` instead of `border` or `outline` avoids compressing the element.

PR #4365, opened 2017-07-26
i18n localStorage: Contain error when storage quota exceeded

larsjohnsen:
There's [been](https://www.reddit.com/r/RESissues/comments/6a8oy6/res_suddenly_stopped_working/) [some](https://www.reddit.com/r/RESissues/comments/6pkhiu/bug_res_dashboard_not_working_with_chrome/) [reports](https://www.reddit.com/r/RESissues/comments/6owj24/res_no_longer_working_my_subreddits_drop_down_not/) where RES fails to start because of this.

The issue appears to originate from Reddit, and seemingly also breaks some Reddit functionality like voting.

### DIFF
--- a/lib/environment/i18n.js
+++ b/lib/environment/i18n.js
@@ -32,7 +32,10 @@ export async function _loadI18n(): Promise<void> {
 		localStorage.setItem(CACHED_LANG_KEY, rawLocale());
 		localStorage.setItem(CACHED_MESSAGES_TOKEN_KEY, buildToken);
 	} catch (e) {
-		console.error('Could not cache i18n', e);
+		console.error('Could not cache i18n - RES will load VERY slowly', e);
+		// XXX: reddit has a habit of filling up these localstorage keys, so speculatively delete them
+		localStorage.removeItem('ads.adserverDownvotePixel');
+		localStorage.removeItem('ads.adserverUpvotePixel');
 	}
 }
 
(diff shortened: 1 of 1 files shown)


lib/environment/i18n.js
```diff
@@ -26,9 +26,14 @@ export async function _loadI18n(): Promise<void> {
...
-	localStorage.setItem(CACHED_MESSAGES_KEY, JSON.stringify(messages));
-	localStorage.setItem(CACHED_LANG_KEY, rawLocale());
-	localStorage.setItem(CACHED_MESSAGES_TOKEN_KEY, buildToken);
+
+	try { // Fails if remaining localStorage space is insufficient
+		localStorage.setItem(CACHED_MESSAGES_KEY, JSON.stringify(messages));
+		localStorage.setItem(CACHED_LANG_KEY, rawLocale());
+		localStorage.setItem(CACHED_MESSAGES_TOKEN_KEY, buildToken);
+	} catch (e) {
+		console.error('Could not cache i18n', e);
```
erikdesjardins: Hmm I don't really like this as-is because it'll make RES silently run slowly.
IMO we might as well fix it for the case we know about, even if it's a bit hacky.
(When the settings console is moved to a separate page we can probably split the i18n files, inline the non-option-related translations and get rid of this entirely)

I'm thinking:
- find the largest localStorage key
- if it's a known key (`ads.adserverDownvotePixel`, `ads.adserverUpvotePixel`, `ads.recent`) delete it
(or maybe just delete those keys unconditionally in this catch block, I doubt any users will complain about us deleting reddit's ad tracking)

PR #4367, opened 2017-07-30
Force Travis to use the Precise image for new Yarn version

erikdesjardins:
The recent Trusty update breaks the build: https://travis-ci.org/honestbleeps/Reddit-Enhancement-Suite/jobs/259170184/config

An update is planned, so hopefully it gets rolled out before Trusty is the only image; I really don't want to switch to sudo-enabled infra to update Yarn manually


mc10:
You can still use `sudo`-less builds for [updating Yarn](https://yarnpkg.com/lang/en/docs/install-ci/#travis-tab); just specify the specific version of Yarn you want.

erikdesjardins:
Good to know. That's better than this change; can you PR it if you have a minute?

PR #4368, opened 2017-07-30
Prevent Hover from closing when mouse is inside selected item

### LINKED ISSUE #3494: Multireddit navigation popup disappears before you can hover over it
Repro steps:
- visit a multi
- hover over the currently selected multi
- move mouse towards popup

This could be a Firefox only issue, @andytuba mentioned on irc:

> <andytuba> but in chrome it works fine for me if i move my mouse directly from the tab onto the menu

Browser: Firefox 49.0.1 x64, RES: 5.0.2.

Express50:
Fixes #3494.

Due to the extra padding on the selected navbar item, the hover div overlaps the selected item div when it is visible. This causes the `mouseleave` event to be fired (as seen [here](https://jsfiddle.net/1LwqLdsm/1/)). To fix this, I added a check before calling Hover.close() from the `mouseleave` event, as well as remove other unnecessary calls to Hover.close().

Relevant issue: #3494
Tested in browser: chrome

### DIFF
--- a/lib/modules/hover.js
+++ b/lib/modules/hover.js
@@ -161,7 +161,10 @@ class Hover {
 	}
 
 	target(element: HTMLElement) {
-		this._target = element;
+		if (this._target !== element) {
+			this.close();
+			this._target = element;
+		}
 
 		return this;
 	}
@@ -286,6 +289,7 @@ class Hover {
 		this.visible = true;
 
 		$(this._target).on('mouseleave', () => this._startHideTimer());
+		$(this._target).on('mouseenter', () => this._cancelHideTimer());
 	}
 
 	_startHideTimer() {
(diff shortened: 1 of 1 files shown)


lib/modules/hover.js
```diff
@@ -161,7 +161,6 @@ class Hover {
 	}
 
 	target(element: JQuery | HTMLElement) {
-		this.close();
```
erikdesjardins: Removing this prevents you from hovering on something else while it's fading out:
temp

(it's more noticeable when moving between multireddits)

Express50:
@erikdesjardins can you please take another look at this? I think I fixed the issue with the hover not showing up again.

erikdesjardins:
Sorry about the delay, 5.8.5 took longer than expected.

Everything seems good now, thanks!

PR #4374, opened 2017-08-05
Support inbox split into /notification/ and /message/

andytuba:
"Notifications" (i.e. comments) are being split out of /message/ into their own feed/URLs. https://redd.it/6rmag5

unread counts showing in floater island:  (can someone help fix styling for the alignment please)
screen shot 2017-08-05 at 4 20 27 pm

unread

unread state/counts updating from other tabs
read

unread state/counts updating from Never-Ending Reddit
ner


lib/modules/orangered.js
```diff
@@ -205,48 +271,112 @@ const setupFaviconBadge = _.once(() => {
...
+		'nativeNewModmailButton',
+	];
+
+	const fb = {};
+	let fbe;
+
+	for (const e of nativeButtonList) {
+		if (nb[e]) {
+			fbe = $(nb[e]).clone()[0];
+			fb[e.replace(/^native/, 'floating')] = fbe;
```
erikdesjardins: This is sketchy, we could eliminate the need for it by replacing
```js
const orangeredElements = () => ({
	...nativeButtons(),
	...floatingButtons()
});
```
with
```js
const orangeredElements = () => ({
	native: nativeButtons(),
	floating: floatingButtons()
});
```
and making `orangeredElements` and make `floatingElements` and `nativeElements` both return `{ messageButton, messageCount, notificationButton, notificationCount ... }`.
Alternatively remove `orangeredElements` and do the same.

...but that would require changing enough other stuff that whatever, we can leave it as-is.

lib/modules/orangered.js
```diff
@@ -205,48 +271,112 @@ const setupFaviconBadge = _.once(() => {
...
 
-	return '/message/inbox/';
+	return `/${basepath}/inbox/`;
 }
-function getUnreadCount(container = document.body) {
-	const mailCount = container.querySelector('.message-count');
-	return mailCount && parseInt(mailCount.textContent, 10) || 0;
+
+function getUnreadCount(container) {
+	const nb = _nativeButtons(container, false);
```
erikdesjardins: Inline this?

lib/modules/orangered.js
```diff
@@ -205,48 +271,112 @@ const setupFaviconBadge = _.once(() => {
...
+		nativeMessageCount,
+		nativeNotificationButton,
+		nativeNotificationCount,
+		nativeModmailButton,
+		nativeNewModmailButton,
+	} = nb;
+
+	const messageCount = nativeMessageCount && parseInt(nativeMessageCount.textContent, 10) || 0;
+	const notificationCount = nativeNotificationCount && parseInt(nativeNotificationCount.textContent, 10) || 0;
+	const modmailHasMail = nativeModmailButton && nativeModmailButton.classList.contains('havemail') && 1 || 0;
```
erikdesjardins: Nit: `... ? 1 : 0`

lib/modules/orangered.js
```diff
@@ -205,48 +271,112 @@ const setupFaviconBadge = _.once(() => {
...
+	const newModmailHasMail = nativeNewModmailButton && nativeNewModmailButton.classList.contains('havemail') && 1 || 0;
```
erikdesjardins: Ditto

lib/modules/orangered.js
```diff
@@ -205,48 +271,112 @@ const setupFaviconBadge = _.once(() => {
...
+	const nativeButtonList = [
+		'nativeMessageButton',
+		'nativeMessageCount',
+		'nativeNotificationButton',
+		'nativeNotificationCount',
+		'nativeModmailButton',
+		'nativeNewModmailButton',
+	];
+
+	const fb = {};
```
erikdesjardins: s/fb/floatingElements/ or something

lib/modules/orangered.js
```diff
@@ -205,48 +271,112 @@ const setupFaviconBadge = _.once(() => {
...
 
 const $orangeredElements = () => Object.values(orangeredElements()).reduce(($all, ele) => $all.add(ele), $());
 
-const nativeButtons = _.once(() => {
-	const nativeInboxButton = document.querySelector('#header-bottom-right #mail');
-	let nativeInboxCount = document.querySelector('#header-bottom-right .message-count');
-	if (!nativeInboxCount && (module.options.updateCurrentTab.value || module.options.updateOtherTabs.value)) {
+const nativeButtons = _.once(() => _nativeButtons(document.body, true));
+const _nativeButtons = (container = document.body, addIfMissing) => {
+	const nativeMessageButton = container.querySelector('#header-bottom-right #mail, #header-bottom-right #message');
```
erikdesjardins: Inline these?

lib/modules/orangered.js
```diff
@@ -160,25 +171,80 @@ function updateTitle(count) {
...
-	$().add($floatingInboxButton).add(nativeInboxButton)
-		.attr('title', count ? 'new mail!' : 'No new mail')
-		.toggleClass('havemail', !!count)
-		.toggleClass('nohavemail', !count)
-		.attr('href', getInboxLink(count));
+function updateInboxElements(data) {
+	const [
+		messageCount,
+		notificationCount,
+		modmailHasMail,
```
erikdesjardins: This doesn't seem right; the third element here is `modmailHasMail`...

lib/modules/orangered.js
```diff
@@ -160,25 +171,80 @@ function updateTitle(count) {
...
-		.css('display', count ? 'inline-block' : 'none')
-		.attr('href', getInboxLink(count))
-		.attr('title', count ? 'new mail!' : 'No new mail')
-		.text(count)
-		.addClass('message-count');
+function updateMailCountElements(data) {
+	const [
+		messageCount,
+		notificationCount,
+		messageTitle,
```
erikdesjardins: ...but here is `messageTitle`. (or replace with object)

lib/modules/orangered.js
```diff
@@ -130,16 +130,27 @@ export function updateFromPage(doc?: HTMLElement) {
...
-
-	updateFaviconBadge(count);
-	updateTitle(count);
-	updateInboxElements(count);
-	updateMailCountElements(count);
-
-	if (count > 0) {
-		window.dispatchEvent(new Event('orangered'));
+const _setUnreadCount = multicast(data => {
+	const [
```
erikdesjardins: This should probably be an object? (see below)

lib/modules/orangered.js
```diff
@@ -130,16 +130,27 @@ export function updateFromPage(doc?: HTMLElement) {
...
+	const sumCount = messageCount + notificationCount;
+
+	updateFaviconBadge(sumCount);
+	updateTitle(sumCount);
+	updateInboxElements(data);
+	updateMailCountElements(data);
+
+	if (sumCount > 0) {
+		const event = new Event('orangered');
+		event.data = {
```
erikdesjardins: Event objects aren't shared, so this won't do anything. (Flow is MVP 😉)

This is what you want:
`const event = new CustomEvent('orangered', { detail: { sumCount, messageCount, notificationCount } });` 
(Chrome clones the object and Firefox drops it--FF only allows `detail` to be a primitive type, nothing we can do)

lib/modules/orangered.js
```diff
@@ -205,48 +271,112 @@ const setupFaviconBadge = _.once(() => {
...
+	const notificationCount = nativeNotificationCount && parseInt(nativeNotificationCount.textContent, 10) || 0;
+	const modmailHasMail = nativeModmailButton && nativeModmailButton.classList.contains('havemail') && 1 || 0;
+	const newModmailHasMail = nativeNewModmailButton && nativeNewModmailButton.classList.contains('havemail') && 1 || 0;
+
+	const messageTitle = nativeMessageButton && nativeMessageButton.getAttribute('title') || '';
+	const notificationTitle = nativeNotificationButton && nativeNotificationButton.getAttribute('title') || '';
+	const modmailTitle = nativeModmailButton && nativeModmailButton.getAttribute('title') || '';
+	const newModmailTitle = nativeNewModmailButton && nativeNewModmailButton.getAttribute('title') || '';
+
+	return [
```
erikdesjardins: Why not an object?

lib/modules/orangered.js
```diff
@@ -130,16 +130,27 @@ export function updateFromPage(doc?: HTMLElement) {
...
-	count = count || 0;
-
-	updateFaviconBadge(count);
-	updateTitle(count);
-	updateInboxElements(count);
-	updateMailCountElements(count);
-
-	if (count > 0) {
-		window.dispatchEvent(new Event('orangered'));
+const _setUnreadCount = multicast(data => {
```
erikdesjardins: The callers of this function still use `_setUnreadCount(count)`, they should probably be renamed (can't comment on those lines)

lib/css/modules/_orangered.scss
```diff
@@ -1,21 +1,8 @@
-#NREMail {
-	width: 15px;
-	height: 10px;
-	margin-left: 4px;
-	background: center center no-repeat;
-
-	&.nohavemail {
-		background-image: url('../images/mailgray.png');
```
erikdesjardins: Delete these images if we're not using them anymore

lib/css/modules/_orangered.scss
```diff
@@ -1,21 +1,8 @@
...
-}
-
-#NREMailCount {
-	line-height: 12px;
-	margin: 0 0 0 1px;
+.res-floater-visibleAfterScroll {
+	.message-count,
+	.notification-count {
+		margin: 0 0 0 1px;
+	}
```
erikdesjardins:
```
lib/css/modules/_orangered.scss
  5:3  error  No trailing whitespace allowed  no-trailing-whitespace
```

lib/css/modules/_floater.scss
```diff
@@ -15,4 +16,8 @@
 	display: flex;
 	opacity: .7;
 	&:hover { opacity: 1; }
+
```
erikdesjardins:
```
lib/css/modules/_floater.scss
  18:25  error  No trailing whitespace allowed          no-trailing-whitespace
  21:16  error  Don't include leading zeros on numbers  leading-zero
```

Hmm the `leading-zero` rule is kinda silly, you could turn it off (`.sass-lint.yml`).

lib/modules/orangered.js
```diff
@@ -205,48 +271,112 @@ const setupFaviconBadge = _.once(() => {
...
 });
 
 const $orangeredElements = () => Object.values(orangeredElements()).reduce(($all, ele) => $all.add(ele), $());
 
-const nativeButtons = _.once(() => {
-	const nativeInboxButton = document.querySelector('#header-bottom-right #mail');
-	let nativeInboxCount = document.querySelector('#header-bottom-right .message-count');
-	if (!nativeInboxCount && (module.options.updateCurrentTab.value || module.options.updateOtherTabs.value)) {
+const nativeButtons = _.once(() => _nativeButtons(document.body, true));
+const _nativeButtons = (container = document.body, addIfMissing) => {
```
erikdesjardins: Super nit: all other (well, except `orangeredElements`) top-level functions are function declarations when they can be, why not this one too?

lib/css/modules/_orangered.scss
```diff
@@ -1,21 +1,8 @@
...
-
-	&.havemail {
-		background-image: url('../images/mail.png');
-	}
-}
-
-#NREMailCount {
-	line-height: 12px;
-	margin: 0 0 0 1px;
+.res-floater-visibleAfterScroll {
```
larsjohnsen: The fix the alignment, `top` and `margin-bottom` has to be reset.

lib/modules/orangered.js
andytuba: These shouldn't be inlined because some of the elements are used to find other buttons elements in `getOrAddInboxCount`.  It also makes it easier to copy-paste the object for destructuring later.

lib/modules/orangered.js
```diff
@@ -205,48 +271,112 @@ const setupFaviconBadge = _.once(() => {
...
-		$floatingInboxCount.attr('href', getInboxLink(true));
-		Floater.addElement($floatingInboxCount);
+	if (!(pinHeader === 'sub' || pinHeader === 'none')) {
+		return DEFAULT;
 	}
 
-	return { $floatingInboxButton, $floatingInboxCount };
+	const nb = nativeButtons();
+
+	const nativeButtonList = [
```
andytuba: I explicitly listed them to control the ordering because I don't trust `Object.keys()` to be consistent cross-browser.

lib/modules/orangered.js
andytuba: How about the good ol' `!!(castedToBoolean)`

lib/modules/orangered.js
andytuba: Yeah, I painted myself into a corner there 😭

lib/css/modules/_orangered.scss
andytuba: screen shot 2017-08-07 at 9 34 25 pm

Is ... less bad now.

lib/modules/orangered.js
erikdesjardins: The spec guarantees it, and I'm confident that the versions of the browsers we support comply with it (IE and old browsers probably don't).

[Object.keys()](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-object.keys) > [EnumerableOwnProperties](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-enumerableownproperties) > [\[\[OwnPropertyKeys\]\]](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys) > [OrdinaryOwnPropertyKeys](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-ordinaryownpropertykeys): "...in ascending chronological order of property creation..."

andytuba:
@XenoBen reports new notification icon not getting copied into floater, and new modmail isn't updating cross-tab.  also that dumb lint issue.

lib/css/modules/_floater.scss
```diff
@@ -3,6 +3,7 @@
 	will-change: transform;
 	z-index: 10000000;
 	right: 8px;
+	line-height: 12px;
```
erikdesjardins: This breaks the comment nav box, not immediately sure the best way to fix it

andytuba:
This will probably never happen in r2.

PR #4375, opened 2017-08-05
Remove hard-coded config for expando scrolling

beck:
This fixes a regression from:
https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/19028db28c5d24508ec647587426a484002237ab#diff-81e82c385dc9912064f4bc5e91b47e58L755

Before:
```
if (options.mediaBrowseScrollStyle) options.scrollStyle = options.mediaBrowseScrollStyle;
```

After:
```
options.scrollStyle = 'top';
```

Hard-coding config causes a regression of issue:
https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2628

Issue summary:
Using `scrollStyle = 'top'` causes expanded items to scroll up and off the
top of the window when `on hide move down` is used.

Not sure why this was hard-coded. My guess: a desire to have a different
scroll style for navigating comments vs media. If so, then a new setting needs
to be introduced. Until then, stop hardcoding and use the settings.

Relevant issue: 
Tested in browser:


larsjohnsen:
To perhaps solve the underlying issue which is that `hide` is delayed and causes the viewport to move, enable [`hideLinkInstant`](https://www.reddit.com/#res:settings/betteReddit/hideLinkInstant).

beck:
How should ` hideLinkInstant` behave? Toggling it on/off I do not see any change (also see a silly bug where the setting itself disappears - [video](https://dl.dropboxusercontent.com/u/1728414/2017%20RES%20hide%20link%20goodbye.mp4)).

Regardless, I think this pull is still valuable. When browsing without hiding (simply using `j` to step trough expanded media) my preference is to use "adopt top", not "top".

larsjohnsen:
The `hideLinkInstant` option is dependent on `fixHideLinks`, which if disabled causes the bug you see.

I agree that it would be worthwhile with an option here. Having `top` always be applied is a good default, but I understand that there's situations disabling manual control is desirable. Among other deficiencies, media browse mode does not behave nicely when switching from an expando post to a non-expando post, in which case some may feel disabling the auto-scroll altogether is the best solution.

beck:
@BeatLink no. I've been loading the extension from source, using this branch, occasionally rebasing off master ¯\_(ツ)_/¯

PR #4382, opened 2017-08-21
userTagger: Clean-up, split votes into ups and downs

larsjohnsen:
- When modifying a users tag, change is reflected on all instances of that tag
- `votes` is splitted into `votesDown` and `votesUp`, so it will be more useful for a future `userVoteWeight` filter
- VoteWeight is more useful: The opacity is the proportional to the ratio of the sum of votes to total number of votes.
- Utilizes `Hover.infocard` for tagging dialog
- `Tag` can be reused for a future subredditTagger module
- Performance improvement by cloning exisiting tag icon
- Always show voteWeight tooltip - remove minimially useful option
- Make the 'saveSourceLink' option useful
- Fix tagging via cli

Tested in browser: Chrome 61

### DIFF
--- a/lib/core/migrate/migrate.js
+++ b/lib/core/migrate/migrate.js
@@ -557,6 +557,22 @@ const migrations = [
 			}
 			Storage.set('RESmodules.newCommentCount.subscriptions', subscriptions);
 		},
+	}, {
+		versionNumber: '5.9.1-userTagger-refresh',
+		async go() {
+			let tags = await Storage.wrapPrefix('tag.', (): any => ({})).getAll();
+			for (const tag of Object.values(tags)) {
+				if (tag.votes > 0) tag.votesUp = tag.votes;
+				if (tag.votes < 0) tag.votesDown = -tag.votes;
+				Reflect.deleteProperty(tag, 'votes');
+				if (tag.tag) tag.text = tag.tag;
+				Reflect.deleteProperty(tag, 'tag');
+				if (tag.ignore) tag.ignored = tag.ignore;
+				Reflect.deleteProperty(tag, 'ignore');
+			}
+			tags = _.mapKeys(tags, (v, k) => `tag.${k}`); // Restore prefix
+			await Storage.setMultiple(tags);
+		},
 	},
 
 ];  // ^^ Add new migrations ^^
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -114,93 +114,7 @@ module.beforeLoad = () => {
 module.go = () => {
 	// If we're on the dashboard, add a tab to it...
 	if (isCurrentSubreddit('dashboard')) {
-		// add tab to dashboard
-		const $tabPage = Dashboard.addTab('userTaggerContents', i18n('userTaggerMyUserTags'), module.moduleID);
-		// populate the contents of the tab
-		const $showDiv = $(string.html`<div class="show">${i18n('userTaggerShow')} </div>`);
-		const $tagFilter = $(string.html`<select id="tagFilter"><option>${i18n('userTaggerTaggedUsers')}</option><option>${i18n('userTaggerAllUsers')}</option></select>`);
-		$($showDiv).append($tagFilter);
-		$tabPage.append($showDiv);
-		$('#tagFilter').change(() => drawUserTagTable());
-
-		const tagsPerPage = parseInt(Dashboard.module.options.tagsPerPage.value, 10);
-		if (tagsPerPage) {
-			const controlWrapper = document.createElement('div');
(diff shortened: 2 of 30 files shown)


lib/core/migrate/migrate.js
```diff
@@ -557,6 +557,22 @@ const migrations = [
...
 			Storage.set('RESmodules.newCommentCount.subscriptions', subscriptions);
 		},
+	}, {
+		versionNumber: '5.9.1-userTagger-refresh',
+		async go() {
+			let tags = await Storage.wrapPrefix('tag.', (): any => ({})).getAll();
+			for (const tag of Object.values(tags)) {
+				if (tag.votes > 0) tag.votesUp = tag.votes;
+				if (tag.votes < 0) tag.votesDown = -tag.votes;
+				Reflect.deleteProperty(tag, 'votes');
```
erikdesjardins: If you wrote this before I changed the rule to only forbid `\.call` and `\.apply`, you can change it back to `delete`. If you prefer this that's fine too.

lib/modules/userTagger.js
```diff
@@ -87,295 +80,396 @@ module.options = {
 		advanced: true,
 		dependsOn: options => options.trackVoteWeight.value,
 	},
-	vwTooltip: {
-		title: 'userTaggerVwTooltipTitle',
```
erikdesjardins: Remove the i18n string for this

lib/modules/userTagger.js
```diff
@@ -87,295 +80,396 @@ module.options = {
...
-		type: 'boolean',
-		value: true,
-		description: 'userTaggerVWTooltipDesc',
```
erikdesjardins: Ditto

lib/modules/userTagger.js
```diff
@@ -87,295 +80,396 @@ module.options = {
...
+	}
+}
+
+function getUsername(element: HTMLAnchorElement): string {
+	if (element.href) {
+		const test = element.href.match(usernameRE);
+		if (test) {
+			return test[1];
+		} else {
+			throw new Error(`Regex failed on ${element.href}, returning.`);
```
erikdesjardins: Nit: remove `, returning`

lib/modules/userTagger.js
```diff
@@ -87,295 +80,396 @@ module.options = {
...
+	render(instance: *) {
+		if (instance.vw) instance.vw.remove();
+		if (instance.renderVoteWeight && (this.votesUp || this.votesDown)) {
+			instance.vw = string.html`
+				<a
+					class="voteWeight"
+					href="javascript:void 0"
+					title="${i18n('userTaggerYourVotesFor', this.id, `+${this.votesUp} -${this.votesDown}`)}"
+					style="${getVoteWeightStyle(this)}"
+				>${module.options.vwNumber.value && `[${this.votes > 0 ? '+' : ''}${this.votes}]` || '[vw]'}</a>
```
erikdesjardins: ```module.options.vwNumber.value ? `[${this.votes > 0 ? '+' : ''}${this.votes}]` : '[vw]'```

lib/modules/userTagger.js
```diff
@@ -87,295 +80,396 @@ module.options = {
...
+		openLink: downcast(body.querySelector('.userTaggerOpenLink a'), HTMLAnchorElement),
+		preview: downcast(body.querySelector('#userTaggerPreview'), HTMLElement),
+		save: downcast(body.querySelector('#userTaggerSave'), HTMLElement),
+		text: downcast(body.querySelector('#userTaggerText'), HTMLInputElement),
+		votesDown: downcast(body.querySelector('#userTaggerVotesDown'), HTMLInputElement),
+		votesUp: downcast(body.querySelector('#userTaggerVotesUp'), HTMLInputElement),
+	};
+
+	let ignored = tag.ignored;
+	$(body.querySelector('.res-usertag-ignore label')).after(
```
erikdesjardins: The native `body.querySelector('.res-usertag-ignore label').after(` might work

erikdesjardins:
Hmm, I'm not sure why the [number of tips changed](https://travis-ci.org/honestbleeps/Reddit-Enhancement-Suite/jobs/276309521#L2041).

If it's intentional / the order is random then you can change/remove [the assertion](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/ea236626ffec84c3dab84cc31e172e0adeb04c67/tests/RESTips.js#L24).

larsjohnsen:
@erikdesjardins The test as is actually indicated a real bug. A userTagger tip erroneously referred to the "vwTooltip" option, so when the i18n-string was gone, the guider wasn't produced.

erikdesjardins:
Tested migration

PR #4388, opened 2017-08-24
Fixed sub-tagger interfering with new "crosspost" function

corylulu:
The crosspost pulls the inner html from the .thing to construct the recommended title. Moving it outside of the <a> tag fixes the problem. 
http://i.imgur.com/jqm3ttc.png

Relevant issue: 
Tested in browser:


corylulu:
Can't figure out what failed on the test here

erikdesjardins:
Looks good, thanks

corylulu:
@erikdesjardins Wasn't sure if it would be desireable to make another link surrounding it to still make it clickable or not, but just to get a quick fix, I just pushed it like this. What do you think?

PR #4397, opened 2017-08-27
Implement i18n in RES gear menu and some top-level interface items

wojtekmaj:
Relevant issue: 
Tested in browser:

### DIFF
--- a/lib/modules/RESTips.js
+++ b/lib/modules/RESTips.js
@@ -52,7 +52,7 @@ module.go = () => {
 	if (module.options.menuItem.value) {
 		const $menuItem = $('<div>', {
 			id: 'RESTipsMenuItem',
-			text: 'tips & tricks',
+			text: i18n('tipsAndTricks'),
 		});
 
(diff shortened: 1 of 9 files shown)


andytuba:
Neat! You can probably reuse some existing strings, e.g. module titles or options, instead of adding new keys to strings to translate -- and using CSS to handle the formatting eg for lowercase menu items.

wojtekmaj:
Actually this is what I did initially and then decided not to, because gear menu is lowercase. And I couldn't use just `text-transform: lowercase` as that would also change "RES" to "res" and such.

wojtekmaj:
Well, I could reuse *some* like night mode, but it's up to you - I thought it would be clearer to have them all in one place. Plus, we might want to name the toggle differently in the future.

andytuba:
Good point. Now that I'm looking into the actual strings, I agree with your choice to just moving the existing text into the localized strings.

erikdesjardins:
Looks good!

Issue #4417, opened 2017-09-08
Move RES hover card over to native hovercard


BenMcGarry:
https://www.reddit.com/r/changelog/comments/6yqqri/testing_out_user_hover_cards/

PR #4418, opened 2017-09-08
Module to hide the new reddit hovercards

BenMcGarry:
As a temporary solution until #4417 is completed.

### DIFF
--- a/lib/css/modules/_redditUserInfo.scss
+++ b/lib/css/modules/_redditUserInfo.scss
@@ -1,0 +1,5 @@
+.res-redditUserInfo-hideAuthorTooltip {
+	.author-tooltip {
+		display: none !important;
+	}
+}
--- a/lib/modules/redditUserInfo.js
+++ b/lib/modules/redditUserInfo.js
@@ -1,0 +1,18 @@
+/* @flow */
+
+import { Module } from '../core/module';
+
+export const module: Module<*> = new Module('redditUserInfo');
+
+module.moduleName = 'redditUserInfoName';
+module.category = 'usersCategory';
+module.description = 'redditUserInfoDesc';
+module.options = {
+	hideAuthorTooltip: {
+		type: 'boolean',
+		value: true,
+		description: 'redditUserInfoHideDesc',
+		title: 'redditUserInfoHideTitle',
+		bodyClass: true,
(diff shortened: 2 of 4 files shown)


lib/css/modules/_redditUserInfo.scss
```diff
@@ -0,0 +1,5 @@
+.res-redditUserInfo-hideAuthorTooltip {
```
larsjohnsen: This file must be imported [here](https://github.com/BenMcGarry/Reddit-Enhancement-Suite/blob/476b6dceb1eaaeb904470a4f323f3f02e3b916b7/lib/css/res.scss#L26)

lib/modules/redditUserInfo.js
```diff
@@ -0,0 +1,18 @@
+/* @flow */
+
+import { Module } from '../core/module';
+
+export const module: Module<*> = new Module('redditUserInfo');
+
+module.moduleName = 'redditUserInfoName';
+module.category = 'usersCategory';
+module.description = 'redditUserInfoDesc';
```
larsjohnsen: Since the module only does one thing, you can remove the option and instead add `module.bodyClass = true` here.

BenMcGarry: @andytuba Do we have any other plans for this module?

andytuba: I was thinking this module could also include options for adding RES's various userinfo buttons to the new hovercard. We could leave it as-is or migrate to the option later, but I'm inclined to prepare for more options later.

larsjohnsen: Anyway, requiring the module for extending Reddit's hovercard to be enabled in order to disable it is a little awkward. I think this option is better suited for betteReddit.

BenMcGarry: I agree it is a little awkward, but this module is only going to be used as a bridge before we move over to the native hovercards. So it would probably be here until the next major release, by then we will have moved across.

erikdesjardins: Agree with the awkwardness, but also agree that we should be able to remove (or possibly repurpose) this module entirely fairly soon...for a certain definition of soon...so seems fine for now.

PR #4419, opened 2017-09-12
accountSwitcher: Add 2FA support

### LINKED ISSUE #4403: Add support for 2FA on account switcher.
It is possible to submit the 2FA prompt code with the password as `password:code`.

We need to add an option to ask for the code when a user switches that is then submitted with the password.

> 22:51:56 <@andytuba> yeah of course we'll have to prompt for code
> 22:52:09 <@andytuba> that's the obnoxious part
> 22:52:23 <@andytuba> i guess we could do a simple modal prompt
> 22:53:13 <@andytuba> incidentally the existing implementation on desktop is two step
> 22:53:34 <@andytuba> send username/password, API responds with "need 2FA", send 2FA token, success
> 22:55:52 <@XenoBen> Hmm
> 22:56:37 <@andytuba> lemme see if we can bundle it together too, to support "user has indicated or a previous response has indicated this account eneds 2fa"
> 22:57:36 <@andytuba> okay cool we can send user, passwd, and otp (2fa code) all in one request
> 22:57:40 <@andytuba> to the regular login endpoint
> 22:58:58 <@andytuba> so RES can prompt up front if user indicated account has 2Fa enabled, or if "login failed: 2fa required" error then RES can prompt for code and send again wit hcode (and update prefs to indicate 2Fa)

larsjohnsen:
Prompts the user for a PIN if 2FA is marked enabled for the selected account.

Closes #4403

### DIFF
--- a/lib/modules/accountSwitcher.js
+++ b/lib/modules/accountSwitcher.js
@@ -37,7 +37,7 @@ module.options = {
 			type: 'password',
 		}, {
 			key: '2fa',
-			name: 'accountSwitcherUses2Fa',
+			name: 'accountSwitcherRequiresOtp',
 			type: 'boolean',
 			value: false,
 		}],
@@ -317,14 +317,16 @@ function findMatchingAccount(username, partialMatch) {
 async function switchTo(user) {
 	const account = findMatchingAccount(user);
 	if (!account) return;
-	const [username, password, uses2Fa] = account;
+	const [username, password, requiresOtp] = account;
 
-	// Don't await promise yet so that the PIN can be entered while logging out
+	// Don't await promise yet so that the OTP can be entered while logging out
 	const logoutPromise = loggedInUser() && ajax({ method: 'POST', url: '/logout' });
 
-	let pin;
-	if (uses2Fa) {
-		pin = window.prompt(`Enter 2FA PIN for user ${user}`);
+	let otp;
+	if (requiresOtp) {
+		otp = {
+			otp: window.prompt(i18n('accountSwitcherOptPrompt', username)),
+		};
 	}
 
 	await logoutPromise;
@@ -334,7 +336,8 @@ async function switchTo(user) {
 		url: '/api/login',
 		data: {
 			user: username,
-			passwd: pin ? `${password}:${pin}` : password,
+			passwd: password,
+			...otp,
 			rem: module.options.keepLoggedIn.value ? 'on' : 'off',
 		},
 		type: 'json',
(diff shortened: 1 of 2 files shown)


lib/modules/accountSwitcher.js
```diff
@@ -312,21 +317,24 @@ function findMatchingAccount(username, partialMatch) {
...
 
+	await logoutPromise;
+
 	const { success, jquery } = await ajax({
 		method: 'POST',
 		url: '/api/login',
 		data: {
 			user: username,
-			passwd: password,
+			passwd: pin ? `${password}:${pin}` : password,
```
andytuba: Alternatively, this could be more structured:
```
{
    user: username,
    passwd: password,
    otp_sent: pin,
    otp: pin,
}
```

lib/modules/accountSwitcher.js
```diff
@@ -312,21 +317,24 @@ function findMatchingAccount(username, partialMatch) {
...
-		await ajax({
-			method: 'POST',
-			url: '/logout',
-		});
+	// Don't await promise yet so that the PIN can be entered while logging out
+	const logoutPromise = loggedInUser() && ajax({ method: 'POST', url: '/logout' });
+
+	let pin;
+	if (uses2Fa) {
+		pin = window.prompt(`Enter 2FA PIN for user ${user}`);
```
andytuba: `i18n("Enter the 6 digit code from your authenticator app for u/%s", username)`

(except properly formatted for RES)

andytuba:
Looking pretty good, thank!

It would be a great improvement (but could await a future PR) to catch the "needs login OTP" error and then 1) turn on "needs 2fa" in settings and 2) ask for OTP and attempt login again

lib/modules/accountSwitcher.js
```diff
@@ -312,21 +317,24 @@ function findMatchingAccount(username, partialMatch) {
...
 
-	if (loggedInUser()) {
-		await ajax({
-			method: 'POST',
-			url: '/logout',
-		});
+	// Don't await promise yet so that the PIN can be entered while logging out
+	const logoutPromise = loggedInUser() && ajax({ method: 'POST', url: '/logout' });
+
+	let pin;
```
andytuba: Should probably be named `otp`

lib/modules/accountSwitcher.js
larsjohnsen: Setting `otp_sent` to false seemingly didn't make the server ignore the `otp` field.

PR #4429, opened 2017-09-18
Gfycat detail route regex.

kennethjeremyau:
Relevant issue: Updating regex for parsing gfyid from details page.
Tested in browser: Chrome 61.0.3163.91


erikdesjardins:
Thanks!

(build failure is spurious)

Issue #3128, opened 2016-06-30
userbar hider should use bodyclass option


andytuba:
instead of `$.fn.toggle`

https://www.reddit.com/r/RESissues/comments/4qnfyr/bug_empty_messagecount_not_hidden_when_userbar_is/

prayashm:
Hi, I want to solve this issue.
Just to confirm, when the unread count is 0, the counter should be hidden right?

andytuba:
Right, for the counter in the user bar. (there's another one supplied by the orangered module, it floats in the top right corner.)  I wouldn't think you'd need to worry about that level of logic as part of this issue, since I expect this feature  to simply hide everything in the user menu except the reveal / expand button itself. 

Thanks for looking into this!

PR #4436, opened 2017-09-25
Exempt message-count from elements list

### LINKED ISSUE #4435: Orangered shown on userbar toggle despite no new mail
Clicking the Toggle Userbar button twice to toggle the userbar off and on again shows the mail icon with an orange [0] next to it like this: https://i.imgur.com/3d2SPXK.png

isiah-lloyd:
`.message-count` from the module `orangered` is defaulted to `display: hidden`. The module `userbarHider` sets all elements in the userbar to visible when the userbar is shown, thus causing the new message indicator to show even when there isn't a new message.
Relevant issue: Closes #4435
Tested in browser: Chrome

### DIFF
--- a/lib/modules/userbarHider.js
+++ b/lib/modules/userbarHider.js
@@ -45,7 +45,6 @@ module.options = {
 };
 
 let userbar, $userbarToggle;
-
 module.go = () => {
 	userbarHider();
 };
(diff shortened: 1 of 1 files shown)


lib/modules/userbarHider.js
```diff
@@ -102,8 +102,8 @@ function updateToggleButton(userbarHidden) {
 }
 
 function toggleUserbarElementsDisplay(userbarHidden) {
-	const elements = $(userbar).children().not($userbarToggle);
-
+	const elements = $(userbar).children().not($userbarToggle).not('.message-count');
```
erikdesjardins: I don't think this will work properly when you have new messages; it won't hide the message-count icon.

I think the method described in #3128 will work better; that is use `BodyClasses.toggle(userbarHidden, 'res-hide-userbar');` here, and to the bottom of `_userbarHider.scss` add

```scss
.res-hide-userbar {
    #header-bottom-right > :not(#userbarToggle)
}
```

(or something along those lines)

PR #4437, opened 2017-09-25
Fix inaccurate viewport size in certain subreddits

### LINKED ISSUE #4420: Auto-scrolling on keyboard nav doesn't work on /r/NFL
- My browser (copy from https://www.whatismybrowser.com): Chrome 60
- My RES version: 5.8.4

*What's the problem?*

The auto-scrolling that should occur when I keyboard-navigate to an element that's off the visible page does not occur. Open any large comment thread on the front page and try to scroll between top-level comments with Ctrl-j or Ctrl-k. The page doesn't scroll. Tested with both legacy and "middle" scrolling option.

I tried looking at this in the debugger and it seemed to have something to do with the viewport being calculated as having a very large height (in the ballpark of 25000). However, I'm not familiar enough with Chrome development or the RES codebase to track it down further.

Not that I could find.

I confirmed this bug in an incognito window with only RES.

leviroth:
Relevant issue: Fixes #4420 
Tested in browser: Chrome

This changes `getViewportSize` to use `document.documentelement.client[Width|Height]` instead of `getBoundingClientRect`, which was sometimes giving the size of the entire document instead of the size of the viewport.

I chose the new technique based on [this Stack Overflow answer](https://stackoverflow.com/questions/1248081/get-the-browser-viewport-dimensions-with-javascript/8876069#8876069). However, I'm pretty apprehensive about the change, since there appear to be a few different approaches and they don't seem to be terribly well documented.

### DIFF
--- a/lib/utils/dom.js
+++ b/lib/utils/dom.js
@@ -64,15 +64,13 @@ click.isProgrammaticEvent = (e: MouseEvent): boolean => e.clientX === 1 && e.cli
 export const getViewportSize = _.memoize((): { width: number, height: number } => {
 	waitForEvent(window, 'resize').then(() => { getViewportSize.cache.clear(); });
 
-	// Temporarily append a dummy node to document.body
-	const tempNode = document.createElement('div');
-	tempNode.style.width = tempNode.style.height = '100%';
-	tempNode.style.position = 'fixed';
-	document.body.appendChild(tempNode);
-
-	const result = _.pick(tempNode.getBoundingClientRect(), ['height', 'width']);
-	document.body.removeChild(tempNode);
-	return result;
+	const viewportSizedElement = document.createElement('div');
+	viewportSizedElement.style.width = viewportSizedElement.style.height = '100%';
+	viewportSizedElement.style.position = 'fixed';
+	document.body.appendChild(viewportSizedElement);
+	const size = _.pick(viewportSizedElement.getBoundingClientRect(), ['height', 'width']);
+	viewportSizedElement.remove();
+	return size;
 });
 
 export function elementInViewport(ele: Element): boolean {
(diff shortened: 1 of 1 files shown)


erikdesjardins:
This seems okay but it may have been done for some subtle reason.

@larsjohnsen if you think this is good then go ahead and merge.

larsjohnsen:
Using `client(Width|Height)` would generally be the correct approach when determining the viewport size, as I see `getBoundingClientRect` only works because Reddit by default sets `html { height: 100%; }`.

The `client`-properties returns a rounded integers, which can cause issues. When zooming using browser, the rect size can become fractional, and if rounded up, a element sized to correspond with `getViewportSize` causes a scrollbar to appear. That may happen with pictures in showImages which reaches `maxWidth`.

It appears that `window.getComputedStyle(document.documentElement)` returns the correct fractional values, and should work fine in this case since the function is memoized.

leviroth:
>It appears that `window.getComputedStyle(document.documentElement)` returns the correct fractional values, and should work fine in this case since the function is memoized.

That seems to have the same issue on /r/NFL where the `documentElement` has the height of the whole page rather than the height of the viewport.

larsjohnsen:
What about creating a temporary element `<div style="width: 100%;height: 100%;position: fixed;"></div>`, appending it to `document.body`, and measuring it?

leviroth:
That looks like it works.

mc10:
Should `visibility: hidden` or `display: none` be set to avoid (temporarily) covering up the page with this div?

larsjohnsen:
@mc10 Since the element is removed synchronously it will not be painted.

larsjohnsen:
Thanks, @leviroth

PR #4439, opened 2017-09-30
selectedEntry: Set always enabled, add style "line"

larsjohnsen:
Since several modules now depend on `selectedEntry`, it's probably best to set it to `alwaysEnabled` in order to reduce complexity.

A new style `line` is added, as a less obtrusive alternative to outline and highlighting:
image

Users that earlier have disabled the module will be migrated to use it.

Tested in: Firefox 57, Chrome 62
Re-creating the PR since I accidentally deleted the branch so #4355 became closed.

### DIFF
--- a/lib/modules/selectedEntry.js
+++ b/lib/modules/selectedEntry.js
@@ -17,7 +17,6 @@ import {
 	idleThrottle,
 } from '../utils';
 import type { ScrollStyle } from '../utils/dom';
-import * as Hover from './hover';
 
 export const module: Module<*> = new Module('selectedEntry');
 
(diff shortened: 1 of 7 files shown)


lib/modules/selectedEntry.js
```diff
@@ -251,88 +218,32 @@ function updateActiveElement(selected, last) {
...
-	if (lastSelected) {
-		return Thing.things().find(v => v.getFullname() === lastSelected);
-	}
+	select(target, { scrollStyle: scrollToSelected ? 'legacy' : 'none' });
 }
 
+const addLastSelectedListener = _.once(() => {
+	addListener(selected => {
+		history.replaceState({ ...history.state, lastSelectedId: selected.getFullname() }, document.title);
+	}, 'beforeScroll');
```
erikdesjardins: Should this be done at idle? Or is it too cheap to matter? I vaguely remember someone saying that replaceState/pushState were slow on underpowered machines, but I could be wrong.

larsjohnsen: `replaceState` is unfortunately fairly costly for this relatively unimportant feature, so I switched to using `sessionStorage` instead.

PR #4440, opened 2017-09-30
Google Drive backup provider: Remember which account to use if there are multiple Google accounts logged in

### LINKED ISSUE #4376: Google Drive backup provider does not remember which account to use if there are multiple Google accounts logged in
- My browser: Chrome 60 on Windows 10
- My RES version: v5.8.4

*What's the problem?*

When there's more than one Google account logged in, RES does not remember which account it should use to backup to Google Drive, constantly asking you.

Yes, see [this thread](https://www.reddit.com/r/RESissues/comments/6rd6m9/google_drive_sync_is_asking_me_to_confirm_my/) which also includes screenshots.

BetterTTV
Google Drawings
Google Play Music
HTTPS Everywhere
Proxy SwitchyOmega
Pushbullet
Secure Shell
uBlock Origin

AgrimPrasad:
As mentioned in #4376 , currently when there's more than one Google account logged in, RES does not remember which account it should use to backup to Google Drive, constantly asking you.

This PR fixes this issue by adding a `login_hint` field in the auth request. This requires an additional `https://www.googleapis.com/auth/userinfo.email` scope in the auth request.

The user's Google email address used to backup or restore for the first time is saved to local storage and re-used for subsequent backups or restores.

If the user logs out of the Google email address saved in local storage, then the local storage email address cache is invalidated and then re-written after the user re-authenticates with a Google account.

Relevant issue: Fixes #4376 
Tested in browser: Firefox 56, Chrome 61, Edge 40

~Outstanding Issue: Can someone help test this in Edge as I don't have ready access at present to this browser?~ Tested with Edge successfully.

### DIFF
--- a/firefox/manifest.json
+++ b/firefox/manifest.json
@@ -80,6 +80,7 @@
 		"https://api.tumblr.com/v2/blog/*/posts",
 		"https://xkcd.com/*/info.0.json",
 		"https://api.steampowered.com/ISteamRemoteStorage/GetPublishedFileDetails/*",
-		"https://content.googleapis.com/drive/v3/*"
+		"https://content.googleapis.com/drive/v3/*",
+		"https://www.googleapis.com/oauth2/v3/*"
 	]
 }
--- a/lib/environment/auth.js
+++ b/lib/environment/auth.js
@@ -1,11 +1,10 @@
 /* @flow */
 
 import { sendMessage } from '../../browser';
-import { CACHED_GOOGLE_EMAIL_ADDRESS } from '../constants/localStorage';
 import { Permissions } from './';
 
 export async function launchAuthFlow(
-	{ domain, clientId, loginHint, scope = ''}: {| domain: string, clientId: string,  loginHint?: string, scope?: string |},
+	{ domain, clientId, loginHint, scope = '' }: {| domain: string, clientId: string, loginHint?: string, scope?: string |},
 	warnUserInteraction: (message: string) => Promise<void>,
 ): Promise<string> {
 	let responseUrl;
--- a/lib/environment/auth.js
+++ b/lib/environment/auth.js
@@ -1,6 +1,7 @@
 /* @flow */
 
 import { sendMessage } from '../../browser';
+import { CACHED_GOOGLE_EMAIL_ADDRESS } from '../constants/localStorage';
 import { Permissions } from './';
 
 export async function launchAuthFlow(
@@ -12,6 +13,11 @@ export async function launchAuthFlow(
 		responseUrl = await sendMessage('authFlow', { domain, clientId, scope, loginHint, interactive: false });
 	} catch (e) {
 		console.error('Noninteractive auth failed:', e);
+
+		// Invalidate google email address
+		if (domain === 'https://accounts.google.com/o/oauth2/v2/auth' && loginHint) {
+			localStorage.removeItem(CACHED_GOOGLE_EMAIL_ADDRESS);
+		}
 
 		await warnUserInteraction('You may be redirected to redditenhancementsuite.com to complete the login process.');
 
--- a/lib/environment/auth.js
+++ b/lib/environment/auth.js
@@ -5,7 +5,7 @@ import { CACHED_GOOGLE_EMAIL_ADDRESS } from '../constants/localStorage';
 import { Permissions } from './';
 
 export async function launchAuthFlow(
-	{ domain, clientId, scope = '', loginHint = '' }: {| domain: string, clientId: string, scope?: string, loginHint?: string |},
+	{ domain, clientId, loginHint, scope = ''}: {| domain: string, clientId: string,  loginHint?: string, scope?: string |},
 	warnUserInteraction: (message: string) => Promise<void>,
 ): Promise<string> {
 	let responseUrl;
@@ -13,11 +13,6 @@ export async function launchAuthFlow(
 		responseUrl = await sendMessage('authFlow', { domain, clientId, scope, loginHint, interactive: false });
 	} catch (e) {
 		console.error('Noninteractive auth failed:', e);
-
-		// Invalidate google email address
-		if (domain === 'https://accounts.google.com/o/oauth2/v2/auth' && loginHint) {
-			localStorage.removeItem(CACHED_GOOGLE_EMAIL_ADDRESS);
-		}
 
 		await warnUserInteraction('You may be redirected to redditenhancementsuite.com to complete the login process.');
 
--- a/lib/environment/auth.js
+++ b/lib/environment/auth.js
@@ -27,7 +27,3 @@ export async function launchAuthFlow(
 
 	return token;
 }
-
-export async function saveEmail(emailAddress: string): Promise<void> {
-	await sendMessage('saveEmail', emailAddress);
-}
--- a/lib/environment/auth.js
+++ b/lib/environment/auth.js
@@ -4,12 +4,12 @@ import { sendMessage } from '../../browser';
 import { Permissions } from './';
 
 export async function launchAuthFlow(
-	{ domain, clientId, scope = '' }: {| domain: string, clientId: string, scope?: string |},
+	{ domain, clientId, scope = '', loginHint = '' }: {| domain: string, clientId: string, scope?: string, loginHint?: string |},
 	warnUserInteraction: (message: string) => Promise<void>,
 ): Promise<string> {
 	let responseUrl;
 	try {
-		responseUrl = await sendMessage('authFlow', { domain, clientId, scope, interactive: false });
+		responseUrl = await sendMessage('authFlow', { domain, clientId, scope, loginHint, interactive: false });
 	} catch (e) {
 		console.error('Noninteractive auth failed:', e);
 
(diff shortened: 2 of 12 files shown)


AgrimPrasad:
This bugfix has been tested in Chrome, Firefox and Edge. Should be ready for review and merge now.

firefox/manifest.json
```diff
@@ -80,6 +80,7 @@
 		"https://api.tumblr.com/v2/blog/*/posts",
 		"https://xkcd.com/*/info.0.json",
 		"https://api.steampowered.com/ISteamRemoteStorage/GetPublishedFileDetails/*",
-		"https://content.googleapis.com/drive/v3/*"
+		"https://content.googleapis.com/drive/v3/*",
+		"https://www.googleapis.com/oauth2/v3/*"
```
larsjohnsen: Can these be moved to optional permissions now? Seems that F[irefox ≥ 55 support that](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/optional_permissions#Browser_compatibility).

erikdesjardins: Theoretically; see the ff-perms branch in my fork. Though I always seem to get `Error: May only request permissions from a user input handler`. I assume it doesn't handle asynchronicity (which is going to be annoying), but calling `chrome.runtime.sendMessage` directly from an event listener doesn't fix it either...

AgrimPrasad: Since adding optional permissions in Firefox seems like a non-trivial task, how about we get this in first and then handle the permissions in a separate PR?

lib/modules/backupAndRestore/providers/GoogleDrive.js
```diff
@@ -28,9 +34,46 @@ export class GoogleDrive extends Provider {
 			await Permissions.request(['https://accounts.google.com/o/oauth2/v2/auth', 'https://content.googleapis.com/drive/v3/*']);
 		});
 
+		// Re-check email address cache, in case it got invalidated due to google account logout
+		if (!this.getCachedEmailAddress()) {
```
larsjohnsen: It doesn't seem like this will ever recheck after `emailAdress` for the first time.

lib/modules/backupAndRestore/providers/GoogleDrive.js
```diff
@@ -28,9 +34,46 @@ export class GoogleDrive extends Provider {
...
+			}
+		}
+
 		return this;
 	}
 
+	getCachedEmailAddress() {
+		const cachedEmailAddress = localStorage.getItem(CACHED_GOOGLE_EMAIL_ADDRESS);
+		if (cachedEmailAddress) {
+			return JSON.parse(cachedEmailAddress);
```
larsjohnsen: Nit: `loginHint` is a string, so converting it into JSON is unnecesssary.

lib/modules/backupAndRestore/providers/GoogleDrive.js
AgrimPrasad: Because the email address cache is invalidated here when a user logs out of their google account: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4440/files#diff-ad1ac56107331cb3882bedd9fd80ffb0R17

The user could get logged out of their google account for several reasons:
1. Manual google account log out.
2. Cookies cleared.
3. Google for Business accounts are automatically logged out after 30 days.

Furthermore, if after logout, the user chooses to log in with a different google account, then this would email address would need to be updated. Note that with the login hint, the user would be prompted by Google to log into the same account which was stored here in local storage, but it's up to the user to log into the same or different google account.

AgrimPrasad:
Is this ok for merging now? I think I've addressed the outstanding concerns. Please let me know if this is not the case, thanks.

lib/modules/backupAndRestore/providers/GoogleDrive.js
```diff
@@ -28,9 +34,46 @@ export class GoogleDrive extends Provider {
...
 		return this;
 	}
 
+	getCachedEmailAddress() {
+		const cachedEmailAddress = localStorage.getItem(CACHED_GOOGLE_EMAIL_ADDRESS);
+		if (cachedEmailAddress) {
+			return cachedEmailAddress;
+		}
+
+		return '';
```
mc10: These lines could just be `return cachedEmailAddress || '';`

AgrimPrasad: Thanks, nice catch. Had missed that while thinking about the JSON parsing.

lib/environment/auth.js
```diff
@@ -1,23 +1,29 @@
...
 	warnUserInteraction: (message: string) => Promise<void>,
 ): Promise<string> {
 	let responseUrl;
 	try {
-		responseUrl = await sendMessage('authFlow', { domain, clientId, scope, interactive: false });
+		responseUrl = await sendMessage('authFlow', { domain, clientId, scope, loginHint, interactive: false });
 	} catch (e) {
 		console.error('Noninteractive auth failed:', e);
 
+		// Invalidate google email address
```
larsjohnsen: It would be cleaner doing in `GoogleDrive.js`. Perhaps inside the `warnUserInteraction` callback?

lib/environment/auth.js
```diff
@@ -1,23 +1,29 @@
 /* @flow */
 
 import { sendMessage } from '../../browser';
+import { CACHED_GOOGLE_EMAIL_ADDRESS } from '../constants/localStorage';
 import { Permissions } from './';
 
 export async function launchAuthFlow(
-	{ domain, clientId, scope = '' }: {| domain: string, clientId: string, scope?: string |},
+	{ domain, clientId, scope = '', loginHint = '' }: {| domain: string, clientId: string, scope?: string, loginHint?: string |},
```
larsjohnsen: `loginHint` does not need a default value.

lib/modules/backupAndRestore/providers/GoogleDrive.js
```diff
@@ -28,9 +34,42 @@ export class GoogleDrive extends Provider {
...
+	setCachedEmailAddress(emailAddress: string) {
+		localStorage.setItem(CACHED_GOOGLE_EMAIL_ADDRESS, JSON.stringify(emailAddress));
+	}
+
+	async verifyToken(): Promise<string> {
+		const decryptedToken = await ajax({
+			method: 'GET',
+			url: 'https://www.googleapis.com/oauth2/v3/tokeninfo',
+			query: { access_token: this.accessToken },
+			headers: { 'Content-Type': 'application/json' },
```
larsjohnsen: Redundant when `type: 'json'` is specified.

lib/modules/backupAndRestore/providers/GoogleDrive.js
```diff
@@ -28,9 +34,42 @@ export class GoogleDrive extends Provider {
...
+
+	setCachedEmailAddress(emailAddress: string) {
+		localStorage.setItem(CACHED_GOOGLE_EMAIL_ADDRESS, JSON.stringify(emailAddress));
+	}
+
+	async verifyToken(): Promise<string> {
+		const decryptedToken = await ajax({
+			method: 'GET',
+			url: 'https://www.googleapis.com/oauth2/v3/tokeninfo',
+			query: { access_token: this.accessToken },
```
larsjohnsen: Can access token be specified as a header, like in the other calls?

lib/modules/backupAndRestore/providers/GoogleDrive.js
AgrimPrasad: Thanks, that makes sense. Should we also remove these redundant headers from other ajax calls where `type: 'json'` has already been specified, e.g. https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4440/files/3ccdb7164d81aa4cc06fea43edd18be54c6c2c42#diff-0a606b28b4a58094611f298cd827e0b9R94 ?

lib/modules/backupAndRestore/providers/GoogleDrive.js
AgrimPrasad: Unfortunately, no, the Google API specifies that this access token should be sent as a query parameter. Ref.: https://developers.google.com/identity/protocols/OAuth2UserAgent#validate-access-token

I also tried to send this info in an `Authorization` header, but got a 400 Bad Request response:

```
curl -X POST \
  https://www.googleapis.com/oauth2/v3/tokeninfo \
  -H 'authorization: Bearer <access token...>' \
  -H 'cache-control: no-cache' \
  -H 'postman-token: c7c9317b-2089-e3a0-c4eb-dc076de14af6'
```

Response: 

```
Status: 400 Bad Request
{
    "error_description": "either access_token, id_token, or token_handle required"
}
```

EchterAgo:
I tested it, works fine for me. Sadly I managed to delete my RES settings while testing :\

AgrimPrasad:
Are there still any outstanding concerns here?

EchterAgo:
Not for me, tested with Chrome and Firefox, running fine for 2 days now.

larsjohnsen:
Thanks!

erikdesjardins:
Sorry, I should have mentioned earlier: adding new permissions in Firefox right now is a deal-breaker, since they'll cause a warning dialog, and we've not inclined to go through [that](https://www.reddit.com/r/Enhancement/comments/1jskcm/announcement_chrome_users_did_your_res_turn_off/) again

erikdesjardins:
Because that probably won't be implemented for a very long time (and other privacy concerns with users' emails\*), I'll instead add a text option where users can fill in the account to use. You'll still be credited in the changelog, etc. because you did a lot of work (partially for naught 😞).

\*: this'd have to be optional, disabled by default anyways because users will take offense to giving RES their email unnecessarily

larsjohnsen:
As for the text option, it would likely be possible to reuse some of this PR. `getActiveEmail` could prompt the user for a loginHint once (only if `loginHint.get()` returned undefined, as not to pester the user) instead of fetching it automatically.  A button to reset `loginHintStorage` could be added to the the module settings page.

Issue #1686, opened 2014-10-19
when visiting thread posted by alt account, switch to alt account


jewel-andraia:
spin-off of #1685 

https://www.reddit.com/r/Enhancement/comments/ws6hf/feature_request_option_to_switch_accounts_when/

optional, default off

PR #4443, opened 2017-10-05
Adds Auto Switch to accountswitcher 

### LINKED ISSUE #4442: Request to Implement: Automatically switch account if OP of post matches an added account
I'm planning to add this to the account switcher module. I may also expand the scope to look if any of the accounts in the account switcher have posted a comment on a post and switch to the account.

isiah-lloyd:
Adds options to switch to an alt account if you are on a comment page that was created by one of your alt accounts or contains a comment created by one of your alt accounts. On switching accounts a query parameter is added to the URL on page refresh so that a notification is displayed to alert the user that their account has been switched.

Relevant issue: Fixes #4442 and #1686
Tested in browser: Chrome

### DIFF
--- a/lib/modules/accountSwitcher.js
+++ b/lib/modules/accountSwitcher.js
@@ -3,7 +3,7 @@
 import _ from 'lodash';
 import { $ } from '../vendor';
 import { Module } from '../core/module';
-import { Alert, formatDate, formatDateDiff, getUserInfo, loggedInUser } from '../utils';
+import { Alert, formatDate, formatDateDiff, getUserInfo, loggedInUser, isPageType, watchForThings} from '../utils';
 import { ajax, multicast, i18n } from '../environment';
 import * as BetteReddit from './betteReddit';
 import * as CommandLine from './commandLine';
@@ -59,6 +59,20 @@ module.options = {
 		value: false,
 		advanced: true,
 	},
+	autoSwitchOnPost: {
+		type: 'boolean',
+		description: 'accountSwitcherAutoSwitchOnPostDesc',
+		title: 'accountSwitcherAutoSwitchOnPostTitle',
+		value: false,
+		advanced: false,
+	},
+	autoSwitchOnComment: {
+		type: 'boolean',
+		description: 'accountSwitcherAutoSwitchOnCommentDesc',
+		title: 'accountSwitcherAutoSwitchOnCommentTitle',
+		value: false,
+		advanced: false,
+	},
 	showCurrentUserName: {
 		type: 'boolean',
 		value: false,
@@ -106,10 +120,18 @@ module.options = {
 	},
 };
 
-let userLink, $downArrow, $downArrowOverlay;
+let userLink, $downArrow, $downArrowOverlay, queryParameter;
 
 export let $accountMenu;
 
+module.beforeLoad = () => {
+	if(isPageType('comments') && module.options.autoSwitchOnPost.value) {
+		watchForThings(['post'], autoSwitcher);
+	}
+	if(isPageType('comments') && module.options.autoSwitchOnComment.value) {
+		watchForThings(['comment'], autoSwitcher);
+	}
+}
 module.go = () => {
 	userLink = document.querySelector('#header-bottom-right > span.user > a');
 	if (userLink) {
@@ -434,12 +456,42 @@ function _notifySwitchedAccountElsewhere() {
 		message,
 	});
 }
-
+function autoSwitcher(thing) {
+	if(thing.getAuthor() != loggedInUser() && window.location.search.indexOf('accountSwitcher=auto') == -1 ) {
+		switchTo(thing.getAuthor());
+		queryParameter = "auto"
+	}
+	if(window.location.search.indexOf('accountSwitcher=auto') > -1 && thing.getAuthor() == loggedInUser() && thing.isPost()) {
+		Notifications.showNotification({
+			moduleID: module.moduleID,
+			notificationID: 'auto-switched',
+			closeDelay: 4000,
+			optionKey: 'autoSwitchOnPost',
+			header: i18n('accountSwitcherAutoSwitchNotificationTitle'),
+			message: i18n('accountSwitcherAutoSwitchPostDesc', loggedInUser()),
+		});
+	}
+	if(window.location.search.indexOf('accountSwitcher=auto') > -1 && thing.getAuthor() == loggedInUser() && thing.isComment()) {
+		Notifications.showNotification({
+			moduleID: module.moduleID,
+			notificationID: 'auto-switched',
+			closeDelay: 4000,
+			optionKey: 'autoSwitchOnComment',
+			header: i18n('accountSwitcherAutoSwitchNotificationTitle'),
+			message: i18n('accountSwitcherAutoSwitchCommentDesc', loggedInUser()),
+		});
+	}
+}
 function manageAccounts() {
 	SettingsNavigation.loadSettingsPage(module.moduleID, 'accounts');
 }
 
 function reloadPage() {
 	NeverEndingReddit.resetReturnToPage();
-	location.reload();
+	if(queryParameter) {
+		window.location.href = window.location.pathname+"?"+$.param({'accountSwitcher':queryParameter})
+	}
+	else {
+		location.reload();
+	}
 }
--- a/locales/locales/en.json
+++ b/locales/locales/en.json
@@ -1629,6 +1629,27 @@
 	"accountSwitcherReloadOtherTabsDesc": {
 		"message": "After switching accounts, automatically reload other tabs."
 	},
+	"accountSwitcherAutoSwitchOnPostTitle": {
+		"message": "Automatically Switch To Post OP"
+	},
+	"accountSwitcherAutoSwitchOnPostDesc": {
+		"message": "Automatically switch to alternative account if it's the OP of the post you are on."
+	},
+	"accountSwitcherAutoSwitchOnCommentTitle": {
+		"message": "Automatically Switch To Comment OP"
+	},
+	"accountSwitcherAutoSwitchOnCommentDesc": {
+		"message": "Automatically switch to alternative account if a comment from one of your alternative accounts is on the page"
+	},
+	"accountSwitcherAutoSwitchNotificationTitle": {
+		"message": "Account Switched"
+	},
+	"accountSwitcherAutoSwitchPostDesc": {
+		"message": "Your account has been switched to <b>$1</b> because it is the OP of this post"
+	},
+	"accountSwitcherAutoSwitchCommentDesc": {
+		"message": "Your account has been switched to <b>$1</b> because a comment by it was found on this post."
+	},
 	"accountSwitcherShowCurrentUserNameTitle": {
 		"message": "Show Current User Name"
 	},


lib/modules/accountSwitcher.js
```diff
@@ -434,12 +456,42 @@ function _notifySwitchedAccountElsewhere() {
 		message,
 	});
 }
-
+function autoSwitcher(thing) {
+	if(thing.getAuthor() != loggedInUser() && window.location.search.indexOf('accountSwitcher=auto') == -1 ) {
```
andytuba: DRY:

`const autoSwitched = window.location.search.indexOf('accountSwitcher=auto') === -1;`

then later  `if (autoSwitched && ...)` or `if (!autoSwitched && ...)`

andytuba: Use `!==` and `===` instead of `!=` and `!=`

andytuba: DRY: put this at the top of `function autoSwitcher`

`const currentUserNotAuthor === thing.getAuthor() !== loggedInUser();`  and then `if (currentUserNotAuthor && !autoSwitched)`

lib/modules/accountSwitcher.js
```diff
@@ -434,12 +456,42 @@ function _notifySwitchedAccountElsewhere() {
...
+		switchTo(thing.getAuthor());
+		queryParameter = "auto"
```
andytuba: Move setting `queryParameter` above `switchTo`.

lib/modules/accountSwitcher.js
andytuba: You should sanity check the type of thing and its relevant option at the very top of this function.

function autoSwitcher(thing) {
     if (thing.isPost()) { if (!module.options.autoSwitchOnPost.value) { return; } }
     if (thing.isComment()) { if (!module.options.autoSwitchOnComment.value) { return; } }

lib/modules/accountSwitcher.js
```diff
@@ -434,12 +456,42 @@ function _notifySwitchedAccountElsewhere() {
...
+}
 function manageAccounts() {
 	SettingsNavigation.loadSettingsPage(module.moduleID, 'accounts');
 }
 
 function reloadPage() {
 	NeverEndingReddit.resetReturnToPage();
-	location.reload();
+	if(queryParameter) {
+		window.location.href = window.location.pathname+"?"+$.param({'accountSwitcher':queryParameter})
```
andytuba: This approach is problematic for two reasons:

1. The `accountSwitcher?` could be copied into URLs which are shared and then launch a confusing pop-up to people who simply click the link.
1. This implementation, `${pathname}?accountSwitcher=foo`, doesn't preserve existing query parameters, such as context, depth, limit

A smoother approach would store "recently reloaded ${location.href} to switch accounts to ${username} " in RES storage, akin to [Comment Hide Persistor](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/456cc098577439b906f9f21e3554515b353a8572/lib/modules/commentHidePersistor.js#L21).

lib/modules/accountSwitcher.js
andytuba: Alternatively, `const autoSwitched = _.once(() =>  window.location.search.indexOf('accountSwitcher=auto') === -1;`  and call `if (autoSwitched() && ...`

lib/modules/accountSwitcher.js
```diff
@@ -106,10 +120,18 @@ module.options = {
 	},
 };
 
-let userLink, $downArrow, $downArrowOverlay;
+let userLink, $downArrow, $downArrowOverlay, queryParameter;
 
 export let $accountMenu;
 
+module.beforeLoad = () => {
+	if(isPageType('comments') && module.options.autoSwitchOnPost.value) {
```
andytuba: Instead of checking `if (isPageType('comments')...` you can simply add `module.includes = ['comments'];` in the "manifest" area above.

locales/locales/en.json
```diff
@@ -1629,6 +1629,27 @@
...
+		"message": "Automatically Switch To Post OP"
+	},
+	"accountSwitcherAutoSwitchOnPostDesc": {
+		"message": "Automatically switch to alternative account if it's the OP of the post you are on."
+	},
+	"accountSwitcherAutoSwitchOnCommentTitle": {
+		"message": "Automatically Switch To Comment OP"
+	},
+	"accountSwitcherAutoSwitchOnCommentDesc": {
+		"message": "Automatically switch to alternative account if a comment from one of your alternative accounts is on the page"
```
andytuba: How about ...

"Automatically switch to account in Account Switcher if that account has submitted a comment to the current post."

locales/locales/en.json
```diff
@@ -1629,6 +1629,27 @@
 	"accountSwitcherReloadOtherTabsDesc": {
 		"message": "After switching accounts, automatically reload other tabs."
 	},
+	"accountSwitcherAutoSwitchOnPostTitle": {
+		"message": "Automatically Switch To Post OP"
+	},
+	"accountSwitcherAutoSwitchOnPostDesc": {
+		"message": "Automatically switch to alternative account if it's the OP of the post you are on."
```
andytuba: How about ...

"Automatically switch to account in Account Switcher if that account submitted the current post."

locales/locales/en.json
```diff
@@ -1629,6 +1629,27 @@
...
+		"message": "Automatically Switch To Comment OP"
+	},
+	"accountSwitcherAutoSwitchOnCommentDesc": {
+		"message": "Automatically switch to alternative account if a comment from one of your alternative accounts is on the page"
+	},
+	"accountSwitcherAutoSwitchNotificationTitle": {
+		"message": "Account Switched"
+	},
+	"accountSwitcherAutoSwitchPostDesc": {
+		"message": "Your account has been switched to <b>$1</b> because it is the OP of this post"
```
andytuba: How about..

`"You have been automatically switched to <b>u/$1</b> because this account submitted this post."

locales/locales/en.json
```diff
@@ -1629,6 +1629,27 @@
...
+	},
+	"accountSwitcherAutoSwitchCommentDesc": {
+		"message": "Your account has been switched to <b>$1</b> because a comment by it was found on this post."
```
andytuba: How about..

`"You have been automatically switched to <b>u/$1</b> because this account submitted a comment on this post."

locales/locales/en.json
```diff
@@ -1629,6 +1629,27 @@
 	"accountSwitcherReloadOtherTabsDesc": {
 		"message": "After switching accounts, automatically reload other tabs."
 	},
+	"accountSwitcherAutoSwitchOnPostTitle": {
+		"message": "Automatically Switch To Post OP"
```
andytuba: How about `Author` instead of `OP`

locales/locales/en.json
```diff
@@ -1629,6 +1629,27 @@
...
+	},
+	"accountSwitcherAutoSwitchOnPostDesc": {
+		"message": "Automatically switch to alternative account if it's the OP of the post you are on."
+	},
+	"accountSwitcherAutoSwitchOnCommentTitle": {
+		"message": "Automatically Switch To Comment OP"
```
andytuba: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4443/files#r142847164

lib/modules/accountSwitcher.js
andytuba: It appears this patch won't let you deliberately switch to a different account. It would be great to check ["recently switched accounts on this post?"](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4443/files#r142846414) before even adding a hook to watchForThings.

lib/modules/accountSwitcher.js
isiah-lloyd: I very much agree with you, I felt icky writing that code but it was the only solution I could think of to retain data across the page refresh to login. I'm trying to learn the Storage module by reading the Comment Hide Persistor code but it's kind of confusing for someone who doesn't have much ES6 knowledge. Is there any documentation available? If this is changed this should fix some other changes you requested as well retaining to the Query Parameter.

larsjohnsen:
I don't think automatically switching account is desirable, as there's many cases where one could visit a thread without replying and don't want to have to log back into the previous account. Maybe just a notification `Noticed another account has created post / commented` with a button to switch to that account would be better?

isiah-lloyd:
@larsjohnsen I don't really have an opinion, just saw that the feature has been requested. Maybe have a toggle to automatically switch or to just notify? 

@andytuba What do you think?

lib/modules/accountSwitcher.js
```diff
@@ -106,10 +120,18 @@ module.options = {
...
+		watchForThings(['post'], autoSwitcher);
+	}
+	if(isPageType('comments') && module.options.autoSwitchOnComment.value) {
```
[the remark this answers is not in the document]
isiah-lloyd: If I understand it right `watchForThings(['post'], autoSwitcher);` would return posts even if we were on the frontpage or a subreddit listing. We don't want to switch a user if they just happened to load one of their posts on the frontpage. I agree that for the comments if statement it is superfluous

lib/modules/accountSwitcher.js
```diff
@@ -59,6 +59,20 @@ module.options = {
 		value: false,
 		advanced: true,
 	},
+	autoSwitchOnPost: {
+		type: 'boolean',
+		description: 'accountSwitcherAutoSwitchOnPostDesc',
+		title: 'accountSwitcherAutoSwitchOnPostTitle',
+		value: false,
+		advanced: false,
```
andytuba: :hocho:  `false` is default

andytuba:
> I don't think automatically switching account is desirable ... maybe just notification?

You could change the option from toggle to enum: do nothing, show notification prompting switching account, switch page.  That change could go into a follow-up PR, especially since this option is disabled by default.

lib/modules/accountSwitcher.js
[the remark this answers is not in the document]
andytuba: It's safer to double-check the page type, since users also encounter comments in user pages and /r/all/comments listings.

lib/modules/accountSwitcher.js
isiah-lloyd: Doing this causes each comment on a user page to be autoSwitched, eventually leading to you being logged out of reddit.

isiah-lloyd:
@andytuba Sorry for the long wait between this commit but I think it's now ready! I switched from using query parameters to saving the time of switch in RES storage and seeing if the account switched within the last minute to show the notification.

lib/modules/accountSwitcher.js
```diff
@@ -434,7 +455,38 @@ function _notifySwitchedAccountElsewhere() {
 		message,
 	});
 }
-
+async function autoSwitched() {
+	return Date.now() - await lastAutoSwitched.get() < 6000;
+}
+function autoSwitcher(thing) {
+	const currentUserIsAuthor = thing.getAuthor() === loggedInUser();
+	autoSwitched().then(isAutoSwitched => {
```
larsjohnsen: This seems like an expensive way to check, as `lastAutoSwitched` is fetched from Storage for every post.

isiah-lloyd:
@larsjohnsen I've updated it so it now only calls the function once before the module loads.

lib/modules/accountSwitcher.js
```diff
@@ -434,7 +463,36 @@ function _notifySwitchedAccountElsewhere() {
 		message,
 	});
 }
-
+async function autoSwitched() {
+	return Date.now() - await lastAutoSwitched.get() < 6000;
```
larsjohnsen: A more reliable way to notify about this may be to add a token to `sessionStorage`. Then, when loading a page, it can be checked whether the token exist and if it does, show the notification; if not, run the code which looks for another user.

The source of the switch can also be stored, i.e. `post` or `comment`, so that it won't be necessary to look for the author in the posts when creating the notification.

isiah-lloyd:
@larsjohnsen I switched to using `sessionStorage`

lib/modules/accountSwitcher.js
```diff
@@ -434,7 +459,37 @@ function _notifySwitchedAccountElsewhere() {
...
+	}
+	if (isAutoSwitched === 'post' && currentUserIsAuthor && !autoSwitchNotificationShown) {
+		Notifications.showNotification({
+			moduleID: module.moduleID,
+			notificationID: 'auto-switched',
+			optionKey: 'autoSwitchOnPost',
+			header: i18n('accountSwitcherAutoSwitchNotificationTitle'),
+			message: i18n('accountSwitcherAutoSwitchPostDesc', loggedInUsername),
+		});
+		autoSwitchNotificationShown = true;
```
larsjohnsen: The notifications ought to be separated out to a new `_.once` function. That way it can only be invoked once, and `autoSwitchNotificationShown` becomes redundant.

lib/modules/accountSwitcher.js
```diff
@@ -434,7 +459,37 @@ function _notifySwitchedAccountElsewhere() {
...
+		sessionStorage.removeItem('RESAccountSwitcherAutoSwitched');
+	}
+	if (isAutoSwitched === 'comment' && currentUserIsAuthor && !autoSwitchNotificationShown) {
+		Notifications.showNotification({
+			moduleID: module.moduleID,
+			notificationID: 'auto-switched',
+			optionKey: 'autoSwitchOnComment',
+			header: i18n('accountSwitcherAutoSwitchNotificationTitle'),
+			message: i18n('accountSwitcherAutoSwitchCommentDesc', loggedInUsername),
```
larsjohnsen: Since the only thing which differs from the above `post` notification is `message`, they can be combined: `message: i18n(isAutoSwitched === 'comment' 
 ? 'accountSwitcherAutoSwitchCommentDesc' : 'accountSwitcherAutoSwitchPostDesc', loggedInUsername),`

isiah-lloyd:
@larsjohnsen Implemented your suggestion for notifications.

isiah-lloyd:
Was just wondering if there is anything I need to do for the PR, or if we are just waiting to merge it?

lib/modules/accountSwitcher.js
```diff
@@ -434,7 +461,28 @@ function _notifySwitchedAccountElsewhere() {
 		message,
 	});
 }
-
+function autoSwitcher(thing) {
+	const loggedInUsername = loggedInUser();
+	if (!loggedInUsername) return;
+	const currentUserIsAuthor = thing.getAuthor() === loggedInUsername;
+	const autoSwitchNotif = _.once(() => {
```
larsjohnsen: Declaring this within the `autoSwitcher` scope makes `_.once` ineffectual, as it would not prevent `showNotification` from being called for `thing`s.

lib/modules/accountSwitcher.js
```diff
@@ -106,10 +120,21 @@ module.options = {
...
 };
 
-let userLink, $downArrow, $downArrowOverlay;
+let userLink, $downArrow, $downArrowOverlay, isAutoSwitched;
 
 export let $accountMenu;
 
+module.beforeLoad = () => {
+	if (module.options.autoSwitchOnPost.value || module.options.autoSwitchOnComment.value) {
+		isAutoSwitched = sessionStorage.getItem('RESAccountSwitcherAutoSwitched');
```
larsjohnsen: `showNotification` may as well be called directly from here, so that it won't be necessary to add the `autoSwitcher` watcher and thus avoids the `isAutoSwitched` conditionals.

lib/modules/accountSwitcher.js
```diff
@@ -305,13 +330,15 @@ export function closeAccountMenu() {
 
 function findMatchingAccount(username, partialMatch) {
 	const accounts = module.options.accounts.value;
-	const matched = accounts.find(([user]) => user.toUpperCase() === username.toUpperCase());
+	const constUsername = username;	// used to please flow tests
```
larsjohnsen: This change not necessary if checking that `thing.getAuthor()` is not null (like how `loggedInUser()` is checked).

isiah-lloyd: Interestingly, that's what I tried first but after some googling it seems that flow does not like checking directly against a function argument but requires that the variable is a `const`. So I cannot just do `if (!username) return;` like I do below.

larsjohnsen: That's a little weird. See if the Flow error changes by typing the username parameter (`function findMatchingAccount(username: string, partialMatch: boolean) {`.

PR #4446, opened 2017-10-06
Give hidden user names the same right margin as non-hidden ones

okdana:
Hey there. I noticed that if you tag yourself (which i started doing so i could find my comments more easily), and you have user-name hiding enabled, the normal margin between the name and the tag is missing:

screen shot 2017-10-05 at 22 15 47

Simply copying the `margin-right` value from reddit's own `a.author` rule seems to fix it.

Tested in browser: Firefox 56

### DIFF
--- a/lib/modules/usernameHider.js
+++ b/lib/modules/usernameHider.js
@@ -136,6 +136,7 @@ const hideUsername = _.memoize((user = loggedInUser()) => {
 			background-color: inherit;
 			border-radius: inherit;
 			padding: inherit;
+			margin-right: 0.5em;
 		}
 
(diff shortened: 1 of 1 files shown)


lib/modules/usernameHider.js
```diff
@@ -136,6 +136,7 @@ const hideUsername = _.memoize((user = loggedInUser()) => {
 			background-color: inherit;
 			border-radius: inherit;
 			padding: inherit;
+			margin-right: 0.5em;
```
erikdesjardins: Note: `inherit` doesn't work here because we set the font-size of the parent to 0 above

PR #4448, opened 2017-10-06
Update min supported version for edge

BenMcGarry:
Update for creators fall update. Release includes multiple browser fixes which was reported by the RES team.


erikdesjardins:
Skipping integration tests as we don't (yet) test Edge

Issue #3561, opened 2016-10-26
Don't add post expando title if it matches the post


andytuba:
https://www.reddit.com/r/AnimalsBeingJerks/comments/59d02v/got_the_family_together_for_a_christmas_photo/

![screen shot 2016-10-25 at 5 35 05 pm](https://cloud.githubusercontent.com/assets/455632/19709161/651bd566-9ad9-11e6-9cd7-6b3b15b21f2e.png)

erikdesjardins:
Even in that example, the imgur album has a period that the title lacks...not sure how often this would work

andytuba:
We could do a levenstein distance sort of thing. 95% match, don't show.

erikdesjardins:
Levenshtein distance would be super useful for filters (and [fast-levenshtein](https://www.npmjs.com/package/fast-levenshtein) is tiny and has zero dependencies), but I'm still not convinced of this use-case.

The small possibility of confusion ("why do some imgur albums not display titles?"), and the marginal possibility that it hides titles (especially short titles) that aren't actually the same doesn't seem worth the (also marginal) benefit of not seeing the same title twice.

vishva007:
Hi, 
Came here from hacktoberfest site. Can I look into this issue ?
Thanks

PR #4452, opened 2017-10-07
Don't add post expando title if it matches the post

nkoxlien:
Fixes issue #3561 using fast-levenshtein. Tested in Chrome.

### DIFF
--- a/lib/utils/Thing.js
+++ b/lib/utils/Thing.js
@@ -207,9 +207,9 @@ export class Thing {
 		}
 		const distance = levenshtein.get(value, title);
 		if (value.length > title.length) {
-			return (((value.length - distance) / value.length) >= 0.95);
+			return distance / value.length <= 0.05;
 		}
-		return (((title.length - distance) / title.length) >= 0.95);
+		return distance / title.length <= 0.05;
 	}
 
 	getTitle(): string {
--- a/lib/utils/Thing.js
+++ b/lib/utils/Thing.js
@@ -195,6 +195,19 @@ export class Thing {
 		return this.isComment() && !!this.element.parentElement && this.element.parentElement.classList.contains('nestedlisting');
 	}
 
+	isSimilarToTitle(value: string): boolean {
+		if (!value){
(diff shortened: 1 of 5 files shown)


lib/utils/Thing.js
```diff
@@ -195,6 +196,22 @@ export class Thing {
...
+		const title = this.getTitle();
+
+		if (!title) {
+			return false;
+		}
+		const distance = levenshtein.get(value, title);
+		if (value.length > title.length) {
+			return (((value.length - distance) / value.length) >= 0.95);
+		}
+		return (((title.length - distance) / title.length) >= 0.95);
```
mc10: I think this would be cleaner if you flipped the inequality; i.e. `return distance / title.length < 0.05`

nkoxlien: Yeah that's cleaner.  I still want it to be inclusive of the case where the string is only 5% different, so in my case it'll be `return distance / title.length <= 0.05;`.  I'll add another commit to change this.

lib/utils/Thing.js
```diff
@@ -195,6 +196,22 @@ export class Thing {
 		return this.isComment() && !!this.element.parentElement && this.element.parentElement.classList.contains('nestedlisting');
 	}
 
+	isSimilarToTitle(value: string): boolean {
```
erikdesjardins: I think this function would make more sense as a standalone `export function isSimilar(a: string b: string)` (or `areSimilar`?) in lib/utils/string.js. That file is exported from /utils as a namespace, so you can call `string.isSimilar(thing.getTitle(), mediaOptions.title)`. `Thing` is already a pretty big class so I want to keep it focused on things (heh) that are strictly related to `thing`s.

nkoxlien: Yeah I wasn't really sure what the best place for it would be, but that seems a lot more fitting.  I'll add another commit moving it to string.js .

erikdesjardins:
Sorry about the delay; thanks!

Issue #1680, opened 2014-10-16
add no-subreddit-style class to body on pages without srstyle


jewel-andraia:
both within subreddits when subreddit style disabled (by RES or via reddit preference) and on non-subreddit pages

https://www.reddit.com/r/Enhancement/comments/2jd81r/feature_request_could_we_have_a_module_for/clapchf

re: #1255

PR #4460, opened 2017-10-11
add no-subreddit-style class to body on pages without srstyle

nkoxlien:
Relevant issue: fixes issue #1680
Tested in browser: Chrome

### DIFF
--- a/lib/modules/styleTweaks.js
+++ b/lib/modules/styleTweaks.js
@@ -330,6 +330,7 @@ async function toggleSubredditStyle(toggle, subreddit = currentSubreddit()) {
 
 		BodyClasses.toggle(toggle, 'res-srstyle-enabled');
 		BodyClasses.toggle(!toggle, 'res-srstyle-disabled');
+		BodyClasses.toggle(!toggle, 'no-subreddit-style');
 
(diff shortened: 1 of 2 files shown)


lib/modules/styleTweaks.js
```diff
@@ -330,6 +331,7 @@ async function toggleSubredditStyle(toggle, subreddit = currentSubreddit()) {
 
 		BodyClasses.toggle(toggle, 'res-srstyle-enabled');
 		BodyClasses.toggle(!toggle, 'res-srstyle-disabled');
+		BodyClasses.toggle(!toggle, 'no-subreddit-style');
```
erikdesjardins: It looks like `res-srstyle-disabled` is already used for this, so you can remove this line and change the other line to add `res-srstyle-disabled`

nkoxlien: For some reason I assumed the issue was asking for an additional class on top of `res-srstyle-disabled` (which wouldn't really make sense looking back on it).  I'll change it to just use the existing "no subreddit style" class.

erikdesjardins:
Thanks!

PR #4462, opened 2017-10-12
Don't add candystripe inside spoiler selfpost text

### LINKED ISSUE #4016: Expando button inside spoiler selfpost text incorrectly adds candystripe
The selector targeting "post's expando button" just needs to be a little stricter.

nkoxlien:
Made the selector stricter so that only expandos that are siblings of NSFW/spoiler post title divs will get candystripe.

Relevant issue: fixes #4016 
Tested in browser: Chrome

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -18,10 +18,7 @@ img {
 	pointer-events: none;
 }
 
-.expando-button {
-	cursor: pointer;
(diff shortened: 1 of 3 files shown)


nkoxlien:
I accidentally forked this branch from another issue branch instead of the master and didn't realize until after I made the pull request :/  I'll make another pull request from a branch with the correct changes.

PR #4463, opened 2017-10-12
Don't add candystripe inside spoiler selfpost text

### LINKED ISSUE #4016: Expando button inside spoiler selfpost text incorrectly adds candystripe

nkoxlien:
Made the selector stricter so that only expandos that are siblings of NSFW/spoiler post title divs will get candystripe.

Relevant issue: fixes #4016
Tested in browser: Chrome

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -18,10 +18,7 @@ img {
 	pointer-events: none;
 }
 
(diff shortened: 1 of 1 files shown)


erikdesjardins:
Thanks!

(CI failure is spurious)

PR #4466, opened 2017-10-13
Open xpost link in new tab

BenMcGarry:
Relevant issue: https://www.reddit.com/r/RESissues/comments/766nn3/clicking_on_one_of_the_subreddits_in_xposted_from/
Tested in browser: Chrome

### DIFF
--- a/lib/modules/xPostLinks.js
+++ b/lib/modules/xPostLinks.js
@@ -51,6 +51,7 @@ function appendToTagline(sub, thing) {
 			$('<a>', {
 				class: 'subreddit hover',
 				href: `/r/${sub}`,
+				target: '_blank',
 				text: `/r/${sub}`,
 			})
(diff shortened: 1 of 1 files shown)


lib/modules/xPostLinks.js
```diff
@@ -51,6 +51,7 @@ function appendToTagline(sub, thing) {
 			$('<a>', {
 				class: 'subreddit hover',
 				href: `/r/${sub}`,
+				target: '_blank',
```
larsjohnsen: Should match the preference, so `thing.getTitleElement().target`.

erikdesjardins:
Reddit does this dynamically on click (😞), so we can't easily detect what the user's preference is. Since the default for logged in users is to not open in a new tab, I think we'll leave it as-is for now.

PR #4469, opened 2017-10-15
added custom lint rule for property name spaces

### LINKED ISSUE #4461: Custom ESLint rule to forbid spaces in module option names
i.e. an object assigned to the options property of anything or which is the options property of an object literal must have keys consisting only of word characters.

bad:
```js
module.options = {
  'some option': {}
}

export default new Host('foo', {
  options: {
    'some option': {}
  }
})
```

good:
```js
module.options = {
  someOption: {}
}

export default new Host('foo', {
  options: {
    someOption: {}
  }
})
```

kraigh:
Relevant issue: fixes #4461 
Tested in browser: chrome

I added a `no-restricted-syntax` value with an AST selector that looks for object property names with spaces in them.

Running the lint there are ~60 errors thrown. Some of them seem reasonable, like test names.

Is there a good way to restrict this rule to only directories where we care about not having spaces in property names?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -125,7 +125,7 @@
 		"no-restricted-globals": [2, "alert"],
 		"no-restricted-imports": [2, "jquery"], // import it from vendor/ instead
 		"no-restricted-modules": 0, // see import/no-nodejs-modules
-		"no-restricted-syntax": [2, "ForStatement"], // use for..of or Array.prototype methods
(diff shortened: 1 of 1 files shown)


erikdesjardins:
Thanks; I completely forgot about `no-restricted-syntax`, that's much simpler than writing a custom plugin!

erikdesjardins:
(integration tests skipped as no executable code changed; everything else passed)

PR #4470, opened 2017-10-15
Don't notify No Participation when reverting vote

mgiuffrida:
Manually reverting a vote shouldn't trigger the No Participation notification.


erikdesjardins:
Thanks!

PR #4473, opened 2017-10-15
Adding Comment Navigator feature - "Auto-IAmA"

### LINKED ISSUE #1442: Automatically show "navigate by: IAmA" in /r/IAmA comment pages
because HOW TOTALLY AWESOME WOULD THAT BE

(don't forget an option)

Antrikshy:
Relevant issue: Adds feature and resolves #1442
Tested in browser: Firefox Developer Edition (Aurora, 57.0b8)

This change adds a setting, Auto-IAmA (`true` by default), which is dependent on Comment Navigator's `showByDefault`. When both are enabled, Comment Navigator appears automatically set to the IAmA setting when the user is browsing comments on r/IAmA or r/casualiama.

Notes:
- I had some issues running Flow check. I'm hoping someone can help if the Travis CI build fails.
- I hope my method of preventing auto-scroll-to-first-comment on page load is not too janky.
- Can someone confirm if I should just copy my new user facing strings as-is to the other localization files?

### DIFF
--- a/lib/modules/commentNavigator.js
+++ b/lib/modules/commentNavigator.js
@@ -219,7 +219,7 @@ module.go = () => {
 	}
 };
 
-let currentCategory = isCurrentSubreddit('IAmA') ? 'IAmA' : '';
+let currentCategory = isCurrentSubreddit('IAmA', 'casualiama') ? 'IAmA' : '';
 
 function onScroll() {
 	const category = currentCategory;
@@ -302,8 +302,11 @@ export function toggleNavigator() {
 }
 
 export function refreshCurrentCategoryPosts() {
-	_posts[currentCategory] = sortTypes[currentCategory] ? sortTypes[currentCategory].getPosts() : sortTypes[currentCategory];
-	resetNavigator(currentCategory);
+	if (currentCategory) {
+		if (!_posts[currentCategory] || currentCategory === 'highlighted') {
+			_posts[currentCategory] = sortTypes[currentCategory].getPosts();
+		}
+	}
 }
 
 function resetNavigator(category) {
--- a/lib/modules/commentNavigator.js
+++ b/lib/modules/commentNavigator.js
@@ -13,6 +13,7 @@ import {
 	scrollToElement,
 	string,
 	waitForEvent,
+	isCurrentSubreddit,
 } from '../utils';
 import * as Floater from './floater';
 import * as UserInfo from './userInfo';
@@ -28,6 +29,13 @@ module.options = {
 		value: false,
 		description: 'commentNavigatorShowByDefaultDesc',
 		title: 'commentNavigatorShowByDefaultTitle',
+	},
+	autoNavigateByIama: {
+		type: 'boolean',
+		value: true,
+		description: 'commentNavigatorAutoNavigateByIamaUserDesc',
+		title: 'commentNavigatorAutoNavigateByIamaUserTitle',
+		dependsOn: options => options.showByDefault.value,
 	},
 	openOnHighlightUser: {
 		type: 'boolean',
@@ -214,7 +222,11 @@ module.go = () => {
 	});
 
 	if (module.options.showByDefault.value) {
-		commentNav();
+		if (module.options.autoNavigateByIama.value && isCurrentSubreddit('IAmA')) {
+			showNavigator('IAmA', true);
+		} else {
+			commentNav();
+		}
 	}
 };
 
@@ -236,14 +248,14 @@ function onScroll() {
 	}
 }
 
-function changeCategory() {
+function changeCategory(suppressAutoScroll: boolean = false) {
 	const index = commentNav().select.selectedIndex;
 	if (index === -1) {
 		return;
 	}
 	currentCategory = (commentNav().select.options.item(index).value: any);
 	if (currentCategory !== '') {
-		getPostsByCategory();
+		getPostsByCategory(suppressAutoScroll);
 		commentNav().buttons.style.display = 'block';
 	} else {
 		commentNav().buttons.style.display = 'none';
@@ -253,7 +265,7 @@ function changeCategory() {
 
 let isOpen = false;
 
-export function showNavigator(category: $Keys<typeof sortTypes> | '' = currentCategory) {
+export function showNavigator(category: $Keys<typeof sortTypes> | '' = currentCategory, suppressAutoScroll: boolean = false) {
 	isOpen = true;
 	commentNav().box.style.display = 'block';
 
@@ -265,7 +277,7 @@ export function showNavigator(category: $Keys<typeof sortTypes> | '' = currentCa
 		commentNav().select.selectedIndex = parseInt(category, 10);
 	}
 	$('#commentNavBy').focus();
-	changeCategory();
+	changeCategory(suppressAutoScroll);
 }
 
 export function hideNavigator() {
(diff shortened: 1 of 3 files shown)


larsjohnsen:
Might be easier to set the value of the [select](https://github.com/Antrikshy/Reddit-Enhancement-Suite/blob/a39eac3b9c5395d594165baf0dfe5ec1fb1b28e0/lib/modules/commentNavigator.js#L61) to `IAmA` when the subreddit `isCurrentSubreddit('IAmA')`. That way you don't need to propagating `suppressAutoScroll`.

The option is probably unnecessary as setting this as default don't really change the behavior of the select.

Antrikshy:
I tried replacing the call to `showNavigator` with `commentNav().select.value = 'IAmA';`. While this updates the value of the select, it does not trigger anything along with it, and the navigator arrows don't appear. I added a call to `changeCategory` right after it, which results in the same behavior as my changes but without the `suppressAutoScroll` variable. 

The reason I pass the suppressor around is that without it, the page otherwise auto-scrolls to the first comment (as determined by the navigator). This is expected to happen after the user manually selects "IAmA", but it's jarring when it happens on page load, forcing the user to scroll up to see the post title and body this way.

With the suppression argument, the auto scrolling is not triggered, giving the user time to scroll down to the comments, and the navigator is then ready for them to use.

larsjohnsen:
You're right; it's equivalent.

The `resetNavigator` call which is suppressed does however seem misplaced. I think it would be clearer if it's in `changeCategory` and then only called if the `changeCategory` changes `currentCategory` to a new value. Since `showNavigator` then already have set the select before calling  `changeCategory`, the scroll will implicitly be suppressed.

Antrikshy:
Hey, perfect! That also resolves the various failing tests that I was not looking forward to fixing. Still think the option isn't required? Not that I can think of an edge case when someone would want to turn it off. I added it initially because I didn't know it would be so easy to roll it up into the "show by default" option.

Antrikshy:
I see the dumb thing I did that's causing that integration test to fail. Gotta move the call to `resetNavigator`.

lib/modules/commentNavigator.js
```diff
@@ -241,13 +253,17 @@ function changeCategory() {
 	if (index === -1) {
 		return;
 	}
-	currentCategory = (commentNav().select.options.item(index).value: any);
-	if (currentCategory !== '') {
+	const newCategory = (commentNav().select.options.item(index).value: any);
+	if (newCategory !== currentCategory) {
+		resetNavigator(newCategory);
```
larsjohnsen: `getPostsByCategory` must be called before `resetNavigator` (so `currentCategory must be set before then).

lib/modules/commentNavigator.js
```diff
@@ -29,6 +30,13 @@ module.options = {
 		description: 'commentNavigatorShowByDefaultDesc',
 		title: 'commentNavigatorShowByDefaultTitle',
 	},
+	autoNavigateByIama: {
```
larsjohnsen: As noted, this option is not really needed.

lib/modules/commentNavigator.js
```diff
@@ -214,7 +222,11 @@ module.go = () => {
 	});
 
 	if (module.options.showByDefault.value) {
-		commentNav();
+		if (module.options.autoNavigateByIama.value && isCurrentSubreddit('IAmA')) {
+			showNavigator('IAmA');
```
larsjohnsen: In order for `IAmA` to be set as initial also when  `showByDefault` enabled, the initial value of `currentCategory` can be set to `isCurrentSubreddit('IAmA') ? 'IAmA' : '';`.

lib/modules/commentNavigator.js
```diff
@@ -241,13 +253,17 @@ function changeCategory() {
...
 		return;
 	}
-	currentCategory = (commentNav().select.options.item(index).value: any);
-	if (currentCategory !== '') {
+	const newCategory = (commentNav().select.options.item(index).value: any);
+	if (newCategory !== currentCategory) {
+		resetNavigator(newCategory);
+	}
+	if (newCategory !== '') {
 		getPostsByCategory();
```
larsjohnsen: Is must also be called when there's no category selected, in order to avoid too much logic.

lib/modules/commentNavigator.js
```diff
@@ -286,7 +302,6 @@ export function getPostsByCategory() {
 		if (!_posts[currentCategory] || currentCategory === 'highlighted') {
```
larsjohnsen: This this function now is only called when the selection is changed, it can be simplified to `_posts[currentCategory] = sortTypes[currentCategory] ? sortTypes[currentCategory].getPosts() : sortTypes[currentCategory];` and renamed to something more fitting like `refreshCurrentCategoryPosts`.

Antrikshy: I don't get why we'd assign `_posts[currentCategory] = ... : sortTypes[currentCategory]`.

lib/modules/commentNavigator.js
Antrikshy: Will do, but does that matter? Doesn't the user need to click on one of the "navigate by" buttons to show the navigator if it isn't shown by default? I don't use the feature much, so I might be missing something.

lib/modules/commentNavigator.js
Antrikshy: I think I found it like this, but ok!

lib/modules/commentNavigator.js
```diff
@@ -248,22 +237,37 @@ function onScroll() {
...
+	currentCategory = (commentNav().select.options.item(index).value: any);
+	refreshCurrentCategoryPosts();
+	if (previousCategory !== currentCategory) {
+		resetNavigator(currentCategory);
 	}
-	if (newCategory !== '') {
-		getPostsByCategory();
+	if (currentCategory !== '') {
 		commentNav().buttons.style.display = 'block';
+		enableControls();
```
Antrikshy: Some change along the way prevented the navigation buttons from activating when IAmA mode was activated quietly.

lib/modules/commentNavigator.js
```diff
@@ -297,26 +301,19 @@ export function toggleNavigator() {
...
 }
 
-export function getPostsByCategory() {
-	if (currentCategory) {
-		if (!_posts[currentCategory] || currentCategory === 'highlighted') {
-			_posts[currentCategory] = sortTypes[currentCategory].getPosts();
-		}
-	}
+export function refreshCurrentCategoryPosts() {
+	_posts[currentCategory] = sortTypes[currentCategory] ? sortTypes[currentCategory].getPosts() : sortTypes[currentCategory];
```
Antrikshy: Oh, I forgot that this change you suggested causes the flow check to fail with really cryptic messages.

lib/modules/commentNavigator.js
larsjohnsen: It can also be opened by keyboardNav (shortcut `n`).

lib/modules/commentNavigator.js
larsjohnsen: Sorry, that was a typo. Anyway, now that the scroll is suppressed when opening the commentNav by default, we may as well remove the empty select option as that way remove a lot of logic related to that.

Antrikshy: Can you correct the typo? Did you mean to set it to a reasonable default like `_posts[currentCategory] = sortTypes[currentCategory] ? sortTypes[currentCategory].getPosts() : {};`?

Also, something causes the scroll to only be suppressed when "IAmA" is pre-set. I removed the empty option and navigated to a non-AMA page. "submitter" was selected, and the page did jump. I could dig deeper and clean it up, but could we do that as a separate PR? It's quite a bit of refactoring, and I'd have to understand the module better to really test it. 

Also, now that I think about it, what would you like to default the option to, if I were to remove the empty option?

Antrikshy:
I reverted the get-posts logic back, re-added a conditional that I removed before that caused Flow checks to fail.

I also enabled IAmA mode on r/casualiama, as #1442 said. I forgot to do this before.

lib/modules/commentNavigator.js
larsjohnsen: I was thinking `_posts[currentCategory] = sortTypes[currentCategory] ? sortTypes[currentCategory].getPosts() : [];` as `sortTypes[currentCategory]` is not defined for `''`, and `getPosts()` return `Arrray<HTMLElement>`. Anyway, the last commit appears to have made everything work, so my suggestion can be disregarded :)

larsjohnsen:
Thanks!

PR #4478, opened 2017-10-16
Show all users if there are no tagged users in UserTagger

### LINKED ISSUE #2081: UserTagger dashboard tab should show all users if no tagged users
https://www.reddit.com/r/Enhancement/comments/2ydqh4/can_i_see_which_users_i_upvotedownvote_the_most/cp8kvw4?context=3

pll33:
Relevant issue: resolves #2081
Tested in browser: Firefox, Chrome

Basically maps `tagData` again without the filter for 'tagged users' and updates the `#tagFilter` dropdown to 'all users'

### DIFF
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -10,6 +10,7 @@ import {
 	batch,
 	click,
 	downcast,
+	filterMap,
 	isCurrentSubreddit,
 	isPageType,
 	loggedInUser,
@@ -646,11 +647,10 @@ function getVoteWeightStyle({ votes, votesUp, votesDown }) {
 
 async function buildTags() {
 	const tagsData = await tagStorage.getAll();
-	return Object.entries(tagsData).map(([k, v]) => {
+	return filterMap(Object.entries(tagsData), ([k, v]) => {
 		const tag = Tag.getUnfilled(k);
-		tag.text = tag.text === null ? '' : tag.text;
 		tag.load(v);
-		return tag;
+		return [tag];
 	});
 }
 
@@ -761,7 +761,7 @@ async function drawUserTagTable(sortMethod, descending) {
 	let tags = await buildTags();
 
 	if (filterType === i18n('userTaggerTaggedUsers')) {
-		tags = tags.filter(tag => !!tag.text);
+		tags = tags.filter(tag => tag.text !== null);
 	}
 
 	switch (currentSortMethod) {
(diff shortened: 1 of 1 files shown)


lib/modules/userTagger.js
```diff
@@ -740,15 +740,25 @@ let currentSortMethod, isDescending;
...
 	currentSortMethod = sortMethod || currentSortMethod;
 	isDescending = (descending === undefined || descending === null) ? isDescending : descending;
-	const filterType = $('#tagFilter').val();
 	const tagsData = await tagStorage.getAll();
-	const tags = filterMap(Object.entries(tagsData), ([k, v]) => {
+	const filterType = $('#tagFilter').val();
+	let tags = filterMap(Object.entries(tagsData), ([k, v]) => {
 		const tag = Tag.getUnfilled(k);
 		tag.load(v);
 		if (filterType === 'tagged users' && !tag.text) return;
```
larsjohnsen: I didn't notice this earlier, but `filterType === 'tagged users'` won't work when the string is translated as the options in the select does not have the `value` attribute defined.

lib/modules/userTagger.js
```diff
@@ -740,15 +740,25 @@ let currentSortMethod, isDescending;
...
 		return [tag];
 	});
 
+	// show 'all users' if there are no 'tagged users'
+	if (filterType === 'tagged users' && tags.length === 0) {
+		$('#tagFilter').val(i18n('userTaggerAllUsers'));
```
larsjohnsen: In order to avoid loading all tags twice, a cleaner solution may be to build a complete `tags`, then creating the `#tagFilter` select with the default value depending on whether there's any tags
`<select id="tagFilter" value="${tags.find(v => v.text) ? 'tagged' : 'all'}">`, and don't do any filtering before then.

pll33: When should `tags` be built? It currently looks like it's being rebuilt every single time the table needs to be drawn. So that's kind of excessive? 

When the `#tagFilter` select is created initially, it's not aware of the tags since those haven't been built yet. So should I build the tags once only when the dashboard functionality is being added (in `function addDashboardFunctionality` instead of `function drawUserTagTable`) and then load those same tags every time the table needs to be redrawn? (I'm guessing by passing them to `drawUserTagTable` as function args?)

Additionally should the tagged users option never be shown if there are no tagged users, or should it only show all users at first, initially by default, when there are no tagged users? The latter of which the user can select 'tagged users' but it will still only show an empty table.

larsjohnsen: Passing it as a function arg sounds good, but since `tagStorage` may change, it may be easier to split `tagStorage` → `Array<Tag>` into a new function and call it from both places.

I think both options can stay regardless of there being any matching tags, as not to make this too complex.

PR #4479, opened 2017-10-17
Fix watching of comments without replies

larsjohnsen:
Replies to such comments wasn't processed.

Fixes regression from #3644.

Tested in browser: Chrome 63

### DIFF
--- a/lib/utils/watchers.js
+++ b/lib/utils/watchers.js
@@ -110,21 +110,13 @@ export function newSitetable(siteTable: HTMLElement) {
 }
 
 function addCommentsObserver(ele) {
-	const observer = observe(ele, { childList: true }, mutation => {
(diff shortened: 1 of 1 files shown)


lib/utils/watchers.js
```diff
@@ -109,23 +110,38 @@ export function newSitetable(siteTable: HTMLElement) {
...
+				resolve();
+
+				for (const e of addedNode.querySelectorAll('.thing')) {
+					registerElement('newComments', e);
+				}
+			});
+		});
+	}
+
+	/*:: if (!sitetable) throw new Error(); */
```
erikdesjardins: You might be able to convince Flow if you resolve the promise with addedNode and assign it outside

```js
if (!sitetable) {
		sitetable = await new Promise(resolve => {
```

PR #4480, opened 2017-10-17
UserTagger preset tags

### LINKED ISSUE #1263: easier user tagger color selcetion
Right now, for flair you have to type something in, the use the dropdown menu to select a colour.

Could we add an option so that instead of a dropdown menu, there was a pallatte menu. With a tiny box for each colour and one for no colour. And if the person didn't bother to fill in text, it would automatically add a few &nbsp;s or something so that the ~~flair~~ usertag colour was visible.

That way you could easily colour tag someone with three keystrokes. Click the ~~flair~~ user tagger button, click the colour, click save.

Also, maybe add an option so that if you set colour and no text, it doesn't save the link. Or better yet, it saves the link but automatically replaces it if you write in a proper user tag for them later.

### LINKED ISSUE #1270: User Tagger editor - macro-style dropdown with text/color to autofill usertag
https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/1263#issuecomment-46223953

pll33:
Relevant issue: resolves #1270 closes #1263
Tested in browser: Firefox

- [x] Select option field in core module
  - [x] Implement field functionality in settingsConsole
- [x] Ability to add/remove presets in userTagger options (table with text field and dropdown color field)
- [x] ~~Option to show/hide presets in dialog~~ (removed)
- [x] "Save as presets" button in tag dialog
  - [x] Bump button over if the preview tag text is too long (added flex-wrap)
- [x] Show presets in tag dialog 
  - [x] When preset tag is clicked in the dialog, it auto-saves that color/text option as the new tag and then closes the dialog
- [x] New strings for userTagger options in locales/en.json
- [x] Fix Flow type checking errors
- [ ] Test on other browsers: Chrome, Edge

Screenshots -

UserTagger options:
screen shot 2017-10-16 at 3 27 32 pm

A few presets:
screen shot 2017-10-17 at 1 51 27 pm

UserTagger dialog:
screen shot 2017-10-17 at 6 24 55 am

~~Is calling it `preset` fine, or should it be called something else?~~ 

Changed it to `presetTags` to avoid any confusion with the existing `presets` module

### DIFF
--- a/lib/modules/settingsConsole.js
+++ b/lib/modules/settingsConsole.js
@@ -795,8 +795,10 @@ function stageCurrentModuleOptions() {
 					const tempArray = input.value.split(',');
 					// convert the internal values of this array into their respective types (int, bool, bool, bool)
 					optionValue = [parseInt(tempArray[0], 10), (tempArray[1] === 'true'), (tempArray[2] === 'true'), (tempArray[3] === 'true'), (tempArray[4] === 'true')];
-				} else if (input.getAttribute('class') && input.getAttribute('class').includes('select') && input.selectedOptions) {
-					optionValue = input.selectedOptions[0].value;
+				} else if (input instanceof HTMLSelectElement) {
+					if (input.selectedIndex >= 0) {
+						optionValue = input.options[input.selectedIndex].value;
+					}
 				} else {
 					optionValue = input.value;
 				}
@@ -835,8 +837,10 @@ function stageCurrentModuleOptions() {
 							const tempArray = input.value.split(',');
 							// convert the internal values of this array into their respective types (int, bool, bool, bool)
 							optionValue = [parseInt(tempArray[0], 10), (tempArray[1] === 'true'), (tempArray[2] === 'true'), (tempArray[3] === 'true')];
-						} else if (input.getAttribute('class') && input.getAttribute('class').includes('select') && input.selectedOptions) {
-							optionValue = input.selectedOptions[0].value;
+						} else if (input instanceof HTMLSelectElement) {
+							if (input.selectedIndex >= 0) {
+								optionValue = input.options[input.selectedIndex].value;
+							}
 						} else {
 							optionValue = input.value;
 						}
--- a/lib/modules/settingsConsole.js
+++ b/lib/modules/settingsConsole.js
@@ -449,7 +449,6 @@ function drawOptionInput(mod, optionName, optionObject, isTable) {
 				}
 				$thisOptionFormEle.append($thisOptionFormSubEle);
 			});
-
 			break;
 		default:
 			throw new Error(`modules.${mod.moduleID}.options.${optionName} has invalid type: ${optionObject.type}`);
--- a/lib/modules/settingsConsole.js
+++ b/lib/modules/settingsConsole.js
@@ -795,7 +795,7 @@ function stageCurrentModuleOptions() {
 					const tempArray = input.value.split(',');
 					// convert the internal values of this array into their respective types (int, bool, bool, bool)
 					optionValue = [parseInt(tempArray[0], 10), (tempArray[1] === 'true'), (tempArray[2] === 'true'), (tempArray[3] === 'true'), (tempArray[4] === 'true')];
-				} else if (input.getAttribute('class') && input.getAttribute('class').includes('select')) {
+				} else if (input.getAttribute('class') && input.getAttribute('class').includes('select') && input.selectedOptions) {
 					optionValue = input.selectedOptions[0].value;
 				} else {
 					optionValue = input.value;
@@ -835,7 +835,7 @@ function stageCurrentModuleOptions() {
 							const tempArray = input.value.split(',');
 							// convert the internal values of this array into their respective types (int, bool, bool, bool)
 							optionValue = [parseInt(tempArray[0], 10), (tempArray[1] === 'true'), (tempArray[2] === 'true'), (tempArray[3] === 'true')];
-						} else if (input.getAttribute('class') && input.getAttribute('class').includes('select')) {
+						} else if (input.getAttribute('class') && input.getAttribute('class').includes('select') && input.selectedOptions) {
 							optionValue = input.selectedOptions[0].value;
 						} else {
 							optionValue = input.value;
--- a/lib/modules/settingsConsole.js
+++ b/lib/modules/settingsConsole.js
@@ -447,7 +447,6 @@ function drawOptionInput(mod, optionName, optionObject, isTable) {
 				if ((optionObject.value === optionValue.value) || nullEqualsEmpty) {
 					$thisOptionFormSubEle.attr('selected', 'selected');
 				}
-				// console.log($thisOptionFormSubEle)
 				$thisOptionFormEle.append($thisOptionFormSubEle);
 			});
 
--- a/lib/modules/settingsConsole.js
+++ b/lib/modules/settingsConsole.js
@@ -426,6 +426,32 @@ function drawOptionInput(mod, optionName, optionObject, isTable) {
 			}).val(niceKeyCode(optionObject.value));
 			$thisOptionFormEle = $('<div>').append(realOptionFormEle).append(thisKeyCodeDisplay);
 			break;
+		case 'select':
+			$thisOptionFormEle = $('<select>').attr({
+				id: optionName,
+				class: 'select',
+				value: optionObject.value,
+			});
+
+			optionObject.values.forEach((optionValue, index) => {
+				const thisId = `${optionName}-${index}`;
+				const $thisOptionFormSubEle = $('<option />', {
+					id: thisId,
+					class: 'select-option',
+					value: optionValue.value,
+					moduleID: mod.moduleID,
+					style: optionValue.style,
+				}).text(optionValue.name);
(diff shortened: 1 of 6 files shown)


lib/modules/userTagger.js
```diff
@@ -337,7 +396,7 @@ function populateDialog(tag: Tag, card) {
 				<label class="fieldPair-label" for="userTaggerColor">Color</label>
 				<select id="userTaggerColor" value="${tag.color}">
 					${colors.map(({ textColor, color }) => string._html`
-						<option style="color: ${textColor}; background-color: ${color}" value="${color}">${color}</option>
+						<option style="color: ${textColor}; background-color: ${color}" value="${color}" ${tag.color === color && string._html`selected`}>${color}</option>
```
larsjohnsen: Seems like this fixes a bug. Could you create a new PR for it?

lib/modules/userTagger.js
```diff
@@ -80,6 +104,36 @@ module.options = {
 		advanced: true,
 		dependsOn: options => options.trackVoteWeight.value,
 	},
+	showPresetTags: {
```
larsjohnsen: If all this does is determining whether the "presets" row is visible, it likely is not necessary. Just not generating the row should suffice.

lib/modules/userTagger.js
```diff
@@ -337,7 +393,7 @@ function populateDialog(tag: Tag, card) {
 				<label class="fieldPair-label" for="userTaggerColor">Color</label>
 				<select id="userTaggerColor" value="${tag.color}">
 					${colors.map(({ textColor, color }) => string._html`
-						<option style="color: ${textColor}; background-color: ${color}" value="${color}">${color}</option>
+						<option style="color: ${textColor}; background-color: ${color}" value="${color}" ${tag.color === color && string._html`selected`}>${color}</option>
 					`)}
 				</select>
 			</div>
```
larsjohnsen: It would be neat to add a presets directly this pop-up. Perhaps a "save preset" button add be located next to the preview?

lib/modules/userTagger.js
```diff
@@ -417,6 +490,19 @@ function populateDialog(tag: Tag, card) {
 	}
 
 	elements.openLink.addEventListener('click', () => openNewTabs('none', ...elements.link.value.split(/\s/)));
+	elements.presetTag.addEventListener('click', (e: MouseEvent) => {
+		const preset = {
```
larsjohnsen: A simpler solution would be to add a listener when generating the preset option. Perhaps something like
```
elements.presetTag.append(...presetTags.map(([text, color]) => {
  const element = Tag.buildTagElement({ text, color });
  element.addEventListener('click', () => {
    tag.load({ text, color });
    tag.save();
    card.close();
  });
})
```

lib/modules/userTagger.js
pll33: Would "save as preset" also save that tag to the current selected user, or would it only append that to the presets in the dialog and the presets table in the settings console?

larsjohnsen: I think the latter sounds most intuitive.

lib/modules/userTagger.js
pll33: Is there a way to conditionally include the presets row in the dialog? I tried something like this within the `body` assignment but it wouldn't load the dialog even though the HTML immediately after assignment was completely fine. 

```
${presetTags.length > 0 && string._html`
    <div class="fieldPair">
        <label class="fieldPair-label" for="userTaggerPresetTags">Presets</label>
        <span id="userTaggerPresetTags"></span>
    </div>`
}
```

pll33:
Added inline CSS for the preview row to flex-wrap so the new button gets pushed down to the next line:

Without flex-wrap:
screen shot 2017-10-28 at 2 09 55 am

With flex-wrap:
screen shot 2017-10-28 at 2 13 39 am

lib/modules/settingsConsole.js
```diff
@@ -799,6 +829,10 @@ function stageCurrentModuleOptions() {
 							const tempArray = input.value.split(',');
 							// convert the internal values of this array into their respective types (int, bool, bool, bool)
 							optionValue = [parseInt(tempArray[0], 10), (tempArray[1] === 'true'), (tempArray[2] === 'true'), (tempArray[3] === 'true')];
+						} else if (input instanceof HTMLSelectElement) {
+							if (input.selectedIndex >= 0) {
```
larsjohnsen: `optionValue = input.value` can be used

lib/modules/userTagger.js
```diff
@@ -416,7 +473,30 @@ function populateDialog(tag: Tag, card) {
...
+			tag.load({ text, color });
+			tag.save();
+			card.close();
+		});
+		return element;
+	}
+
+	function saveAsPreset() {
+		const { text, color } = extract();
+		if (text && color) {
```
larsjohnsen: Alert user if this conditional fails

lib/modules/userTagger.js
```diff
@@ -365,6 +416,10 @@ function populateDialog(tag: Tag, card) {
 				<label class="fieldPair-label" for="userTaggerVotesDown" title="Downvotes you have given this redditor">Downvotes</label>
 				<input type="number" style="width: 50px;" id="userTaggerVotesDown" value="${tag.votesDown}">
 			</div>
+			<div class="fieldPair">
```
larsjohnsen: Add `${!presetTags.length && 'hidden'}` in order to hide when there's no presets.

lib/modules/userTagger.js
pll33: How should the alert be set up?

lib/modules/settingsConsole.js
pll33: Should I keep that `else if` for the sake of readability, or remove it since the next `else` is `optionValue = input.value` anyways?

lib/modules/userTagger.js
larsjohnsen: A simple `window.alert` will probably suffice.

lib/modules/userTagger.js
pll33: The `fieldPair` won't immediately show when a user clicks 'save as preset' without refreshing the dialog. I was thinking of removing `class="hidden"` with classList.remove() from `elements.presetTag.parentElement` when saveAs is clicked but is there an alternate/simpler way you were thinking?

lib/modules/settingsConsole.js
larsjohnsen: Doesn't seem necessary to keep it, as the final `else` is clear enough.

lib/modules/userTagger.js
larsjohnsen: If this has the `hidden` attribute by default, `buildPresetTagElement` can include the line `elements.presetTag.parentElement.hidden = false` to make it visible.

lib/modules/userTagger.js
```diff
@@ -416,7 +473,30 @@ function populateDialog(tag: Tag, card) {
...
+		});
+		return element;
+	}
+
+	function saveAsPreset() {
+		const { text, color } = extract();
+		if (text && color) {
+			elements.presetTag.append(buildPresetTagElement(text, color));
+			const newPresetTags = presetTags.concat([[text, color]]);
+			Options.stage.add(module.moduleID, 'presetTags', newPresetTags);
```
larsjohnsen: `newPresetTags` can be replaced with `[...presetTags, [text, color]]`.

larsjohnsen: Also, by using `Options.set` it is not necessary to `commit` afterwards.

lib/modules/userTagger.js
pll33: `display:flex` in fieldPair is overriding `display:none` in the hidden attribute which is why I was thinking to set up a `hidden` class. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden (see note)

larsjohnsen: Oh, right, that's problematic. Making `display: flex` not apply when the `hidden` attribute is present might be the easiest solution. https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/3b04202a7672a431df4adb5097183a9488ffc564/lib/css/modules/_hover.scss#L71 can be changed to `.fieldPair:not([hidden])`.

lib/modules/userTagger.js
pll33: Actually right now it seems like you're able to set up a preset tag in the settings console with no text. That will show up in the dialog under presets but tagged users won't show up on the dashboard since the text is considered null. 

So should users be able to set up tags without text, or if not, is there a way to validate that the fields in the preset table in the userTagger settings console are not empty?

Additionally it seems somewhere in my changes I have introduced a new bug that allows you to save a tag without any text. I don't really see where my additions to userTagger would cause this to happen but I'm taking a closer look. Also I'm not getting that issue in v5.8.6 on Chrome though I haven't checked out v5.9.0 or master yet.

pll33: Update: Looks like that bug is in `master` too but I've only tested it out with Firefox.

erikdesjardins: FWIW it is intended that you can create tags with no text, in case you just want to use colors to denote different users.

erikdesjardins:
Thanks; sorry about the delay!

PR #4482, opened 2017-10-19
keyboardNav: Move to the previously selected thing

larsjohnsen:
Shortcut: Ctrl+Z

I fairly often see something interesting just after starting to press the jump to the next comment thread-shortcut, so it's nice to have a way to quickly return to that position.

Tested in browser:  Chrome 63

### DIFF
--- a/locales/locales/en.json
+++ b/locales/locales/en.json
@@ -564,6 +564,12 @@
 	"keyboardNavMoveToParentDesc": {
 		"message": "Move to parent (in comments)."
 	},
+	"keyboardNavUndoMoveTitle": {
+		"message": "Undo Move"
+	},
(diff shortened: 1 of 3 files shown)


locales/locales/en.json
```diff
@@ -564,6 +564,12 @@
 	"keyboardNavMoveToParentDesc": {
 		"message": "Move to parent (in comments)."
 	},
+	"keyboardNavUndoMoveTitle": {
+		"message": "Undo Move"
+	},
+	"keyboardNavUndoMoveDesc": {
+		"message": "Move to the previously selected thing."
```
erikdesjardins: Maybe note that it only works once, i.e. you can't ctrl-z all the way back to the OP. Not sure the best way to phrase that, though.

PR #4483, opened 2017-10-20
Update Contributing doc to match new Chrome menus (Tools -> More tools)

mtblanton:
Relevant issue: None. Was going through Contributing doc as a new contributor and saw that the path to get to Chrome extensions had changed from Menu -> Tools -> Extensions to Menu -> More tools -> Extensions.

Tested in browser: Chrome
image


erikdesjardins:
Thanks!

PR #4487, opened 2017-10-22
Show nightModeAutomaticNightModeDenied message only when the user actually denied permission

wojtekmaj:
The problem was, that we were displaying the message about user denying location permission in case of ANY geolocation error. As a result, I - being in a high speed train at the moment - have received 865898 false alerts about denying location permission.

We need to decide how we should handle other kinds of errors. So far, I have just placed switch() in order to make it easy to implement.

### DIFF
--- a/lib/modules/nightMode.js
+++ b/lib/modules/nightMode.js
@@ -297,16 +297,24 @@ async function getNightModeTimes() {
 			} catch (err) {
 				console.warn('Failed to init automatic night mode:', err);
 
-				return Alert.open(i18n('nightModeAutomaticNightModeDenied', 'confirm'), { cancelable: true })
-					.then(() => {
-						// they clicked confirm, turn it off
-						Options.set(module, 'automaticNightMode', 'none');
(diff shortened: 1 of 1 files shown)


lib/modules/nightMode.js
```diff
@@ -297,16 +297,24 @@ async function getNightModeTimes() {
...
+								return {
+									startingTime: new Date(0),
+									endingTime: new Date(0),
+								};
+							});
+					case err.POSITION_UNAVAILABLE:
+					case err.TIMEOUT:
+					case err.UNKNOWN_ERROR:
+					default:
+						return null;
```
erikdesjardins: This function can't return `null` or it'll cause errors elsewhere, see the [Flow output in CI](https://travis-ci.org/honestbleeps/Reddit-Enhancement-Suite/jobs/291191270). We can probably just `throw err;` in this case.

wojtekmaj: Done, Sir! Apologies for the delay.

PR #4488, opened 2017-10-22
Adding keyboard shortcut to edit selected comment or self-post

### LINKED ISSUE #2466: "Edit my selected comment" keynav button
requested in chat by mutoid

Antrikshy:
This resolves #2466.
Tested in browser: Firefox Developer Edition (57.0b10).

Pressing 'e' triggers the edit button under a selected comment or self-post, currently enabled on comment and profile pages.

I am not all that familiar with the codebase, so I may have missed things. I hope it doesn't break existing functionality. Do the current integration tests support any way of testing features that require the user to be logged in?

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -573,7 +573,9 @@ module.options = {
 		description: 'keyboardNavEditDesc',
 		title: 'keyboardNavEditTitle',
 		mustBeLoggedIn: true,
-		callback: edit,
+		callback() {
+			click(getFirstElementInThingByQuery('.entry .edit-usertext', ASSERT, getSelected()));
+		},
 	},
 	followPermalink: {
 		type: 'keycode',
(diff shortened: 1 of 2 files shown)


lib/modules/keyboardNav.js
```diff
@@ -1104,6 +1113,12 @@ function reply(selected = getSelected()) {
 	));
 }
 
+function edit(selected = getSelected()) {
+	if (selected.isComment() || selected.isSelfPost()) {
```
larsjohnsen: Probably no need to have this conditional, as `.edit-usertext` will be available only things which can be edited.

lib/modules/keyboardNav.js
```diff
@@ -1104,6 +1113,12 @@ function reply(selected = getSelected()) {
...
+		click(getFirstElementInThingByQuery('.edit-usertext', ASSERT, selected));
```
larsjohnsen: `.entry .edit-usertext` so that child it doesn't check child comments (in case `e` is pressed on a non-editable thing).

lib/modules/keyboardNav.js
```diff
@@ -1104,6 +1113,12 @@ function reply(selected = getSelected()) {
 	));
 }
 
+function edit(selected = getSelected()) {
```
larsjohnsen: Might as well inline this in the shortcut definition since it's so small.

erikdesjardins:
Looks good, thanks!

PR #4494, opened 2017-10-25
Command line command for custom toggles

### LINKED ISSUE #4454: Command line command for custom toggles
<div class="md">
<p>!GitHub Custom Toggles CLI</p>
<p><code>tog[gle] :toggle name [on|off]</code></p>
</div>
<br><br>
http://ift.tt/2fXhxar

octagonal:
resolves #4454
Relevant issue: https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/4454
Tested in browser: Chrome (Version 61.0.3163.100 (Official Build) (64-bit))

### DIFF
--- a/lib/modules/customToggles.js
+++ b/lib/modules/customToggles.js
@@ -53,15 +53,26 @@ function anyTogglesEnabled(toggles) {
 }
 
 function registerCommandLine() {
+	const notFoundErr = 'Could not find any matching toggle';
+
+	const getName = value => Object.entries(togglesByName())
+		.map(([name]) => name)
+		.filter(name => name.toLowerCase().match(value.toLowerCase()))
+		.sort()[0];
+
 	CommandLine.registerCommand(/^tog$|^toggle$/, ' tog[gle] [name] - toggles a custom toggle',
-		() => {},
 		(command, value) => {
-			const toggles = togglesByName()[value];
-			if (toggles) {
-				$(`#${toggles[0][2]}`).parent().toggleClass('enabled', toggleToggle(toggles));
-			} else {
-				return 'Unknown custom toggle.';
+			const match = getName(value);
+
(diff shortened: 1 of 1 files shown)


lib/modules/customToggles.js
```diff
@@ -50,6 +52,20 @@ function anyTogglesEnabled(toggles) {
 		.some(enabled => enabled);
 }
 
+function registerCommandLine() {
+	CommandLine.registerCommand(/^tog$|^toggle$/, ' tog[gle] [name] - toggles a custom toggle',
+		() => {},
```
larsjohnsen: It would be nice to [match partial names](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/5d76b584ac180e36208ceb3939c2c08822900de6/lib/modules/settingsConsole.js#L71), and see what action this perform when executed (i.e. will this activate / disable the toggle).

lib/modules/customToggles.js
```diff
@@ -50,6 +52,20 @@ function anyTogglesEnabled(toggles) {
...
+		(command, value) => {
+			const toggles = togglesByName()[value];
+			if (toggles) {
+				$(`#${toggles[0][2]}`).parent().toggleClass('enabled', toggleToggle(toggles));
```
larsjohnsen: `onClickToggleMenuItem` can probably be used here.

lib/modules/customToggles.js
octagonal: Fixed! Btw, I noticed that the help message will only show if the helper function returns `null`. So in the module command it will never show. See my updated code

lib/modules/customToggles.js
octagonal: You were right, fixed!

PR #4503, opened 2017-10-26
Cache: Rename to LRUCache, and make typing stricter

mc10:
Since the `LRUCache` doesn't need all of the methods of `Map`, and has some type differences, it seems to make more sense to use composition over inheritance.


erikdesjardins:
Indeed, I shouldn't have shoehorned it in just to avoid forwarding a few methods...

Issue #4500, opened 2017-10-25
Set visible max char count on user tags, show full on hover.


BenMcGarry:
Related post: http://redd.it/78rkbl

00:33:33 <@SpookoBen> should we set a max char count for them?
00:34:15 <@SpookoBen> or max displayed until hover or something
00:34:20 <@andytuba> i like that better
00:34:25 <@andytuba> max displayed until hover

Around 30 chars before cut off and rest shown on hover.

niralpokal:
I can work on this if it is still available

niralpokal:
I was able to make a change that only displays 30 characters and will display the full text if the user hovers. I only made this change to user tagging, but are there other things that you can tag? Let me know if this was what you were looking for or if you want it to be different. 

Below is an example:
![ex](https://user-images.githubusercontent.com/14898222/32035491-83b2f210-b9ce-11e7-8d23-1187c1b522a9.png)

BenMcGarry:
Looks good to me, @andytuba?

andytuba:
Looks okay, could use an ellipsis to indicate "more here" -- and an option to disable truncating.

Could you put a PR, @niralpokal?

Thanks!

niralpokal:
I made a PR but it failed on the linting test, I can fix that up. One thing I wanted to ask was that @larsjohnsen  pointed out that maybe I could use CSS instead. I honestly didnt think of that but that could work aswell. I can change it to CSS but I was wondering if you guys preferred I apply the style inline or with a class?

PR #4505, opened 2017-10-27
Truncated long user tags. (#4500)

niralpokal:
Relevant issue: #4500 


larsjohnsen:
I wonder if CSS would be more suitable here. Something like `width: 30ch; text-overflow: ellipsis;` and perhaps transition to full-width on hover.

niralpokal:
@larsjohnsen  I didn't even consider that, that would work really well. I asked the question in the main issue and wanted to get more thoughts on what they preferred.

PR #4509, opened 2017-10-28
Truncate user tags if too long. (#4500)

niralpokal:
Relevant issue: #4500 
Tested in browser:  Chrome, Firefox

I tried to use css but the max-width: 30ch; Did not seem to work well with anchor tags.

### DIFF
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -80,6 +80,13 @@ module.options = {
 		advanced: true,
 		dependsOn: options => options.trackVoteWeight.value,
 	},
+	truncateTag: {
+		title: 'userTaggerTruncateTagTitle',
+		type: 'boolean',
+		value: true,
+		description: 'userTaggerTruncateTagDesc',
+		advanced: true,
+	},
 };
 
 export const usernameRE = /(?:u|user)\/([\w\-]+)/;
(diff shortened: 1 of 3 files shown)


lib/modules/userTagger.js
```diff
@@ -158,9 +165,9 @@ export class Tag {
 				<a
 					class="userTagLink ${(text || color) ? 'hasTag' : 'RESUserTagImage'}"
 					${color && string._html`style="background-color: ${color}; color: ${bgToTextColorMap[color]} !important;"`}
-					title="set a tag"
+					title="${(text) ? text : 'set a tag'}"
```
andytuba: This could be simply `text || 'set a tag'`

andytuba:
> I tried to use css but the max-width: 30ch; Did not seem to work well with anchor tags.

Could you go into more detail?

niralpokal:
I looked into the CSS problem further and the problem has to do with the anchor tag being an inline element. I would have to change it display to inline-block or block.  I haven't tried that yet but I can test this and see if that works. 
 
I made a quick JS Fiddle for testing this: https://jsfiddle.net/6c3Lckua/

lib/modules/userTagger.js
niralpokal: Thanks! I didn't notice that, I can make that change!

niralpokal:
I updated the code to do the CSS way and it works. The only problem now is that it makes the overflow start at  38 characters and the tags display not centered. I can adjust these with the margin and changing the max-width property but I wanted to see what your thoughts were:
ex1

andytuba:
Neato. 30ish is fine. I'd say go ahead with fixing up the margins.

Issue #1748, opened 2014-11-02
"My subreddits" dropdown - text input to filter list


andytuba:
https://www.reddit.com/r/Enhancement/comments/2l2bym/request_when_opening_the_subreddit_drop_list_can/

PR #4510, opened 2017-10-28
Add quick-filter to my subreddits list

thybag:
Hello,

Add a quick-filter box to the top of the "my subreddits list", that allows people to quickly filter down their subreddit list to find the sub they are looking for.

Mostly inspired by how often i end up having to use the browsers built in search to find what i'm looking for in my subreddit list.

Have also added it to settings so that users can disable it if they don't like it there.

image

resolves https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/1748

### DIFF
--- a/lib/modules/subredditManager.js
+++ b/lib/modules/subredditManager.js
@@ -1202,13 +1202,13 @@ function addFilterSearch(theBody) {
 	const search = document.createElement('input');
 
 	tableCol.setAttribute('colspan', '3');
-	search.setAttribute('placeholder', 'Filter subreddits...');
+	search.setAttribute('placeholder', i18n('subredditManagerFilterPlaceholder'));
 
 	tableCol.appendChild(search);
 	tableRow.appendChild(tableCol);
 	tableHead.appendChild(tableRow);
 	// focus on open (delay to allow render)
-	setTimeout(() => { search.focus(); }, 100);
+	requestAnimationFrame(() => { search.focus(); });
 
 	// Filter subreddit list
 	search.addEventListener('keyup', () => {
--- a/lib/modules/subredditManager.js
+++ b/lib/modules/subredditManager.js
@@ -1201,7 +1201,7 @@ function addFilterSearch(theBody) {
 	const tableCol = document.createElement('td');
 	const search = document.createElement('input');
 
-	tableCol.setAttribute('colspan', 3);
+	tableCol.setAttribute('colspan', '3');
 	search.setAttribute('placeholder', 'Filter subreddits...');
 
 	tableCol.appendChild(search);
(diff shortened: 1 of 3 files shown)


thybag:
Unsure how to get around Flow not thinking a `Node` can have a `style` attribute. Any advice would be much appreciated.

lib/modules/subredditManager.js
```diff
@@ -1189,6 +1195,36 @@ function toggleSubredditDropdown(e: Event) {
...
+	const search = document.createElement('input');
+
+	tableCol.setAttribute('colspan', '3');
+	search.setAttribute('placeholder', 'Filter subreddits...');
+
+	tableCol.appendChild(search);
+	tableRow.appendChild(tableCol);
+	tableHead.appendChild(tableRow);
+	// focus on open (delay to allow render)
+	setTimeout(() => { search.focus(); }, 100);
```
erikdesjardins: `requestAnimationFrame` would probably be better here, since it'll be called immediately before the first paint after this is injected.

Still a bit sketchy, though I can't really think of a better option. (`autofocus` seems to only work once)

lib/modules/subredditManager.js
```diff
@@ -1189,6 +1195,36 @@ function toggleSubredditDropdown(e: Event) {
...
 }
 
+function addFilterSearch(theBody) {
+	const tableHead = document.createElement('thead');
+	const tableRow = document.createElement('tr');
+	const tableCol = document.createElement('td');
+	const search = document.createElement('input');
+
+	tableCol.setAttribute('colspan', '3');
+	search.setAttribute('placeholder', 'Filter subreddits...');
```
erikdesjardins: This text can be localized too; `i18n('subredditManagerFilterSubreddits')` or something

lib/modules/subredditManager.js
```diff
@@ -1189,6 +1195,36 @@ function toggleSubredditDropdown(e: Event) {
...
+	tableRow.appendChild(tableCol);
+	tableHead.appendChild(tableRow);
+	// focus on open (delay to allow render)
+	requestAnimationFrame(() => { search.focus(); });
+
+	// Filter subreddit list
+	search.addEventListener('keyup', () => {
+		for (const row: HTMLElement of theBody.children) {
+			const subredditName = (row.firstChild && row.firstChild.innerText) ? row.firstChild.innerText.toLowerCase() : '';
+			if (search.value === '' || subredditName.indexOf(search.value.toLowerCase()) !== -1) {
```
erikdesjardins: `subredditName.includes(search.value.toLowerCase())`

lib/modules/subredditManager.js
```diff
@@ -1189,6 +1195,36 @@ function toggleSubredditDropdown(e: Event) {
...
+				row.style.display = 'table-row';
```
erikdesjardins: Nit: maybe unset this instead: `row.style.display = '';`, that way it's more "reverting" what we do in the else block (not that it's likely to ever matter, but ¯\\\_(ツ)_/¯)

thybag:
Requested changes applied 👍

erikdesjardins:
Thanks!

Issue #3502, opened 2016-10-02
Collapse/expand control for right sidebar (a la multireddit navbar)


andytuba:
https://www.reddit.com/r/Enhancement/comments/54hfeo/feature_request_grippy_for_right_side_bar/

Maybe autocollapse on small window sizes

andytuba:
Stretch goal: remember to stay collapsed, per subreddit

erikdesjardins:
If a control is added to the edge of the sidebar, care must be taken to avoid breaking subreddit styles. I could easily see it affecting absolutely-positioned sidebar elements (even when uncollapsed).

Because of that, and it being a de facto adblocker, and due to support requests when hiding the sidebar (re)moves other custom content, I'm not really in favour of implementing this.

Also note: if you know what you're doing [read: you accept that it will break stuff], this can be done manually with a custom toggle + a CSS snippet for `.side { display: none }`.

honestbleeps:
with a common moderator complaint being that people don't read the rules especially on mobile (due to lack of sidebar) I also have a hard time supporting the ability to hide it.

sparksis:
I've downloaded an extension (User CSS) and added the following to reddit.

```
@media (max-width: 960px) {
  .side {
    display: none;
  }
}

```

I think having a toggle in the options to inject this CSS would be amazing.  The media tag hides the side bar **when the screen size is less than or equal to half my screen width** (1080p).

I think this nicely addresses @honestbleeps issue of hiding the sidebar altogether while adding a quality of life improvement.  

PS. I have subreddit styles globally disabled so @erikdesjardins comment is still valid.

williamparry:
Waiting on @andytuba to review my [pull request](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4511).

PR #4511, opened 2017-10-28
First cut of hideSidebar module

williamparry:
Relevant issue: #3502
Tested in browser: Chrome

### DIFF
--- a/lib/modules/hideSidebar.js
+++ b/lib/modules/hideSidebar.js
@@ -1,0 +1,116 @@
+/* @flow */
+import { $ } from '../vendor';
+import { Module } from '../core/module';
+import { i18n } from '../environment';
+import * as Menu from './menu';
+import {
+	CreateElement
+} from '../utils';
+
+export const module: Module<*> = new Module('hideSidebar');
+
+module.moduleName = 'hideSidebarName';
+module.category = 'productivityCategory';
+module.description = 'hideSidebarTabDesc';
+
+let hideSidebar;
+let propToChange;
+let valueToChangeTo;
+let valueToChangeFrom;
+let $side;
+let $content;
+let contentOffset;
+let headerHeight;
+let documentHalf;
+let sideContent;
+let sideWidth;
+let rulesCheck = ['reddiquette', 'rules'];
+
+const toggleSidebar = () => {
+
+	if (!$content.data('sidebar-hidden')) {
+		let check = true;
+		if (rulesCheck.some(function (v) { return sideContent.indexOf(v) >= 0; })) {
+			check = confirm("Have you read the rules?");
+		}
+		if (check) {
+			doHideSidebar();
+		}
+	} else {
+		doShowSidebar();
+	}
+}
+
+function recursiveEach($element) {
+	$element.children().each(function () {
+		let $this = $(this);
+		let thisOffset = $this.offset();
+		if ($this.css("position") == "absolute" &&
+			thisOffset.top >= headerHeight &&
+			thisOffset.top <= contentOffset.top &&
+			thisOffset.left < documentHalf
+		) {
+			$this.css("visibility", "visible");
+		}
+		recursiveEach($this);
+	});
+
+}
+
+const doHideSidebar = () => {
+	$side.css({
+		"visibility": "hidden",
+		"width": 0
+	});
+	recursiveEach($side);
+	$content.css(`${propToChange}-right`, valueToChangeTo);
+	hideSidebar.innerText = "show sidebar";
+	$content.data('sidebar-hidden', true);
+}
+
+const doShowSidebar = () => {
+	$side.css({
+		"visibility": "visible",
+		"width": sideWidth + "px"
+	});
+	$content.css(`${propToChange}-right`, valueToChangeFrom);
+	hideSidebar.innerText = "hide sidebar";
+	$content.data('sidebar-hidden', false);
+}
+
+module.go = () => {
+	$side = $(".side");
+	$content = $(".content[role=main]");
+	documentHalf = $(document).width() / 2;
+	contentOffset = $content.offset();
+	headerHeight = $("#header[role=banner]").height();
+	sideContent = $side.text().toLowerCase();
+	sideWidth = $side.width();
+
+	let padding = $content.css("padding").split(" ");
+	let margin = $content.css("margin").split(" ");
+
+	if (padding[1] > margin[1]) {
+		propToChange = "padding";
+		valueToChangeFrom = padding[1];
+		valueToChangeTo = parseInt(padding[3]) > 0 ? padding[3] : margin[3];
+	} else {
+		propToChange = "margin";
+		valueToChangeFrom = margin[1];
+		valueToChangeTo = parseInt(margin[3]) > 0 ? margin[3] : padding[3];
+	}
+
+	let oldHideSidebar = CreateElement.tabMenuItem({
+		text: 'hide sidebar'
+	});
+
+	hideSidebar = document.createElement('a');
+	hideSidebar.innerText = oldHideSidebar.innerText;
+	$(oldHideSidebar).replaceWith(hideSidebar);
+
+	hideSidebar.addEventListener('click', e => {
+		e.preventDefault();
+		toggleSidebar();
+	});
+	// TODO: Localstorage
+};
--- a/locales/locales/en.json
+++ b/locales/locales/en.json
@@ -4132,5 +4132,21 @@
 	},
 	"menuGearIconClickActionToggleMenuNoHover": {
 		"message": "Open menu (no hover)"
+	},
+	"hideSidebarName": {
+		"message": "Hide Sidebar",
+		"description": ""
+	},
+	"hideSidebarDesc": {
+		"message": "show/hide the sidebar",
+		"description": ""
+	},
+	"hideSidebarTabTitle": {
+		"message": "Hide Sidebar",
+		"description": ""
+	},
+	"hideSidebarTabDesc": {
+		"message": "Hides the Sidebar",
+		"description": ""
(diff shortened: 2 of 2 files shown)


williamparry:
Ah I accidentally added commit 8859411

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+const doHideSidebar = () => {
+	$side.css({
+		"visibility": "hidden",
+		"width": 0
+	});
+	recursiveEach($side);
+	$content.css(`${propToChange}-right`, valueToChangeTo);
+	hideSidebar.innerText = "show sidebar";
```
andytuba: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/locales/locales/README.md#translating-modules

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+	if (padding[1] > margin[1]) {
+		propToChange = "padding";
+		valueToChangeFrom = padding[1];
+		valueToChangeTo = parseInt(padding[3]) > 0 ? padding[3] : margin[3];
+	} else {
+		propToChange = "margin";
+		valueToChangeFrom = margin[1];
+		valueToChangeTo = parseInt(margin[3]) > 0 ? margin[3] : padding[3];
+	}
+
```
andytuba: Everything above this line inside `go` should probably live in a `const setup = _.once(() => { ... }` that's called at the beginning of `toggleSidebar`

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+	let oldHideSidebar = CreateElement.tabMenuItem({
+		text: 'hide sidebar'
+	});
+
+	hideSidebar = document.createElement('a');
```
andytuba: from irc:
> [The default behavior of styling the toggle as "selected"] drew attention to the toggle state which didn't really serve a purpose

How about addressing exactly that -- remove the `selected` styling (or classname) instead of rebuilding the whole element?  Or just leave it using the default behavior. IMO it's better to highlight that non-default state, even if it's the user's "default" preference, in order to help the user diagnose where the sidebar disappeared to.

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+		propToChange = "padding";
+		valueToChangeFrom = padding[1];
+		valueToChangeTo = parseInt(padding[3]) > 0 ? padding[3] : margin[3];
+	} else {
+		propToChange = "margin";
+		valueToChangeFrom = margin[1];
+		valueToChangeTo = parseInt(margin[3]) > 0 ? margin[3] : padding[3];
+	}
+
+	let oldHideSidebar = CreateElement.tabMenuItem({
```
andytuba: `const`

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+let contentOffset;
+let headerHeight;
+let documentHalf;
+let sideContent;
+let sideWidth;
+let rulesCheck = ['reddiquette', 'rules'];
+
+const toggleSidebar = () => {
+
+	if (!$content.data('sidebar-hidden')) {
```
andytuba: Any particular reason to store this in `$.data` instead of a variable here in the module?

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+module.go = () => {
+	$side = $(".side");
+	$content = $(".content[role=main]");
+	documentHalf = $(document).width() / 2;
+	contentOffset = $content.offset();
+	headerHeight = $("#header[role=banner]").height();
+	sideContent = $side.text().toLowerCase();
+	sideWidth = $side.width();
+
+	let padding = $content.css("padding").split(" ");
```
andytuba: `const` -- you're never going to change the value of `padding`. same with `margin`

andytuba: It appears you expect `.css('padding')` to return all four values at once, which is an implementation detail of jQuery that I wouldn't trust. Instead, you should specifically call `.css('padding-top')` and `.css('padding-left')` etc to get the exact values you need.  Additionally, you should put them in more descriptively named variables like `paddingTop`, `paddingLeft`.

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+
+	hideSidebar = document.createElement('a');
+	hideSidebar.innerText = oldHideSidebar.innerText;
+	$(oldHideSidebar).replaceWith(hideSidebar);
+
+	hideSidebar.addEventListener('click', e => {
+		e.preventDefault();
+		toggleSidebar();
+	});
+	// TODO: Localstorage
```
andytuba: What do you expect to put in localStorage, which subreddits have sidebar collapsed? Maybe instead you could make a module table option listing subreddit and "hidden" toggle.

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+			thisOffset.left < documentHalf
+		) {
+			$this.css("visibility", "visible");
+		}
+		recursiveEach($this);
+	});
+
+}
+
+const doHideSidebar = () => {
```
andytuba: These could simply be named `hideSidebar` and `showSidebar`.

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+		"visibility": "hidden",
+		"width": 0
+	});
+	recursiveEach($side);
+	$content.css(`${propToChange}-right`, valueToChangeTo);
+	hideSidebar.innerText = "show sidebar";
+	$content.data('sidebar-hidden', true);
+}
+
+const doShowSidebar = () => {
```
andytuba: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4511/files#r147560466

lib/modules/hideSidebar.js
```diff
@@ -0,0 +1,116 @@
...
+	contentOffset = $content.offset();
+	headerHeight = $("#header[role=banner]").height();
+	sideContent = $side.text().toLowerCase();
+	sideWidth = $side.width();
+
+	let padding = $content.css("padding").split(" ");
+	let margin = $content.css("margin").split(" ");
+
+	if (padding[1] > margin[1]) {
+		propToChange = "padding";
```
andytuba: May as well `propToChange = 'padding-right'` here instead of interpolating it later.

lib/modules/hideSidebar.js
williamparry: Re: ```const``` - sure.
Re: ```.css('padding')``` I did some investigation, and while running the function several times has a [performance hit](https://jsperf.com/res-jquery-shorthand-test/1) I looked up how [jQuery does the calculation](https://github.com/jquery/jquery/blob/master/src/css/var/getStyles.js#L15) and did a test of my own:

```
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="UTF-8">
	<meta name="viewport" content="width=device-width, initial-scale=1.0">
	<meta http-equiv="X-UA-Compatible" content="ie=edge">
	<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
	<title></title>

	<style>
		#test {
			padding-left: 10px;
			padding-top: 15px;
			/*margin-left: 20px;*/
		}
	</style>

</head>
<body>
	
	<div id="test"></div>

	<script>
		
	$(function() {

		var $t = $("#test");
		var pr = $t.css("padding-right");
		var pl = $t.css("padding-left");
		var mr = $t.css("margin-right");
		var ml = $t.css("margin-left");

		console.log(pr, pl, mr, ml);

	});

	$(function() {

		var $t = $("#test");
		var ps = $t.css("padding").split(" ");
		var ms = $t.css("margin").split(" ");

		// If no margin set, returns undefined, which will break parseInt
		console.log(ps[1], ps[3], ms[1], ms[3]);

	});
	</script>

</body>
</html>
```

Coupled with this from the [jQuery docs](http://api.jquery.com/css/):

> Retrieval of shorthand CSS properties (e.g., margin, background, border), although functional with some browsers, is not guaranteed.

It shows you're right - I'll make it more explicit.

lib/modules/hideSidebar.js
williamparry: I've put in:

```hideSidebarMenuItem.parentNode.classList.remove("selected")``` instead of rebuilding the element.

Given the action to hide the sidebar is explicitly triggered by the user and the functionality is to do with the aesthetics of the screen (readability etc) I'd argue that having the selected state for this toggle is counter-intuitive:

screen shot 2017-10-29 at 12 08 21 pm

However, I'm happy to leave it as the default (there's consistency argument). Let me know.

lib/modules/hideSidebar.js
williamparry: Done.

lib/modules/hideSidebar.js
williamparry: Done.

lib/modules/hideSidebar.js
williamparry: Done.

lib/modules/hideSidebar.js
williamparry: Done.

lib/modules/hideSidebar.js
williamparry: Done. Please check (and check how I set up ```i18n``` in ```module.beforeLoad```)

lib/modules/hideSidebar.js
williamparry: Done.

PR #4513, opened 2017-10-29
Prepare workaround for potential extension CSS precedence change

erikdesjardins:
Tested in browser: Chrome 64, Firefox 55, Edge 16

This change may be backed out of Chromium; it does make !important rules more powerful, but it makes non-!important rules much less useful.

Won't be merged until shortly before 64 goes stable (probably).


erikdesjardins:
Okay, now I highly doubt this will be reverted in Chromium. And reverting this PR won't be difficult if it does happen.

PR #4518, opened 2017-10-30
Use openerTabId with openNewTabs on Firefox

### LINKED ISSUE #190: Shift + Enter on a link will open a tab at the end of the tab bar, rather than next to the current one
I'm using RES 4.1.3 installed from AMO on Firefox 15 in a clean profile with default settings.  Pressing Shift + Enter on links will open the link in a new tab at the end of the tab bar, rather than next to the current tab.  This is in contrast to opening comments in a new tab (Shift + C), which opens the tab next to the current one as expected.  As you might imagine, this is annoying for keyboard navigation since it means I typically have to Control+Tab through multiple intermediary tabs to get back to the original Reddit tab I opened the link from.

As I understand it, the difference here between opening links and comments is that links use tabs.open provided by the Addon SDK API:

https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/dea134e085ab50010cda4be51abc7f104ec8f473/lib/reddit_enhancement_suite.user.js#L5322
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/06671380eaa199ecbcddbd4b22b781a6805c79cd/XPI/lib/main.js#L207

...While comments just call window.open:

https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/dea134e085ab50010cda4be51abc7f104ec8f473/lib/reddit_enhancement_suite.user.js#L5359

...Which is consistent with Firefox asking me to allow popups when trying to open comments in a new tab, but not asking me to do so for links.  The same bad tab opening behavior happens when using the number keys to follow links in comments.

This is probably specific to the XPI version of RES, but I haven't tested it elsewhere.

### LINKED ISSUE #4230: Single Click Opener placing tabs in the wrong location
- My browser (copy from https://www.whatismybrowser.com): Firefox 54 on Windows 10
- My RES version: 5.6.1

*What's the problem?*
"Single Click Opener" now opens the link and comments in two tabs next to my current tab, instead of at the end. I have browser.tabs.insertRelatedAfterCurrent to false which normally results in any new tab opening at the far right of any open tabs. 

https://www.reddit.com/r/RESissues/comments/69r274/res_560_lc_lc_doesnt_respect_tab_opening_location/
https://www.reddit.com/r/Enhancement/comments/69x8nc/new_behaviour_for_single_click_opener/
https://www.reddit.com/r/RESissues/comments/6a9hmr/single_click_opener_putting_tabs_in_the_wrong_spot/

Add-On Compatibility Reporter
Bulk Media Downloader
Checker Plus for Gmail
F.B. Purity
Forecastfox (fix version)
FrankerFaceZ
Google Search link fix
Greasemonkey
Image Search for Google
NativeShot
Nightly Tester Tools
OpenDownload2
Restart
S3.Google Translator
Stylish
Tab Mix Plus
uBlock Origin
Undo Close Tab
Youtube Plus

### LINKED ISSUE #4514: Opening comments in new tabs in clunky (should be easy fix)
- My browser: Firefox 57 on Windows 10 `(Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:57.0) Gecko/20100101 Firefox/57.0)`
- My RES version: v5.8.6 (though I'm looking at master code)

*What's the problem?*
When RES opens a new tab it does it using the code at [firefox/background.entry.js#L31](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/firefox/background.entry.js#L31), this means that if a user subsequently closes the tab, Firefox doesn't know which tab to go to. The intuitive behaviour would be to go back to the tab that opened the child tab.

As of version 57 of firefox (which will be released as stable in november), chrome.tabs.create supports openTabId ([mdn](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/create#Browser_compatibility)) which means that fixing this should be fairly simple.
I can submit a PR, but I'd like to know if my suspicion of correct first?

Not that I can find

Not relavant

bomjacob:
Relevant issue: fixes #4514 
Tested in browser: `Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:57.0) Gecko/20100101 Firefox/57.0`

Edit(erikdesjardins): fixes #4230
Edit(mc10): fixes #190

### DIFF
--- a/firefox/background.entry.js
+++ b/firefox/background.entry.js
@@ -28,13 +28,13 @@ addListener('download', ({ url, filename }) => {
 	chrome.downloads.download({ url, filename });
 });
 
-// Firefox doesn't support openerTabId, and needs cookieStoreId to open in correct container
-addListener('openNewTabs', ({ urls, focusIndex }, { index: currentIndex, cookieStoreId }) => {
+// Firefox needs cookieStoreId to open in correct container
+addListener('openNewTabs', ({ urls, focusIndex }, { id: tabId, cookieStoreId }) => {
 	urls.forEach((url, i) => {
 		chrome.tabs.create({
 			url,
 			active: i === focusIndex,
-			index: ++currentIndex,
+			openerTabId: tabId,
 			cookieStoreId,
 		});
 	});
--- a/firefox/beta/manifest.json
+++ b/firefox/beta/manifest.json
@@ -8,7 +8,7 @@
 	"applications": {
 		"gecko": {
 			"id": "jid1-xUfzOsOFlzSOXg@jetpack",
-			"strict_min_version": "55.0"
+			"strict_min_version": "57.0"
 		}
 	},
 	"icons": {
--- a/firefox/manifest.json
+++ b/firefox/manifest.json
@@ -8,7 +8,7 @@
 	"applications": {
 		"gecko": {
 			"id": "jid1-xUfzOsOFlzSOXg@jetpack",
-			"strict_min_version": "55.0"
+			"strict_min_version": "57.0"
 		}
 	},
 	"icons": {


firefox/background.entry.js
```diff
@@ -28,13 +28,13 @@ addListener('download', ({ url, filename }) => {
...
 
-// Firefox doesn't support openerTabId, and needs cookieStoreId to open in correct container
-addListener('openNewTabs', ({ urls, focusIndex }, { index: currentIndex, cookieStoreId }) => {
+// Firefox needs cookieStoreId to open in correct container
+addListener('openNewTabs', ({ urls, focusIndex }, { id: tabId, cookieStoreId }) => {
 	urls.forEach((url, i) => {
 		chrome.tabs.create({
 			url,
 			active: i === focusIndex,
-			index: ++currentIndex,
```
mc10: Is the `index` no longer necessary? It's still being used in Chrome's background.entry.js.

erikdesjardins: Yes; ideally we would never use `index`, but:
In Chrome if you don't specify `index` all tabs go to the end, instead of being grouped next to the `openerTabId` tab (like ctrl/middle-clicking would do).
In Firefox if you don't specify `index`, new tabs are grouped next to the `openerTabId` tab, or at the end if you've disabled `insertRelatedAfterCurrent`, which is #4230 (either way, the same as ctrl/middle-clicking).

PR #4519, opened 2017-10-30
[WIP] Integrate RES userinfo hover-card with new reddit-hovercards

thybag:
This is still in progress so code is still pretty dodgy.

Basic hover cards now look like
image
V2 profile hover cards now look like
image

Overview of features;
* tagger next to name
* Ignore user button added
* Highlight user button added

Issues;
* current method of detecting hover cards being loaded isn't great
* No has gold / gift gold button currently (unsure where to fit this)
* Quick compose is gone - left with reddits built in message
* Friend is gone - is this the same as reddits follow?

Be great to get some early feedback on this (currently old code is just at bottom of file & my code is a bit of a copy & pastey mess)

### DIFF
--- a/lib/modules/userInfo.js
+++ b/lib/modules/userInfo.js
@@ -107,31 +107,28 @@ module.go = () => {
 		//author-tooltip 
 		const addedNode = downcast(mutation.addedNodes[0], HTMLElement);
 		if (addedNode.classList.contains('author-tooltip')){
-			setTimeout(() => { renderHoverCardInfo(addedNode) }, 500);
-			
+			setTimeout(() => { renderHoverCardInfo(addedNode); }, 500); // Need better way to await reddit data load
 		}
 	});
 };
 
-
 async function renderHoverCardInfo(hoverCard) {
 	const authorLink = hoverCard.querySelector('.author-tooltip__profile-link');
-	const header = hoverCard.querySelector('.author-tooltip__head');
-	const links = hoverCard.querySelector('.author-tooltip__link-list');
-
 	const [, username] = UserTagger.usernameRE.exec(authorLink.href);
-
-	applyUserTag(hoverCard, header, links, username);
-
+	applyResData(hoverCard, username);
 }
 
-async function applyUserTag(hoverCard, header, links, username) {
+async function applyResData(hoverCard, username) {
+	const header = hoverCard.querySelector('.author-tooltip__head');
+	const links = hoverCard.querySelector('.author-tooltip__link-list');
+	const $hoverCard = $(hoverCard);
+
 	const userTaggerEnabled = Modules.isRunning(UserTagger);
 	const userTag = userTaggerEnabled && await UserTagger.Tag.get(username);
 
 	let userHTML;
 	if (userTaggerEnabled) {
-		userHTML = `<div class="fieldPair"><div class="fieldPair-label">${i18n('userInfoUserTag')}</div> <div class="fieldPair-text" style="display:flex"><a class="author" style="display: none;" href="/u/${username}"/></div></div>`;
+		userHTML = `<a class="author" style="display: none;" href="/u/${username}"/>`;
 	}
 
 	if (userTag && userTag.link) {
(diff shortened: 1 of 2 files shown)


andytuba:
Friends is different from follow. Friends is adding somebody to r/friends, follow is subscribing to their profile subreddit.

thybag:
image
currently unsure how to best handle links overflowing as 5 links looks a bit ugly

thybag:
image
Unsure what people think of this approach - Have added the gold logo to top right of users who have reddit gold?

BenMcGarry:
@thybag i would say use the gold icon instead of that on the hover card, would look a lot neater and i would assume most users know what it means.

thybag:
Removed gold icon since it seems reddit already has its own >.<

lib/modules/userInfo.js
```diff
@@ -101,61 +96,160 @@ module.options = {
...
+ * attachUserTagger
+ *
+ * @param Element header
+ * @param string username
+ */
+async function attachUserTagger(header: HTMLElement, username) {
+	const userTaggerEnabled = Modules.isRunning(UserTagger);
+	const userTag: Tag = userTaggerEnabled && await UserTagger.Tag.get(username);
+
+	if (userTaggerEnabled) {
```
erikdesjardins: In general Flow doesn't do refinement "backwards", i.e. inference proceeds as:
- `userTaggerEnabled` inferred as `boolean` by its declaration
- `userTag` inferred as `false | <return value of Tag.get>` by its declaration
- `userTaggerEnabled` refined to `true` within the if-statement, but this doesn't propagate "backwards" to the existing declaration of `userTag`

You can either replace `if (userTaggerEnabled)` with `if (userTag)`, or change this to:

```js
if (Modules.isRunning(UserTagger)) {
    const userTag = await UserTagger.Tag.get(username);
    // ...
    return userTag;
}
return false;
```

which IMO is simpler for humans (not just Flow) to understand.

thybag: Hmm, wonder how/why it worked initially as that was pretty much a copy and paste, prefer your suggestion tho, will update

mc10:
If you wanted to keep quick compose, you could attach a click handler to the "Message" link and have it show RES's compose panel instead, probably with an option to turn off this functionality.

lib/modules/userInfo.js
```diff
@@ -101,61 +96,142 @@ module.options = {
...
 module.go = () => {
-	if (module.options.hoverInfo.value) {
-		$(document.body).on('mouseover', UserTagger.usernameSelector, handleMouseOver);
-	}
+	observe(document.body, { childList: true }, mutation => {
+		if (!mutation.addedNodes.length) return;
+		// author-tooltip
+		const addedNode = downcast(mutation.addedNodes[0], HTMLElement);
+		if (addedNode.classList.contains('author-tooltip')) {
+			setTimeout(() => renderHoverCardInfo(addedNode), 50);
```
larsjohnsen: Can be avoided with
```
waitForChild(addedNode, '.author-tooltip__profile-link')
  .then(link => {
     const [, username] = UserTagger.usernameRE.exec(authorLink.getAttribute('href'));
     applyResData(addedNode, username);
  })
);
```

thybag: Doesn't `waitForChild` only work for direct children? (the `author-tooltip__profile-link` is fairly nested)

larsjohnsen: You're right. A similar function which observes with option `subtree: true` would have to be added.

thybag:
Hopefully can switch to using https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4526 once its all merged 👍

larsjohnsen:
@thybag #4526 has been merged

lib/modules/userInfo.js
```diff
@@ -101,61 +73,131 @@ module.options = {
...
+ * @param Element hoverCard
+ * @param string username
+ */
+async function applyResData(hoverCard, username) {
+	// Get main elements
+	const header = hoverCard.querySelector('.author-tooltip__head');
+	const links = hoverCard.querySelector('.author-tooltip__link-list');
+
+	// Render tagger
+	const userTag = await attachUserTagger(header, username);
```
larsjohnsen: To avoid adding too much visual noise, the tagger probably ought to be placed next to the username.  `userTag.link` doesn't seem to fit well inside the new userInfo, so the link can probably be ignored as it seemingly is of low value.

The `attachUserTagger` function can then pretty much be replaced with:
```
UserTagger.applyToUser(
	downcast(header.querySelector('.author-tooltip__profile-link'), HTMLAnchorElement),
	{ renderTaggingIcon: true, renderVoteWeight: false }
);
```

thybag: Possibly - next to the name felt more natural since its with reddits own additional icons get added (gold, cake day etc), vs next the the /u/etc on the full profiles :/

larsjohnsen: A issue with placing it on the next line is that the native userTagger is position as if it is fixed height
image

andytuba: I'll ask about implementing a frontend API method for "fix repositioning of reddit user card".

PR #4522, opened 2017-10-31
Applying a preset will prompt the user to reload (#2711)

### LINKED ISSUE #2711: presets.cleanSlate should show prompt("Reload?") instead of notification

mtblanton:
Relevant issue: presets.cleanSlate should show prompt("Reload?") instead of notification (fixes #2711)
Tested in browser: Chrome and Firefox

This shows optionName instead of a blank field on the reload prompt. It should be fixed, but I wasn't able to figure out how to fix it.


erikdesjardins:
Thanks!

(CI failure is unrelated)

PR #4526, opened 2017-11-01
waitForDescendant and other MutationObserver utils

erikdesjardins:
Relevant issue: re https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4519/files#r148139027
Tested in browser: Chrome 62

~~Currently breaks things.~~


larsjohnsen:
This works fine if looking for existing matches is separated into `watchForFuture...` variants of the utils.

erikdesjardins:
Ah, yep, that'll do it.

PR #4527, opened 2017-11-01
gallery: Add button to switch from slideshow to filmstrip

larsjohnsen:
Useful for when a slideshow isn't suitable, like when there's a lot of text to each piece and the total height of the text and image exceeds the viewport height.

The `🎞` button to the right:
image

Edit: Changed character

Tested in browser: Chrome 63, Edge 16


erikdesjardins:
Seems like Windows doesn't have a font with that character:
image
(tried different font families, etc.)

erikdesjardins:
Now I can see the icon, but alignment is off:

image

erikdesjardins:
CI failure is something else

PR #4548, opened 2017-11-16
Add support for Google Poly media host

Ajedi32:
A small, pretty straightforward implementation of a media host for Google's new Poly 3D model catalog.

As far as I know Poly doesn't have an API, but it's pretty straightforward to just embed the viewer with an iframe.

Tested in Chrome 62.

Try it out here: https://www.reddit.com/domain/poly.google.com/


erikdesjardins:
👍

PR #4550, opened 2017-11-17
userTagger: Improve tagger alignment

larsjohnsen:
- Fixes height when there's color but no text
- Fixes some circumstances where it was misaligned with the username / voteweight.

Tested in browser: Firefox 57, Chrome 64


erikdesjardins:
CI failure is Flow timing out (working on it)

PR #4553, opened 2017-11-19
Fix settings layout to enhance clarity

wojtekmaj:
Settings, especially on non-English locales, tend to get messy. In Polish, often short setting names must be longer to make ANY sense. As a result, settings may look like this:

obraz

(and I couldn't be bothered to find the worst case)

After the fix, settings behave like so:

* Label takes 20%, but not less than 200px,
* Setting takes 20%, grows when necessary in some special cases
* Description takes whatever is left

and it looks like that:

obraz

### DIFF
--- a/lib/css/res.scss
+++ b/lib/css/res.scss
@@ -605,8 +605,8 @@ div.usertext-edit {
 
 .optionTitle {
 	font-size: 14px;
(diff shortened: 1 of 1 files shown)


erikdesjardins:
Thanks!

PR #4555, opened 2017-11-19
package.json: Update deps/devDeps

mc10:
Didn't update JSZip, as its API actually changed.


package.json
```diff
@@ -81,9 +81,9 @@
     "autoprefixer": "7.1.6",
     "ava": "0.23.0",
     "babel-core": "6.26.0",
-    "babel-eslint": "7.2.3",
+    "babel-eslint": "8.0.2",
```
erikdesjardins: babel-eslint 8 depends on Babel 7, so we shouldn't update it yet

package.json
```diff
@@ -109,7 +109,7 @@
     "extricate-loader": "2.0.0",
     "file-loader": "1.1.5",
     "firefox-extension-deploy": "1.1.0",
-    "flow-bin": "0.53.1",
+    "flow-bin": "0.59.0",
```
erikdesjardins: This is going to require changes for the build to pass (#4490)

PR #4564, opened 2017-11-27
selectedEntry: Use class instead of attribute

larsjohnsen:
Using attribute causes a style recalculation on comment pages, though I don't know why that's the case.

Tested in browser: Chrome 64


erikdesjardins:
There's also one usage in `_commentQuickCollapse.scss`, I believe

PR #4566, opened 2017-11-27
(Optionally) render crosspost metadata above expando (all types)

### LINKED ISSUE #4385: New x-post support
As per modmail we have a few incompatibilities, listed here for tracking.

- [x] ~~RES selector colour extends into body of x-post.~~ Appears to be intended, background is set to transparent on `.link .crosspost-preview-content .usertext-body .md`

https://i.imgur.com/fzShNDQ.png
- [x] Admin colour not present on x-posts

https://i.imgur.com/rkFUDJo.png

- [ ] Image x-posts do not show as intended, overwritten by RES expando.

Expected: 

Actual:

- [x] X-post requires nightmode support

https://i.imgur.com/ciKBBKp.png

This is what i could find from a brief look.

andytuba:
Relevant issue: fixes #4385 
Tested in browser:  Chrome

(Ignore "xposted from ..." in the tagline of the screenshot, that's just a quirk in the xpost module due to my phrasing of the title.)

crosspost

crosspost-resize

screen shot 2017-11-26 at 9 39 52 pm

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1069,10 +1069,6 @@ const retrieveExpandoOptions = _.memoize(async (thing, { siteModule, detectResul
 		mediaOptions.title = '';
 	}
 
-	if (module.options.crossposts.value === 'withMetadata' && thing.isCrosspost()) {
-		mediaOptions.crosspostData = thing.element.dataset;
-	}
-
 	const attribution = module.options.showSiteAttribution.value &&
 		thing && thing.isPost() && !thing.isSelfPost() &&
 		siteModule.domains.length && siteModule.attribution !== false;
@@ -1092,6 +1088,9 @@ const retrieveExpandoOptions = _.memoize(async (thing, { siteModule, detectResul
 		},
 		generateMedia() {
 			const element = generateMedia(mediaOptions, { href });
+			if (module.options.crossposts.value === 'withMetadata' && thing.isCrosspost()) {
+				element.prepend(crosspostMetadataTemplate(thing.element.dataset));
+			}
 			if (attribution) addSiteAttribution(siteModule, element);
 			return element;
 		},
@@ -1206,14 +1205,6 @@ function generateMedia(options: ExpandoMedia, context: {| href: string |}) {
 			break;
 		default:
 			throw new Error(`Unreachable: invalid media type ${options.type}`);
-	}
-
-	if (options.crosspostData) {
-		try {
-			element.prepend(crosspostMetadataTemplate(options.crosspostData));
-		} catch (e) {
-			console.error(`showImages: could not create crosspost metadata expando for ${context.href}`, e);
-		}
 	}
 
 	return element;
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -941,7 +941,7 @@ async function checkElementForMedia(element) {
 
 	if (!mediaInfo) return;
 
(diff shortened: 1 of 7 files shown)


lib/modules/showImages.js
```diff
@@ -1185,6 +1210,14 @@ function generateMedia(options: ExpandoMedia, context: {| href: string |}) {
 			throw new Error(`Unreachable: invalid media type ${options.type}`);
 	}
 
+	if (options.crosspostData) {
+		try {
+			element.prepend(crosspostMetadataTemplate(options.crosspostData));
+		} catch (e) {
+			console.error(`showImages: could not create crosspost metadata expando for ${context.href}`, e);
```
erikdesjardins: Is this reachable? At a glance it doesn't appear that anything in `crosspostMetadataTemplate` can throw

lib/utils/dom.js
```diff
@@ -343,3 +343,16 @@ function addStyle(css) {
 
 	return style;
 }
+
+export function dataAttributes(element: HTMLElement): any {
```
erikdesjardins: Can [dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) be used instead?

tests/showImages.js
```diff
@@ -106,6 +106,18 @@ module.exports = {
...
 	},
+	'crosspost expando': browser => {
+		browser
+			.url('https://en.reddit.com/by_id/t3_7fsw5w/')
+			.waitForElementVisible('.expando-button')
+			.click('.expando-button')
+			.waitForElementVisible('.res-expando-box')
+			.assert.cssClassPresent('.crosspost-preview', 'res-crosspost-preview')
+			.assert.containsText('.crosspost-preview .title', 'Original post with expandoable image for crossposting')
+			.assert.attributeEquals('.crosspost-preview a.author', 'href', '/user/andytubatest/')
```
erikdesjardins: The original author is /u/andytuba, so this should be /user/andytuba/, I believe.

You may also need to use `attributeContains`, since I think it does the `element.href` URL expansion (i.e. it's not just `element.getAttribute('href')` like the name suggests...)

lib/modules/showImages.js
andytuba: It might fail if Reddit changes the data attributes.

erikdesjardins: How though? ``` string.html` "${null}" "${undefined}" "${false}" ` === ` "" "" "" ` ```, and all the property accesses are one level deep, so it seems like it would just render nothing instead of throwing?

lib/modules/showImages.js
```diff
@@ -1185,6 +1210,14 @@ function generateMedia(options: ExpandoMedia, context: {| href: string |}) {
 			throw new Error(`Unreachable: invalid media type ${options.type}`);
 	}
 
+	if (options.crosspostData) {
```
larsjohnsen: I think this is a better fit in [`generateMedia` in `retrieveExpandoOptions`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/c65e3cc6e66d14d96efaafcef48f4aff36376c50/lib/modules/showImages.js#L1070), alongside with adding the site attribution. That way `crosspostData` can be accessed directly from `thing` (and not stored as `crosspostData`).

andytuba:
Thanks for the polish!

PR #4570, opened 2017-12-02
Init/profiling cleanup, eliminate some cyclic dependencies

erikdesjardins:
Tested in browser: Chrome 64, Firefox 57


lib/modules/styleTweaks.js
```diff
@@ -365,16 +365,12 @@ async function toggleSubredditStyle(toggle, subreddit = currentSubreddit()) {
 
 const toggleElements = (() => {
 	const getStylesheet = _.once(async () => {
-		const query = () => (
-			document.head.querySelector('link[title=applied_subreddit_stylesheet]') ||
-			document.head.querySelector('style[title=applied_subreddit_stylesheet]') ||
```
erikdesjardins: This is no longer used by reddit...

lib/modules/styleTweaks.js
```diff
@@ -365,16 +365,12 @@ async function toggleSubredditStyle(toggle, subreddit = currentSubreddit()) {
...
-			document.head.querySelector('style[data-apng-original-href]') // apng extension fix (see #1076)
```
erikdesjardins: ...and APNG is unnecessary now that Chrome has native support.

lib/core/profiling.js
```diff
@@ -1,72 +0,0 @@
...
-		const time = (end[b] - end[a]) | 0;
-		return `<span style="color: ${time < 0 ? 'green' : 'red'}">${time}</span>ms`;
-	}
-
-	afterLoad.then(() => {
-		Alert.open([
-			`beforeLoad stalled for ${diff('headReady', 'loadOptions')}`,
-			`go stalled for ${diff('bodyReady', 'beforeLoad')}`,
-			`afterLoad stalled for ${diff('loadComplete', 'go')}`,
-			`addModuleBodyClasses was late by ${diff('bodyStart', 'loadOptions')}`,
```
erikdesjardins: This kind of profiling was misleading, since it was more dependent on how the browser decided to schedule extensions than actually how fast or how soon they ran.

Issue #1770, opened 2014-11-11
RES Icons blurry on high DPI displays


imjoshdean:
On high DPI displays, such as Macbook Pro with Retina, the RES icon spite map is blurry. These should be updated to work. The process is (in theory) simple:
1. Increase the original images to 2x size.
2. Keep the same positioning in the sprite image
3. Update the CSS to include background-size property set equal to the height and width of the original
4. ???
5. Profit!

larsjohnsen:
Seems to have been fixed.

PR #4572, opened 2017-12-04
Replace expando button images with SVGs

wojtekmaj:
As a someone with HiDPI screen my eyes were bleeding when I looked at current expando buttons, so I fixed them.

Generally, I tried to re-create them pixel perfect, with one exception: all pluses are now slightly larger to match native Reddit expando buttons.

Here's a comparison. 1st column are old expando buttons, 2nd are the new shiny SVGs and 3rd are native Reddit expandos.

obraz

Tested on Chrome, looks sweet.

obraz


erikdesjardins:
Part of #1770

PR #4581, opened 2017-12-08
Remove other favicons to prevent Chrome from switching to them

erikdesjardins:
Relevant issue: https://www.reddit.com/r/Enhancement/comments/7i99t5/orange_icon_on_browser_tab_with_res_enabled/
Tested in browser: Chrome 64

### DIFF
--- a/lib/modules/orangered.js
+++ b/lib/modules/orangered.js
@@ -194,6 +194,9 @@ function updateFaviconBadge(count) {
 const setupFaviconBadge = _.once(() => {
 	if (!module.options.showUnreadCountInFavicon.value) return false;
 
+	// remove other icons, since Chrome (bizarrely) switches to one of them if RES changes the favicon
+	for (const e of document.head.querySelectorAll('link[rel="icon"]')) e.remove();
+
 	const favicon = new Favico();
 
 	if (module.options.resetFaviconOnLeave.value) {


lib/modules/orangered.js
```diff
@@ -196,6 +196,9 @@ const setupFaviconBadge = _.once(() => {
 
 	const favicon = new Favico();
 
+	// remove other icons, since Chrome (bizarrely) switches to one of them if RES changes the favicon
+	for (const e of document.head.querySelectorAll('link[rel="icon"]')) e.remove();
```
andytuba: Maybe move this above `new Favico()` just in case its initializer tries to sniff the existing icons?

corylulu:
@erikdesjardins 

> If links for both PNG and ICO favicons are present, PNG-favicon-compatible browsers select which format and size to use as follows. Firefox and Safari will use the favicon that comes last. Chrome for Mac will use whichever favicon is ICO formatted, otherwise the 32×32 favicon. Chrome for Windows will use the favicon that comes first if it is 16×16, otherwise the ICO. If none of the aforementioned options are available, both Chromes will use whichever favicon comes first, exactly the opposite of Firefox and Safari. Indeed, Chrome for Mac will ignore the 16×16 favicon and use the 32×32 version, only to scale it back down to 16×16 on non-retina devices. Opera will choose from any of the available icons completely at random.

Based on this, I found that the following solves the issue:

`$("link[rel='shortcut icon']").attr("sizes", "16x16 32x32");` 

Don't actually need to remove the old ones, just specify `sizes="16x16 32x32"` and it worked. I'm guessing after that, it takes priority since it's a base64 string.

erikdesjardins:
Good to know, thanks. I tried that yesterday and it worked, but now a bit more is changed on reddit's side and it doesn't. 

So I think #4590 will be a bit less fragile than relying on Chrome continuing to pick `"16x16 32x32"`. (Certainly, blindly deleting them, like _this_ PR, isn't)

corylulu:
Yeah, same thing happened for me. I would just do both to double up, but yeah, the current implementation should work.

PR #4586, opened 2017-12-09
Add an option to hide other expandos on expando open

wojtekmaj:
So, here's an idea.

When you browse Reddit using your keyboard, previously opened expandos close nicely as you select the next thing on the page. Mouse users need to do it manually, which is especially painful and necessary if you're browsing through videos.

This option may potentially conserve memory and help with a mess when browsing Reddit.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -434,6 +434,12 @@ module.options = {
 		value: true,
 		description: 'showImagesAutoplayVideoDesc',
 	},
+	hideOtherExpandos: {
+		title: 'showImagesHideOtherExpandosTitle',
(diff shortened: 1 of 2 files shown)


lib/modules/showImages.js
```diff
@@ -968,6 +974,18 @@ async function checkElementForMedia(element) {
 
 	if (thing) thingExpandoBuildListeners.fire(thing);
 
+	if (module.options.hideOtherExpandos.value) {
```
erikdesjardins: @larsjohnsen is this a good place to add this listener? And do you think it might interfere with anything else that automatically opens/closes expandos?

PR #4587, opened 2017-12-10
Use CSS transforms on expando media resize for 60 FPS experience

wojtekmaj:
**The problem:** Resizing media causes thousands of reflows, DOM events etc. On all low- and mid- range devices (and even on my Core i5 Surface Pro on Firefox) performance of resizing is far from buttery smooth.

**Solution:** When still dragging, use CSS transforms to resize the media. That makes it full 60 FPS experience. Quite an improvement over 7.5 FPS I was getting.

However, this solution causes image container to be updated only after resizing ends, and not during it. So, some people might not like this behavior. In that case, this more GPU/CPU-friendly solution may be applied based on some internal performance measurements we would do or simply on some setting we could create.

What do you think?

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1707,16 +1707,19 @@ function setMediaMaxSize(media) {
 	if (maxHeight) media.style.maxHeight = `${maxHeight}px`;
 }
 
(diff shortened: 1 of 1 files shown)


larsjohnsen:
The relevant code has received some significant changes (#4012). Can you see whether the performances has improved?

PR #4588, opened 2017-12-12
eslint: Enable guard-for-in, new-cap, disallow `in`

mc10:
The changes should be pretty minor.


erikdesjardins:
Integration test failures are due to reddit's net neutrality modal, and this PR doesn't introduce any new ones

PR #4590, opened 2017-12-12
Select from (new) favicons

erikdesjardins:
Tested in browser: Chrome 64


erikdesjardins:
Integration test failure is on reddit's side

Issue #4560, opened 2017-11-26
[Feature Request]


JesseAldridge:
Track and display my upvote/downvote ratio for individual subreddits.  That way I can unsubscribe from subreddits with content that I consistently downvote.

dummycode:
I mentioned [something similar](https://www.reddit.com/r/Showerthoughts/comments/54x51k/spotify_should_add_a_feature_that_keeps_track_of/) on reddit forever ago for Spotify. My social media often gets cluttered and features like this always seem useful.

emilyford352:
Going to look into how extensive of a change this would be and if I have enough experience to fix it!

ClarSco:
This is already available on reddit (RES?).

It can be found by navigating to https://www.reddit.com/user/<your username>/ then clicking on "show karma breakdown by subreddit" in the sidebar.

andytuba:
@ClarSco That's how many points you've received from each subreddit. OP is requesting that RES track how many upvotes/downvotes the user has given in each subreddit.

PR #4599, opened 2017-12-22
Fix userTagger vote click when downvoting after upvoting

pll33:
Discovered while working on a feature with similar code for vote tracking #4560 

Actual: When upvoting then downvoting a post, it was counting as a negative downvote and incrementing the total vote weight. (up = 0, down = -1)

Expected: (up = -1, down = 1)

Tested in browser: Firefox


larsjohnsen:
Thanks!

PR #4613, opened 2018-01-09
getVideoTimes: Prevent error when thumbnails are disabled

larsjohnsen:


erikdesjardins:
Integration test failure due to reddit dying (presumably)

PR #4618, opened 2018-01-10
Contain subreddit list scrolling

Alexendoo:
Currently when the user scrolls to the bottom of the subreddit list the whole page scrolls down, this moves this subreddit list up causing some confusion. [`overscroll-behavior`](https://wicg.github.io/overscroll-behavior/) is a new property to deal with this, it's supported in chrome as of 63 (the current stable release) and will be available in firefox come version 59

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -32,7 +32,13 @@ rules:
   no-important: 0
   no-invalid-hex: 2
   no-mergeable-selectors: 2
-  no-misspelled-properties: 2
+  no-misspelled-properties:
+    - 2
+    -
+      extra-properties:
+        # necessary until a version of sass-lint is released that depends on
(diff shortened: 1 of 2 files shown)


andytuba:
Does the `sass-lint` package need upgrading to learn about this new property? Or should we trust in the w3c write-up and mark it "sass-lint ignore this"?

Alexendoo:
Whoops, I missed that sass-lint didn't like that property.

Looks like sass-lint does not yet have a release containing the property but it's in the develop branch sasstools/sass-lint@ba8887eac8061dbdd0487025f75137c9c31eeba6. I've added an exemption to `.sass-lint.yml` for the meanwhile.

erikdesjardins:
Looks good, thanks!

PR #4619, opened 2018-01-10
Sent RES update notification changelog url to individual page.

BenMcGarry:
Send the user to the individual change log page on the RES has been updated notification.


erikdesjardins:
I did this intentionally, so if we release a patch on the same day people still see the old changelog

erikdesjardins:
da2c3fdf21259d4ad130c4d7f16d1dcae936920d, though maybe worth a comment

BenMcGarry:
Ah ok, i didnt see that commit.

PR #4620, opened 2018-01-10
Fix typos in selectedEntry styles

larsjohnsen:
Restores CSS override for some custom subreddit styles.


erikdesjardins:
Double checking: this doesn't affect performance?

PR #4626, opened 2018-01-11
Filterline: Clarify title text for visiblity and temporarily disable toggle buttons

larsjohnsen:


erikdesjardins:
Hmm, I feel like we might still need to change this (being able to hide the filterline, but only temporarily, doesn't seem particularly useful, so I think some confusion is understandable), but this is fine for now

PR #4631, opened 2018-01-11
Use event.key over event.keyCode/which

mc10:
There are a few `keyCode` instances that I did not change, as they were trickier to modify correctly.

### DIFF
--- a/lib/modules/commentTools.js
+++ b/lib/modules/commentTools.js
@@ -1091,7 +1091,7 @@ function addAutoCompletePop() {
 	$('body').append($autoCompletePop);
 
 	$('body').on({
-		keyup: autoCompleteTrigger,
+		keypress: autoCompleteTrigger,
 		keydown: autoCompleteNavigate,
 		blur: autoCompleteHideDropdown,
 	}, '.usertext .usertext-edit textarea, #BigText, #wiki_page_content');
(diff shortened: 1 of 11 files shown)


lib/modules/commentTools.js
```diff
@@ -1078,24 +1078,6 @@ function lod() {
 	});
 } */
 
-export const KEYS = {
```
andytuba: Can we keep this as `NAMED_KEYS = { SPACE: ' ', ENTER: 'Enter', ESCAPE: 'Escape', ... }` for stronger validation?

mc10: Sure, sounds good to me.

andytuba:
Thanks! Neat little refactor.

erikdesjardins:
This needs testing in Edge (which I can do at some point), since apparently it has "non-standard key identifiers"... https://caniuse.com/#feat=keyboardevent-key

mc10:
Good catch; seems like the issue is [fixed but not yet released](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8860571/). We'll need to add a [shim similar to this](https://github.com/shvaikalesh/shim-keyboard-event-key/blob/master/index.js) for Edge to work.

BenMcGarry:
Depending when we expect this 5.12 to land we may be able to just avoid the shim. The fix should be released April timeish.

mc10:
I added a polyfill for Edge, so this should be mergeable. I don't have Edge to test with though.

erikdesjardins:
Tested in Edge 16

PR #4632, opened 2018-01-11
settingsConsole: Make minor cleanup changes

mc10:
Should be fairly straightforward.


lib/modules/settingsConsole.js
```diff
@@ -894,7 +895,7 @@ export function open(moduleIdOrCategory?: string) {
 	isOpen = true;
 	// hide the ad-frame div in case it's flash, because then it covers up the settings console and makes it impossible to see the save button!
 	const adFrame = document.getElementById('ad-frame');
```
erikdesjardins: We can probably remove this entirely, I don't think ads have had the id `ad-frame` for a long time

Issue #4434, opened 2017-09-22
[RE-Request] Locally store version of comment I replied to.


brendanvos:
It would be nice if users had the option to automatically locally store the contents of comments at the time the user replies to the comment.

This would allow the user to compare the comment at the time of reply to the current state of the comment after the OP has made edits.

The functionality of this feature could be expanded to also locally store additional versions of the comment after edits have been made, upon the user loading the comment again.

Criticalstone:
I would like to contribute to this, will take a look at it this weekend

andytuba:
If anyone wants to implement this, I would recommend adding a new option to res' "Save Comments" module which watches for comment form submitted (might be rough bc form submission is handled via Reddit J's) or new comment created from op

miggs125:
Is anyone assigned to this issue? I would like to work on it.

Criticalstone:
@miggs125 Seems like @asoderman made a PR on this issue? #4642

PR #4642, opened 2018-01-16
Comment save when replying 

asoderman:
I implemented a feature that saves a local copy of the comment the user is replying to. The user is able to display it by clicking the "previous version" button on the comment. 

Relevant issue: 
#4434

### DIFF
--- a/lib/modules/commentSaveOnReply.js
+++ b/lib/modules/commentSaveOnReply.js
@@ -1,0 +1,99 @@
+/* @flow */
+import { i18n } from '../environment';
+import { Module } from '../core/module';
+
+export const module: Module<*> = new Module('commentSaveOnReply');
+
+module.moduleName = 'commentSaveOnReplyName'; // add the friendly name to locales/locales/en.js for i18n
+module.category = 'commentsCategory'; // categories from locales/locales/en.js
+module.description = 'commentSaveOnReplyDesc'
+module.options = {
+};
+
+
+module.go = () => { // Optional: runs after <body> is ready and `beforeLoad` (in all modules) is complete
+	// Do stuff now!
+	// This is where your code goes...
+    var replyButtons = document.getElementsByClassName('access-required')
+    for (var i = 0; i < replyButtons.length; i++) {
+        replyButtons[i].addEventListener('click', saveComment, false)
+    }
+    addPreviousCommentButton()
+};
+var addPreviousCommentButton = function () {
+    // Add a button to the button list to view 
+    // the previous versions of the comment
+    var buttonsLists = document.getElementsByClassName("flat-list buttons")
+    for(var i = 0; i < buttonsLists.length; i++) {
+        var button = document.createElement("LI")
+        var link = document.createElement("a")
+        link.addEventListener("click", loadComment, false)
+        link.href = "javascript:void(0)"
+        link.innerHTML = i18n("commentSaveOnReplyButton") + "-RES"
+        button.appendChild(link)
+        buttonsLists[i].appendChild(button)
+    }
+}
+var updateButton = function (button, removed) {
+    // removed is a bool for whether or not the preview was just removed
+    // takes the <a> element when clicked
+    if (removed) {
+        button.innerHTML = i18n("commentSaveOnReplyButton") + "-RES"
+        button.removeEventListener("click", hidePreviousComment, false)
+        button.addEventListener("click", loadComment, false)
+    } else {
+        button.innerHTML = i18n("commentSaveOnReplyHideButton") + "-RES"
+        button.removeEventListener("click", loadComment, false)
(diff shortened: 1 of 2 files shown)


asoderman:
I moved my implementation to the saveComments module and made the recommended changes. I had to partially rewrite functionality so please let me know if I forgot any changes.

I probably need to implement some sort of filter for filtering autosaved comments to not clutter the users saved comments. Also a way to disable this feature.

lib/modules/saveComments.js
```diff
@@ -54,12 +54,22 @@ module.go = async () => {
 		addTabs({ onSavedPage: false });
 	}
 };
+module.options = {
+	enableCommentSaveOnReply: {
```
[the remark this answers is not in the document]
asoderman: I don't see any other `module.options` however I moved this one to the top of the file.

lib/modules/saveComments.js
```diff
@@ -39,10 +54,17 @@ const savedCommentStorage = Storage.wrapBlob('RESmodules.saveComments.savedComme
...
+
 module.beforeLoad = async () => {
 	savedCommentIDs = new Set(Object.keys(await savedCommentStorage.getAll()));
 
 	watchForThings(['comment'], addSaveLinkToComment);
+	if (module.options.enableCommentSaveOnReply.value) {
+		watchForThings(['comment'], addReplyListener);
+	}
+	if (module.options.enableInlinePreviousComment.value) {
+		watchForThings(['comment'], addPreviousCommentButton);
```
larsjohnsen: I think it would be noisy to have this button added to all comments regardless of whether they have been saved. What do you of extending the `saveElement` button to show the previous comment?

lib/modules/saveComments.js
```diff
@@ -54,6 +76,117 @@ module.go = async () => {
 	}
 };
 
+// Comment save on reply implementation
+
+function addReplyListener(thing) {
+	if (thing) {
```
larsjohnsen: This will always be truthy if invoked from `watchForThings`.

lib/modules/saveComments.js
```diff
@@ -54,6 +76,117 @@ module.go = async () => {
...
+		const replyButton = thing.element.querySelector('a.access-required[data-event-action=comment]');
```
larsjohnsen: As @andytuba mentioned, it might be more reliable to listen to new comments and check whether the user is the owner, like is done here: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/b287af15f1a7197f6d29f3acefeea776e7fcde21/lib/modules/newCommentCount.js#L183-L186

lib/modules/saveComments.js
```diff
@@ -54,6 +76,117 @@ module.go = async () => {
...
+		updateButton(this, true);
+	}
+}
+
+async function loadPreviousCommentHandler(event: Event) {
+	event.preventDefault();
+	const commentThing = Thing.from(this);
+	if (commentThing) {
+		const commentElement = commentThing.element;
+		const comment = await getComment(getId(commentThing));
```
larsjohnsen: Note that `getId` doesn't work on deleted comments. I think being able to also restore those comments would be useful.

lib/modules/saveComments.js
```diff
@@ -39,10 +54,17 @@ const savedCommentStorage = Storage.wrapBlob('RESmodules.saveComments.savedComme
 |} => { throw new Error('Saved comment not found'); });
 let savedCommentIDs;
 
+
 module.beforeLoad = async () => {
 	savedCommentIDs = new Set(Object.keys(await savedCommentStorage.getAll()));
```
larsjohnsen: Retrieving all comments replied to could become costly. One solution could be to first only retrieve the keys, and defer retrieving the data till actually needing to render the saved comment.

lib/modules/saveComments.js
asoderman: What do you think about storing the reply comment id as a property of saved comments, then in the event of a deleted comment retrieving saved comments by their child (reply) id? Just brainstorming.

lib/modules/saveComments.js
asoderman: I'm not sure how to fix this as it doesn't appear there is a method to return a `BlobWrapper` keys.

lib/modules/saveComments.js
asoderman: It appears the id does remain visible in deleted comments via `<div class="reportform report-t1_drow9l2"></div>`. I will try to extend functionality to deleted comments.

lib/modules/saveComments.js
larsjohnsen: Looks like the storage would have to be have to converted to `PrefixWrapper` in order to improve performance, but that can probably wait.

asoderman:
@andytuba Is everything good here?

PR #4656, opened 2018-01-24
hideChildComments: Fix state on posts loaded after toggling all

larsjohnsen:
Relevant issue: https://www.reddit.com/r/Enhancement/comments/7s0txi/when_i_try_to_hide_all_the_child_comments_it_only/
Tested in browser: Chrome 64


erikdesjardins:
For the peanut gallery: despite the post's title, this has never worked (tested in the latest patch of every minor back to 5.0, when you couldn't even do it manually)

PR #4657, opened 2018-01-24
stylesheets: display available customToggles

larsjohnsen:
There has been some confusion about what the `toggleName` column does, and entering the wrong text there prevents the snippet from working.

This changes the option from `text` to `enum`. The value remains the same.

image

Tested in browser: Chrome 65


andytuba:
There might even be a GitHub ticket I opened which this could close..  I know I opened a different ticket for "automatically backfill custom toggles for `toggleName` values in snippets and other options."

PR #4660, opened 2018-01-27
add an option to view profiles in the legacy format

overdodactyl:
This setting, placed in the Profile Redirect module, would allow users to view profile pages in a format similar to the old legacy version ( `/u/username/posts` and `/u/me/comments` pages both go to the new profile page).

Settings page:

setting_page

Active Communities and Trending Profiles are dependent on Legacy Userstyle. When enabled, they show the corresponding section in a user's sidebar.

Sample `u/username/comments` page:

comment_page

Sample `u/username/posts` page in night mode:

dm_posts

Tested in browser: Firefox


overdodactyl:
Now that Reddit has the option to opt out of the new profiles, should I go ahead and close this?

I guess the only reason to keep it around would be if there's interest in it and there's reason to believe Reddit will remove the option once the redesign hits. If that's the case I'll make sure to keep up-to-date with any changes (looks like there's just a few things that would need adjusting at this point)...otherwise I probably won't bother :)

erikdesjardins:
Yeah. I imagine two things could happen:

- reddit keeps old profiles for{ever, a long time}
- new profiles are converted to be like the redesign, where classnames are generated and can change every deploy, so it won't be possible to do this

I think the latter is inevitable either way, so I'm going to close this.

overdodactyl:
Sounds good! At the time there were enough requests/questions about opting out of viewing the new profiles I thought this might be an avenue worth looking into in :)

PR #4661, opened 2018-01-27
Allow option to prompt for password when switching accounts

qenya:
Adds the ability to type in the password every time you switch to a different account, instead of having to store it in RES settings (which is not terribly secure, as a note in the module description warns.)

Based on an old [feature request](https://www.reddit.com/r/Enhancement/comments/13pkfn/feature_request_prompt_for_a_password_when/).

Tested in browser: Firefox 57.0.4

### DIFF
--- a/lib/modules/accountSwitcher.js
+++ b/lib/modules/accountSwitcher.js
@@ -317,10 +317,12 @@ function findMatchingAccount(username, partialMatch) {
 async function switchTo(user) {
 	const account = findMatchingAccount(user);
 	if (!account) return;
-	const [username, password, requiresOtp] = account;
+	const [username, storedPassword, requiresOtp] = account;
 
-	// Don't await promise yet so that the OTP can be entered while logging out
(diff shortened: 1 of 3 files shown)


lib/modules/accountSwitcher.js
```diff
@@ -336,7 +336,7 @@ async function switchTo(user) {
 		url: '/api/login',
 		data: {
 			user: username,
-			passwd: password,
+			passwd: password ? password : window.prompt(i18n('accountSwitcherPasswordPrompt', username)),
```
erikdesjardins: Let's move this before `otp`, since usually passwords are requested before two-factor, and so logout can happen in the background

Quintinius: Sure thing! Is this (6aae1ae) OK?

qenya: Sure thing! Is this (6aae1ae) OK?

erikdesjardins:
(integration test failure is spurious)

PR #4662, opened 2018-01-29
Upgrade Babel to 7.0

mc10:
This shouldn't be merged until Babel 7.0 is actually released.


mc10:
Sounds good to me, thanks for the help!

PR #4671, opened 2018-02-10
Update deps

erikdesjardins:
Tested in browser: Chrome 65


erikdesjardins:
Crosspost expando failure is due to reddit changes

PR #4684, opened 2018-02-20
bin/travisDeploy: Use `set -e` to exit on error

mc10:
We probably don't need the changelog if the deploy failed.

### DIFF
--- a/build/travisDeploy.sh
+++ b/build/travisDeploy.sh
@@ -1,11 +1,8 @@
 #!/bin/bash
+set -e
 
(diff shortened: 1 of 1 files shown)


erikdesjardins:
The deploy will fail if any browser fails, and Edge often does for reasons beyond our control, so running both is the idea. (https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/8d66577462d37670670b70efa141de5c036cd2ef is what I wanted to do originally)

mc10:
Ah okay, makes sense.

PR #4694, opened 2018-03-02
WIP: Add Greasemonkey target

SirCmpwn:
This isn't done, but I'm opening it up for discussion and feedback on the overall approach.

As of the latest release, [qutebrowser](https://github.com/qutebrowser/qutebrowser) now [supports Greasemonkey userscripts](https://github.com/qutebrowser/qutebrowser/pull/3040). Several other non-mainstream web browsers support userscripts as well. I thought it would be fun to try and port RES *back* to a Greasemonkey script.

Greasemonkey support was removed back in 2014 (#714) and I'm sure you guys aren't too eager to see it return. However, since the move to webpack, supporting Greasemonkey is much easier and more maintainable. I'd be happy to field support requests on this front, or we can call it officially unsupported (and maybe express that fact in the UI somewhere?).

Status: loads and many features work. There are some TODOs littered throughout, but the main ones are:

* [ ] Make locale loading use the old approach on non-Greasemonkey targets
* [ ] Figure out why my storage shim behaves differently from the Chrome version
* [ ] Finish shimming out background actions with browser equivalents (or no-ops)
* [ ] Document the limitations of the Greasemonkey target
* [ ] Fix source maps (they're messed up when the greasemonkey header is patched in, other targets still work)
* [ ] Test everything

Usage:

```shell
npm start greasemonkey # for development
npm run build greasemonkey # for actual use
ln -s $(pwd)/dist/greasemonkey/foreground.entry.js ~/.local/share/qutebrowser/greasemonkey/res.js
```

### DIFF
--- a/greasemonkey/header.js
+++ b/greasemonkey/header.js
@@ -1,0 +1,10 @@
+// ==UserScript==
+// @name        Reddit Enhancement Suite
+// @namespace	  https://redditenhancementsuite.com/
+// @description	community-driven unofficial browser extension for reddit
+// @include http://reddit.com/*
+// @include http://*.reddit.com/*
+// @include https://reddit.com/*
+// @include https://*.reddit.com/*
+// ==/UserScript==
+
--- a/lib/core/init.js
+++ b/lib/core/init.js
@@ -1,5 +1,6 @@
 /* @flow */
 
+import '../css/res.scss';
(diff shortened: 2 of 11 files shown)


Araly:
any news on this ?

ddevault:
I could use some help with it. I don't really use Reddit anymore so this patch is just a curiosity to me.

PR #4704, opened 2018-03-08
Update go-mode hotkey filter

corylulu:
To match the other filters, will only show hotkeys that are enabled options, specific to the page, etc. Previously was showing every option on every page, despite the options having `include` set for specific pages.

~~Changed the names of all the 'Go to' hotkeys to start with "goto[Inbox|Frontpage|etc]" so it was more obvious which ones were linked with the Go-Mode function, based on it's description.~~
*Reordered the options so that all the 'Go to' hotkeys and options were grouped together at the bottom.*

Also made the popup show the friendly name.

chrome_2018-03-08_16-20-25

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -1184,8 +1184,14 @@ function navigateTo(newWindow, href) {
 
 const goModePanel = _.once(() => {
 	const goModes = filterMap(Object.entries(module.options), ([key, opt]) => {
-		if (!opt.goMode) return;
-		return [{ key, niceKeyCode: niceKeyCode(opt.value) }];
+		if (option.goMode && option.type === 'keycode' &&
+			option.callback &&
+			(!option.dependsOn || option.dependsOn(module.options)) &&
+			(!option.include || matchesPageLocation(option.include)) &&
+			(!option.mustBeLoggedIn || loggedInUser()) &&
+			(!option.requiresModule || Modules.isRunning(option.requiresModule))) {
+			return [{ key, niceKeyCode: niceKeyCode(option.value) }];
+		}
 	});
 
 	return $(string.html`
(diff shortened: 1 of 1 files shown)


lib/modules/keyboardNav.js
```diff
@@ -663,39 +663,39 @@ module.options = {
 		dependsOn: options => options.useGoMode.value,
 		callback() { if (module.options.useGoMode.value) toggleGoMode(); },
 	},
-	inbox: {
+	gotoInbox: {
```
andytuba: If you change option keys, you also need to add a migration to move the value from the old key to the new key.

andytuba: It may not be worth changing them in this PR.

corylulu: Okay, I'll revert that part.

corylulu: Any other ideas on how to best denote which options are go-mode enabled? Should I just group them better, all at the bottom of the list?

corylulu:
Okay, I instead grouped all the options together and at the bottom.

chrome_2018-03-08_16-34-19

lib/modules/keyboardNav.js
```diff
@@ -1183,9 +1183,15 @@ function navigateTo(newWindow, href) {
 }
 
 const goModePanel = _.once(() => {
-	const goModes = filterMap(Object.entries(module.options), ([key, opt]) => {
-		if (!opt.goMode) return;
-		return [{ key, niceKeyCode: niceKeyCode(opt.value) }];
+	const goModes = filterMap(Object.entries(module.options), ([, option]) => {
```
larsjohnsen: Since filtering for valid shortcuts is already performed in `getActiveCommandOptions`, this may be simplified to 
```
getActiveCommandOptions()
  .filter(opt => opt.goMode)
  .map(opt => ({ key: i18n(opt.title), niceKeyCode: niceKeyCode(opt.value) }))
```

corylulu: This did not work before. Things like Legacy Profile and New Profile links showed on every page, despite only functioning on select pages.

PR #4705, opened 2018-03-09
Reordered test actions to avoid issues with built-in userinfo

corylulu:
I can't reproduce the issue myself, but I think, like the [comment describes](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/tests/selectedEntry.js#L14), the reddit built-in usertip might be overlaying the `loadMoreChildrenOfSticky`. So I reordered so they'd work top to bottom to hopefully avoid that issue.


corylulu:
Yay! That worked :)

mc10:
Seems reasonable to me.

Issue #3932, opened 2017-01-25
Render keynav help off-screen and slide it in


andytuba:
instead of positioning in in the top right and animating width/height.  fewer dom reflows.

PR #4706, opened 2018-03-09
Render keynav help off-screen and slide it in

jeromew21:
Relevant issue: https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/3932 DOM reflows when height and width are animated
Tested in browser: Chrome 64

Replaced jquery call to `.toggle(300);` with a custom slide animation.


erikdesjardins:
Thanks!

PR #4707, opened 2018-03-09
Use gfycat's new iframe embed

nickretallack:


larsjohnsen:
Are there any advantages to embedding this instead of continuing to use RES' video player?  By keeping RES responsible for generating the video element, we retain fluid resizing, preloading when using media browse, and generally much faster load times.

PR #4713, opened 2018-03-16
[RFC] Change selected entry from background colour to right border.

BenMcGarry:
As discussed between @honestbleeps and myself on IRC. It might be worth swapping selected entry from background colour to a right border for the following reasons:

1) It looks cleaner.
2) Would resolve keynav issues on subreddits where they are intentionally styling it out.

The proposal in commit looks like this: 

Thoughts?

### DIFF
--- a/lib/modules/selectedEntry.js
+++ b/lib/modules/selectedEntry.js
@@ -59,7 +59,7 @@ module.options = {
 		value: true,
 		description: 'selectedEntrySetColorsDesc',
 	},
-	backgroundColor: {
+	borderColor: {
 		title: 'selectedEntryBackgroundColorTitle',
 		type: 'color',
 		value: '#F0F3FC',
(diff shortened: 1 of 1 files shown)


larsjohnsen:
Note that `addLine` already implements a line on the right side
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/24edda029d651acd473f521599d639da04ea2f0a/lib/modules/selectedEntry.js#L336-L339
using `box-shadow` instead of `border` as to not shrink the context box.

It should suffice to migrate the value of `setColors` to `addLine`, and then disable `setColors`.

honestbleeps:
does `box-shadow` make more sense here than `outline`, actually?

larsjohnsen:
IIRC `outline` cannot be drawn on only one side of the element.

honestbleeps:
whoops, my bad, you're right! sorry about that

mc10:
Hmm I'm not sure if I'm in favor of this; I just tried out `addLine` and it's a *lot* more subtle than `setColors`, especially on a wide screen, where you have to glance across the screen to see the indicator. If we migrate this for everyone, there will likely be a lot of complaints.

PR #4714, opened 2018-03-20
flow: Update to 0.68.0, fix new issues

mc10:
Not all Flow issues have been fixed yet; I'm still working through them.


lib/modules/hosts/imgur.js
```diff
@@ -157,44 +224,44 @@ export default new Host('imgur', {
...
 		}
 	},
-	async handleLink(href, getInfo) {
+	async handleLink(href, getInfo: () => Image | Promise<Image>) {
 		const baseUrl = 'https://imgur.com/';
 		const shareLinkPreferred = this.options.preferredImgurLink.value === 'share';
 		const resolutionSuffix = this.options.imgurImageResolution.value;
 		const useGif = this.options.useGifOverGifv.value;
 
 		const info = await getInfo();
```
mc10: @erikdesjardins I don't really understand Flow; is there a way to force the pattern matching after this line to work? I tried adding an `objType` field but it seems like it's not enough for Flow to distinguish between the different `Image` types.

erikdesjardins: Ah, sorry, I had pushed to this branch and fixed it earlier. I should have merged it right away.

I'd have to investigate to see why it doesn't work now, and I don't have much time right now unfortunately

Issue #2731, opened 2016-02-22
gomode should leave noparticipation


andytuba:

bartnic1:
Hello, I'm new to making contributions and wanted to take this issue on as for my first pull request; with your permission I'd like to work on it. Is there anything I should know as a new contributor? I've read the contributing.md document and I have experience with github.

Thanks!

andytuba:
Go for it!

Make sure to run linting and functional tests, or at least check for a red X / green checkmark on your PR a few minutes after submitting it.

andytuba:
(it may help to run the automatic lint fix before pushing)

PR #4724, opened 2018-04-02
Fixed issue #2731; go mode will now escape NP mode

bartnic1:
Relevant issue: #2731

Resubmitting pull request to see whether timeout error on previous test will repeat itself.

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -1021,6 +1021,7 @@ async function handleCallbacks(options) {
 			console.error('Could not execute callback', i18n(option.title), e);
 		}
 	}
+	npEscapeActive = goModeActive;
 }
 
 const promptUniqueShortcut = _.memoize(async (shortcut, options: any) => {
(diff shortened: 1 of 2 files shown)


lib/modules/keyboardNav.js
```diff
@@ -1194,7 +1195,17 @@ function reply(selected = getSelected()) {
 	));
 }
 
+let npEscapeActive = false;
```
mc10: Not entirely sure why this variable is necessary; can't you just use `goModeActive` directly? Also, I think it's worth considering not checking for `goMode` at all, if it would always make sense to escape NP mode.

lib/modules/noParticipation.js
```diff
@@ -108,8 +108,8 @@ function isSubscriber() {
 	return document.body.classList.contains('subscriber');
 }
 
-function nonNpLocation() {
-	const nonNpUrl = new URL(location.href);
+export function nonNpLocation(path: ?string) {
```
mc10: This can be `path: string | null`, which is slightly stronger.

bartnic1:
Not sure what's causing this timeout issue, I'll test it a bit more later today.

lib/modules/keyboardNav.js
bartnic1: Initially the reason was that goModeActive was being set to false before navigateTo() was being called in handleCallbacks(), so I made a more delayed boolean (it switches to false only after all callbacks have been called). I could have changed the order that things were handled in handleCallbacks(), but I wanted to avoid causing any unintentional side-effects. 

The only thing with always escaping NP mode is that, if the user changes their mind and exits go mode (or was never in go mode), I need some check to ensure that NP mode isn't being escaped, because navigateTo() is also being used in other non-go-mode functions.

lib/modules/noParticipation.js
bartnic1: Thank you! I'll make the change

bartnic1:
I'll have to come back to this a bit later. I'll see if I can get everything working locally before attempting another pull request

PR #4726, opened 2018-04-03
Capability to set default sort and time options sidebar search

a-marquez:
Clean and simple.

Relevant issue: #4725
Tested in browser: chrome


a-marquez:
I have rebased this to the latest from upstream, is is no longer failing (I believe it was a false negative last time). Could I have someone please review this or point me to the next step?

larsjohnsen:
Thanks!

PR #4734, opened 2018-04-09
Use toggle interface for nightMode and NSFW mode

larsjohnsen:
- NSFW mode applies to all active tabs when toggled
- `nsfw` command no longer needs to have `on` or `off` specified
- `ls` command for toggling nightMode is removed

Tested in browser: Chrome 65


andytuba:
Thanks for this refactor! That'll make it even easier to add all the menu toggles to the redesign.

PR #4740, opened 2018-04-11
keyboardNav linkNumbers: Add option to change/reject altMode modifier

### LINKED ISSUE #4251: keyboardNav linkNumbers breaks Alt-number tab switching in Chromium
- My browser: Chromium "Version 58.0.3029.110 Built on Ubuntu , running on Ubuntu 14.04 (64-bit)"
- My RES version: v5.6.2
- Am I in reddit beta: no

*What's the problem?*
When linkNumbers in keyboardNav is set to "On", switching tabs via Alt-1..9 doesn't work anymore.

To fix this behavior for users who want to use linkNumbers and user Alt-1..9 for tab switching, add an option to disable [linkNumbers's Alt shorcuts](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/v5.6.2/lib/modules/keyboardNav.js#L884-L887).

Other possible solutions:
- Allow Shift as an alternative modifier to Alt. May conflict with existing shortcuts though, depending on the keyboard layout.
- Only call [event.preventDefault()](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/v5.6.2/lib/modules/keyboardNav.js#L943) when a link with the pressed number key exists when pressing Alt-0..9.

Note that both alternative solutions listed above don't completely fix the use case of tab switching, thus disabling Alt-1..9 shortcuts should be implemented either way.

https://www.reddit.com/r/RESissues/comments/53pusg/bug_cannot_switch_tabs_directly_in_using_alt_keys/

larsjohnsen:
Option available under http://reddit.com/#res:settings/keyboardNav/linkNumberAltModeModifier

Relevant issue: Closes #4251
Tested in browser:  Chrome 65

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -128,6 +128,24 @@ module.options = {
 		description: 'keyboardNavLinkToggleExpandoDesc',
 		title: 'keyboardNavLinkToggleExpandoTitle',
 	},
+	linkNumberAltModeModifier: {
+		dependsOn: options => options.linkNumbers.value,
+		type: 'enum',
+		values: [{
+			name: 'none',
+			value: 'none',
+		}, {
+			name: 'Shift',
+			value: 'shift',
+		}, {
+			name: 'Alt',
+			value: 'alt',
+		}],
+		value: 'alt',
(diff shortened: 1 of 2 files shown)


erikdesjardins:
CI failure is spurious

PR #4744, opened 2018-04-17
profileRedirect: use location.replace to avoid continuous redirect

mudkip908:
Currently pressing the "Back" button after being redirected by profileRedirect will result in going back to the same page. Using location.replace prevents this behavior.


erikdesjardins:
Thanks!

PR #4749, opened 2018-04-21
tests: Fix type checking issues

mc10:
This splits the type-checking changes off from #4662.


mc10:
Guess the AVA changes depend on [1.0.0-beta.2](https://github.com/avajs/ava/releases/tag/v1.0.0-beta.2), so this PR might not be separate-able if they don't release a non-1.0.0 release.

PR #4753, opened 2018-04-29
nightmode: Fix automatic night mode handling from outdated tabs

larsjohnsen:
Fixes regression from #4730

Tested in browser: Chrome 66


erikdesjardins:
Test failure is spurious

PR #4759, opened 2018-05-04
Restore manual resize events for Firefox

larsjohnsen:
This causes some performance loss, but is evidently necessary.

Relevant issue: https://www.reddit.com/r/RESissues/comments/8gtfpz/images_overlaps_in_latest_res_update/  https://www.reddit.com/r/Enhancement/comments/8gljh7/firefox_5903_images_overlap_in_inline_viewer/
Tested in browser: Firefox 59, 61

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1401,6 +1401,8 @@ class Image extends Media {
 			});
 		}
 
+		waitForEvent(this.image, 'load').then(() => triggerResizeEvent(this.image));
+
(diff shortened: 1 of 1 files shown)


larsjohnsen:
No, I am not able to reproduce unless I enable `disableAnimations`. 

The new polyfill layer uses strategy `scroll` in order to detect resizes. The `object` strategy, which we used earlier, seems to work fine with `disableAnimations` enabled, however.

PR #4761, opened 2018-05-04
Use resize-observer-lite fork which uses different strategy

larsjohnsen:
The `scroll` strategy which was used does not work when animations are globally disabled, causing media to not flow properly. The other available strategy, `object`, does not have this issue. 

Replaces #4759 

@erikdesjardins Can you fork the repo [larsjohnsen/resize-observer-lite](https://github.com/larsjohnsen/resize-observer-lite) to the organization, like is done with jquery.tokeninput?

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "lodash": "4.17.5",
     "moment": "2.22.1",
     "numeral": "2.0.6",
-    "resize-observer-lite": "0.2.3",
+    "resize-observer-lite": "Reddit-Enhancement-Suite/resize-observer-lite#59af4a859cb7da7859fec1f6bceecf26dd278916",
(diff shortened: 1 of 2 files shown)


erikdesjardins:
Done. Everyone in the org should have write access to https://github.com/Reddit-Enhancement-Suite/resize-observer-lite

Another potential option is to change the disableAnimations CSS to avoid applying to the relevant media elements; maybe you've already considered that and discarded it?

larsjohnsen:
Thanks.

I haven't really explored adding an expectation to the disableAnimations CSS as I am satisfied with just switching the resize detection strategy, as there doesn't appear to be any significant performance difference between the two.

erikdesjardins:
Sounds good to me.

PR #4763, opened 2018-05-07
Fix account switcher on redesign

andytuba:
1. fixed logged-in check prerequisite for logout
1. fetch modhash as prerequisite for logout

tested in chrome

accountswitcher-d2x
accountswitcher-r2

### DIFF
--- a/lib/environment/foreground/ajax.js
+++ b/lib/environment/foreground/ajax.js
@@ -58,6 +58,14 @@ export async function ajax(options: AjaxOptions<*>) {
 		}
 	}
 
+	if (sameOrigin) {
+		if (method !== 'GET' && method !== 'HEAD') {
+			// Send modhash for same-origin non-GET requests
+			const hash = await loggedInUserHash();
+			if (hash) headers['X-Modhash'] = hash;
+		}
+	}
+
 	const response = await (sameOrigin ?
 		fetch(url, {
 			method,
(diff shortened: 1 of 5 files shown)


lib/environment/foreground/ajax.js
```diff
@@ -108,12 +116,6 @@ function buildRequestParams({ method = 'GET', url, query = {}, headers = {}, dat
 	if (sameOrigin) {
 		// Add `app=res` to same-origin request URLs
 		urlObj.searchParams.set('app', 'res');
-
-		if (method !== 'GET' && method !== 'HEAD') {
-			// Send modhash for same-origin non-GET requests
-			const hash = loggedInUserHash();
-			if (hash) headers['X-Modhash'] = hash;
-		}
```
andytuba: This could have stayed in `buildRequestParams` if I made it async. too. Think I should move it back?

lib/utils/user.js
```diff
@@ -2,27 +2,62 @@
...
+		if (!document.querySelector('#header a[href*="/login"]')) {
+			return true;
+		}
+	}
+};
+
 export const loggedInUser = _.once((): string | void => documentLoggedInUser(document));
 
 export const documentLoggedInUser = (document: Document | Element): string | void => {
+	const meta = document.querySelector('meta[name="jsapi:username"]');
```
andytuba: Doesn't exist yet. Hopefully it will soon.

lib/environment/foreground/ajax.js
```diff
@@ -58,6 +58,14 @@ export async function ajax(options: AjaxOptions<*>) {
 		}
 	}
 
+	if (sameOrigin) {
+		if (method !== 'GET' && method !== 'HEAD') {
+			// Send modhash for same-origin non-GET requests
+			const hash = await loggedInUserHash();
```
andytuba: This block could have stayed in `buildRequestParams` if I made that function async. too. 

However, since `ajax.invalidate` works better with a synchronous function and doesn't care about the modhash header, it seems fine to move it into the main ajax function.

PR #4768, opened 2018-05-14
filteReddit support for reddit redesign

andytuba:
Relevant issue: 
Tested in browser:


lib/modules/filteReddit/postCases/Score.js
```diff
@@ -21,6 +21,9 @@ export class Score extends Case {
 
 	evaluate(thing: *) {
 		const score = thing.getScore();
-		return isNaN(score) ? false : numericalCompare(this.value.op, score, this.value.val);
+		if (typeof score !== "number" || isNaN(score)) {
```
andytuba: yet somehow everything gets filtered anyway. filtering has gotten beyond the grasp of my feeble pres-RES-5 mind.

larsjohnsen: Note that even if this returns `false` the post thing is hidden _when_ the filter's state is inverted (i.e. only display posts which returns `true`). https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/9b4a2fa91d42e6b4b7fed50da6dddb6ba7f6028c/lib/modules/filteReddit/Filter.js#L133-L134 `Case#evaluate` should probably just throw an error instead when it's not able to evaluate a post.

Another way to solve this could be to just disable cases until Thing is built out.
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/9b4a2fa91d42e6b4b7fed50da6dddb6ba7f6028c/lib/modules/filteReddit/postCases/IsVisited.js#L12-L14

andytuba:
In between @larsjohnsen's very considerate upgrades to FilteReddit, particularly for "handling data not available" and me learning how to handle [disjoint unions with exact types in Flow](https://flow.org/en/docs/types/unions/#toc-disjoint-unions-with-exact-types), this might actually be ready for testing.

I added an escape hatch option to disable filtering in redesign. I was considering adding a migration to turn that on if the user is on a beta channel.. maybe I could do that by parsing current version?

Mathnerd314:
Has this been abandoned or is there a non-conflicted version somewhere?

larsjohnsen:
@Mathnerd314  I pushed a rebase, though I have not done any further testing of this.

Mathnerd314:
So the UI shows up and works, and it hides/shows things when you invert the filters. But all the filters match nothing, even the score filter. And it also doesn't seem to handle the redesign's infinite scroll at all, in that hiding hides only the first page of links and the rest are still visible.

I see in the console that the redesign is making requests to `https://gateway.reddit.com/desktopapi` and the responses have useful JSON data containing the posts. Is RES using that JSON or working backwards from the DOM?

andytuba:
[`watchForRedditEvents`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4768/files#diff-82842b227f37a48ba2525c1d2a7e9875R532) should trigger "examine and maybe filter this post/comment, both on initial pageload, SPA navigation to a new post listing, and infinite scroll. (RES doesn't examine the API responses initiated by the Reddit first-party code).

andytuba:
If the filters don't match something, that's a fault somewhere in [`Thing.js`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4768/files#diff-bb572568d990225b26b143267b2777c3) -- either the filter doesn't have redesign support, or it's buggy.

blackwood:
Hi, I'm potentially looking to contribute to this pull request, if this is still the most relevant/active PR related to enabling the filters for the new design. Let me know—thanks.

PR #4771, opened 2018-05-16
Reddit Link Bar for "Original Content" w.r.t. current interface redesign

prasoc:
Relevant issue: 
Tested in browser: 

Looks like this:

image

We need to create a translation of "Original Content" for other languages apart from English. This is under the new menu option `subredditManagerLinkOriginalTitle` and `subredditManagerLinkOriginalDesc`

### DIFF
--- a/lib/modules/styleTweaks.js
+++ b/lib/modules/styleTweaks.js
@@ -190,6 +190,11 @@ module.beforeLoad = () => {
 				background: ${highlightTopLevelColor} !important;
 			}
 		`);
+		addCSS(`
+			.Comment.top-level {
+				border-top: 2px solid ${highlightTopLevelColor};
+			}
+		`);
 	}
 };
 


lib/modules/subredditManager.js
```diff
@@ -821,217 +828,48 @@ function redrawSubredditBar() {
...
+	const originalAll = originalShortcuts.find(({ pathname }) => pathname === '/r/all');
+	const originalPopular = originalShortcuts.find(({ pathname }) => pathname === '/r/popular');
 	const originalProfilePosts = originalShortcuts.find(({ pathname }) => pathname === '/r/profileposts/');
+	const originalOC = originalShortcuts.find(({ pathname }) => pathname === '/original');
 
-	const headerContents = document.querySelector('#sr-header-area');
-	if (headerContents) {
-		// Clear out the existing stuff in the top bar first, we'll replace it with our own stuff.
-		$(headerContents).html('');
+	$("#hamburgers").show();
```
andytuba: I appreciate the effort you put into this! but it'll make for kinda weird UX and might be fragile. I'm hoping to get JSAPI integration for the hamburger menu set up soon that will make this much simpler.

locales/locales/en@lolcat.json
```diff
@@ -5818,5 +5818,13 @@
     "commentNavigatorShowByDefaultTitle": {
         "message": "Show By Default",
         "description": ""
+    },
```
andytuba: Thanks for due diligence, but you only need to edit `en.json`. All the other languages will copy `en.json`'s values into they get translations hooked up.

lib/css/res.scss
```diff
@@ -1725,3 +1725,48 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
...
 	top: 30px !important;
 }
+
+#hamburgers {
+	display: none;
+}
+
+#hamburgers .header-txt {
+	padding: 0px 16px;
+	font-size: 10px;
```
andytuba: nit: please be consistent with tabs vs spaces

lib/modules/subredditManager.js
andytuba: If you have a free moment, it would be nice to make a separate PR with just the `originalOC` changes, that could get merged sooner!

prasoc: OK I will split the `originalOC` code and pull it in

prasoc:
I have minimised the pull request to only have functionality for the "Original Content" nav item (incl. translations which needs to be implemented)

PR #4772, opened 2018-05-17
Remove Edge KeyboardEvent.key polyfill

BenMcGarry:
Issue appears to have been resolved in 1803 release. https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8860571/

http://www.movable-type.co.uk/dev/keyboardevent-key-values.html returns correct values under Edge.

Tests appear good.


BenMcGarry:
Tests appear good.

PR #4774, opened 2018-05-17
Remove Edge isIntersecting polyfill

BenMcGarry:
References: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12156111/

Appears to correctly function in Edge 17 per the following test, implemented in Edge 16 (polyfilled in 15): http://bl.ocks.org/nolanlawson/raw/bf31e883ca28579bc063c97fa5d1882b/

### DIFF
--- a/edge/edge.entry.js
+++ b/edge/edge.entry.js
@@ -79,6 +79,8 @@ Element.prototype.replaceWith = function replaceWith(...nodes) {
 	}
 };
 
+
+// https://developer.microsoft.com/en-us/microsoft-edge/platform/status/requestidlecallback/
 if (typeof requestIdleCallback === 'undefined') {
 	window.requestIdleCallback = fn => requestAnimationFrame(() => { requestAnimationFrame(fn); });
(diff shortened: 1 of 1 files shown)


edge/edge.entry.js
```diff
@@ -79,20 +79,6 @@ Element.prototype.replaceWith = function replaceWith(...nodes) {
...
-if (!IntersectionObserverEntry.prototype.hasOwnProperty('isIntersecting')) {
-	Object.defineProperty(IntersectionObserverEntry.prototype,
-		'isIntersecting', {
-			get() {
-				return this.intersectionRatio > 0;
-			},
-		});
-}
-
-if (typeof requestIdleCallback === 'undefined') {
```
larsjohnsen: `requestIdleCallback` is a separate polyfill and AFAICS not implemented in Edge yet.

BenMcGarry: Thanks, i've put it back in.

PR #4782, opened 2018-05-30
hover: Support fixed positioning

larsjohnsen:
Mainly fixes a issue in the redesign's lightboxes where `position: absolute` cannot be used to keep the hover container relative to a given target. 

- If a target parent has `position: fixed`, also set `position: fixed` on the hover container
- Closes at once if target is in a lightbox which is scrolled, as to prevent position mismatch with the target

Relevant issue: 
Tested in browser: Chrome 60

### DIFF
--- a/lib/modules/hover.js
+++ b/lib/modules/hover.js
@@ -63,10 +63,13 @@ module.options = {
 };
 
 module.beforeLoad = () => {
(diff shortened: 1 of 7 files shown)


erikdesjardins:
FF integration tests expected to fail, see #4830

PR #4784, opened 2018-05-31
Update deps

erikdesjardins:
Tested in browser: Chrome 60


erikdesjardins:
Chrome tests pass, Firefox tests are broken right now

PR #4795, opened 2018-06-15
Fix tests: try clicking on body of comment to avoid reddit popups

erikdesjardins:


erikdesjardins:
The current Firefox dev version in Sauce Labs seems broken (unable to select elements that appear to clearly exist, like `.res-show-images`, but at least Chrome passes now

PR #4796, opened 2018-06-15
messaging tests: Simplify nested promises

mc10:
I think this can be safely done, as `Promise.resolve().then()` should just return the internal `Promise` and resolve to whatever that `Promise` resolves to.


erikdesjardins:
I originally did this because timing mattered for interceptors (since `new Promise(...)` calls its argument synchronously but `Promise.resolve().then(...)` runs in the next tick) but now that that functionality has been removed it doesn't make a difference

PR #4799, opened 2018-06-18
showImages: Always display native controls on non-muted videos

larsjohnsen:
Makes the behavior of Imgur videos with sound the same as Streamable's.

Relevant issues: https://www.reddit.com/r/Enhancement/comments/8rxnck/show_imgur_audio_controls_on_hover/ 
Tested in browser: Chrome 68


erikdesjardins:
I think we might as well rebase and hotfix this

PR #4805, opened 2018-06-29
"Night Mode" for redesign (with backward compat)

### LINKED ISSUE #4927: Enable RES nightmode when redesign nightmode enabled

j-blandford:
Alters RES's "Night Mode" module. This change gives dark styling to the extra interfaces that are added by RES, such as the settings menu and the subreddit info dialogs. 

It also gets rid of the gear menu's "night mode" toggle: now the toggle is delegatated to the official Reddit toggle.

It keeps everything the same on the old redesign: the cog menu is untouched, and the styling for buttons is kept, the only change is for the new design.

Fixes https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/4927

### DIFF
--- a/lib/modules/nightMode.js
+++ b/lib/modules/nightMode.js
@@ -13,6 +13,7 @@ import {
 	currentSubreddit,
 	HOUR,
 	MINUTE,
+	isAppType,
 } from '../utils';
 import { Session, Storage, i18n } from '../environment';
 import { RedditUserCookie } from '../types/reddit';
@@ -111,7 +112,7 @@ module.onToggle = enabled => {
 module.loadDynamicOptions = () => {
 	// Old design needs to hook into the document object as early as possible. this is not needed
 	// in the redesign, as it's initialized within the Reddit React application as soon as page loads
-	if (className() === 'res-nightmode') {
+	if (isAppType("r2")) {
 		// To avoid the flash of unstyled content, the very first thing we should do is get a hold
 		// of the document object and add necessary classes...
 		if (typeof localStorage === 'object' && localStorage.getItem(localStorageKey)) {
@@ -128,7 +129,7 @@ module.beforeLoad = () => {
 	}
 
 	// Add the night mode toggle in the gear menu *only for the old design*
-	if (className() === 'res-nightmode') {
+	if (isAppType("r2")) {
 		toggle = new CustomToggles.Toggle('nightMode', i18n('nightModeToggleText'), module.options.nightModeOn.value);
 		toggle.onToggle(type => {
 			if (type === 'manual') overrideNightMode();
@@ -165,7 +166,7 @@ export function isNightModeOn(): boolean {
 
 	// `toggle` contains the current nightMode state
 	// `userCookie` contains the decoded "USER" cookie data
-	return (toggle ? toggle.enabled : module.options.nightModeOn.value) || (userCookie.prefs.nightmode && className() === 'res-d2x-nightmode');
+	return (toggle ? toggle.enabled : module.options.nightModeOn.value) || (userCookie.prefs.nightmode && isAppType("d2x"));
 }
 
 export function getUserCookie(): RedditUserCookie {
--- a/lib/modules/nightMode.js
+++ b/lib/modules/nightMode.js
@@ -15,9 +15,9 @@ import {
 	MINUTE,
 } from '../utils';
 import { Session, Storage, i18n } from '../environment';
+import { RedditUserCookie } from '../types/reddit';
 import * as CustomToggles from './customToggles';
 import * as StyleTweaks from './styleTweaks';
-import { RedditUserCookie } from '../types/reddit';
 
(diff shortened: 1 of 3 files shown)


lib/modules/nightMode.js
```diff
@@ -121,17 +127,20 @@ module.beforeLoad = () => {
...
-		Options.set(module, 'nightModeOn', toggle.enabled);
-	});
-	toggle.onStateChange(() => {
-		if (toggle.enabled) addNightMode();
-		else removeNightMode();
-		StyleTweaks.toggleSubredditStyleIfNecessary();
-	});
-	toggle.addCLI('ns');
+	// Add the night mode toggle in the gear menu *only for the old design*
+	if (className() === 'res-nightmode') {
```
erikdesjardins: Use `isAppType('d2x')` (desktop2x, the codename of the redesign...maybe this could be more intuitive) or `isAppType('r2')` instead of comparing the classname

lib/modules/nightMode.js
```diff
@@ -144,15 +153,26 @@ module.always = () => {
...
+
+	const userCookie = getUserCookie();
+
 	// `toggle` contains the current nightMode state
-	return toggle ? toggle.enabled : module.options.nightModeOn.value;
+	// `userCookie` contains the decoded "USER" cookie data
+	return (toggle ? toggle.enabled : module.options.nightModeOn.value) || (userCookie.prefs.nightmode && className() === 'res-d2x-nightmode');
+}
+
+export function getUserCookie(): RedditUserCookie {
```
erikdesjardins: There might be a better way to do this via the jsapi, @andytuba?

andytuba: Jsapi doesn't have the cookie, but that's a good idea. I'll make a note for next week.

lib/modules/nightMode.js
prasoc: Fixed. It does look cleaner than using a triple-equality comparison

lib/modules/nightMode.js
prasoc: What is the "jsapi"? Where is it located? I cannot find anything online related to this term

andytuba: https://www.reddit.com/r/redesign/wiki/jsapi

andytuba: I'd like to come back to "get redesign nightmode pref" later, but this should hold for now.

lib/modules/nightMode.js
```diff
@@ -144,15 +154,26 @@ module.always = () => {
...
+	const userCookie = getUserCookie();
+
 	// `toggle` contains the current nightMode state
-	return toggle ? toggle.enabled : module.options.nightModeOn.value;
+	// `userCookie` contains the decoded "USER" cookie data
+	return (toggle ? toggle.enabled : module.options.nightModeOn.value) || (userCookie.prefs.nightmode && isAppType("d2x"));
+}
+
+export function getUserCookie(): RedditUserCookie {
+	const redditCookie = document.cookie.split('; ').filter(((item: string) => item.includes(redditNightModeCookie + '='))).toString();
```
andytuba: It would be nice to refactors this to a function like

`lib/user.js`
```javascript
function getPrefsFromCookie() {
    return JSON.parse(atob(cookies.get('USER') || ''));
}
```

lib/modules/nightMode.js
```diff
@@ -144,15 +154,26 @@ module.always = () => {
...
 
 export function isNightModeOn(): boolean {
 	if (!Modules.isRunning(module)) return false;
+
+	const userCookie = getUserCookie();
+
 	// `toggle` contains the current nightMode state
-	return toggle ? toggle.enabled : module.options.nightModeOn.value;
+	// `userCookie` contains the decoded "USER" cookie data
+	return (toggle ? toggle.enabled : module.options.nightModeOn.value) || (userCookie.prefs.nightmode && isAppType("d2x"));
```
andytuba: `(userCookie && userCookie.prefs && userCookie.prefs.nightmode)` -- not guaranteed to have the `USER` cookie, especially before checking if `isAppType('d2x')`

lib/modules/nightMode.js
```diff
@@ -121,17 +128,20 @@ module.beforeLoad = () => {
...
+			if (type === 'manual') overrideNightMode();
+			Options.set(module, 'nightModeOn', toggle.enabled);
+		});
+		toggle.onStateChange(() => {
+			if (toggle.enabled) addNightMode();
+			else removeNightMode();
+			StyleTweaks.toggleSubredditStyleIfNecessary();
+		});
+		toggle.addCLI('ns');
+	}
```
andytuba: FUTURE: Add a "on nightmode toggled" watcher for the redesign. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/onChanged

lib/modules/nightMode.js
kenny10291977: H

PR #4814, opened 2018-08-01
switch to dart-sass, dropping python dependency

erikdesjardins:


mc10:
Would it be useful to [add the fibers import](https://github.com/webpack-contrib/sass-loader#examples) as recommended by the docs?

erikdesjardins:
That package requires a native module -> node-gyp -> Python 2, which I want to avoid.

PR #4816, opened 2018-08-04
showImages: Downloading an image/GIF sets filename to post title

kdreagle:
Relevant issue: N/A
Tested in browser: Chrome 68

Suggested by [/u/dracho](https://www.reddit.com/user/dracho) in [this post](https://www.reddit.com/r/RESissues/comments/94chjf/please_make_the_save_image_button_save_the_image/). 

This is my first time contributing to open source, so excuse any mistakes. I wasn't sure if there should've been a module option for this, but I don't see any reason to have this feature turned off.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -1086,7 +1086,6 @@ const retrieveExpandoOptions = _.memoize(async (thing, { siteModule, detectResul
 		].filter(v => v).map(s => s.toLowerCase()),
 		buttonInfo: getMediaButtonInfo(mediaOptions),
 		generateMedia() {
-			if (thing) mediaOptions.title = thing.getTitle();
 			const media = generateMedia(mediaOptions, { href });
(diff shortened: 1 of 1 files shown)


lib/modules/showImages.js
```diff
@@ -1582,7 +1583,7 @@ function trackMediaLoad(link, thing) {
 	}
 }
 
-function setMediaControls(media, lookupUrl, downloadUrl) {
+function setMediaControls(media, lookupUrl, downloadUrl, title) {
```
larsjohnsen: I'd rather avoid propagate `title` as a `Media` instance can be reused by other expandos. It would be easier to retrieve `thing` by `Thing.from(media)` along with the other logic below.

kdreagle: Oy vey, I should've check out the Thing class first. Thanks! Apparently that was actually causing the flow errors.

erikdesjardins:
Thanks!

PR #4820, opened 2018-08-14
Fixes error for when `customFiltersP/C` is null

corylulu:
This has been throwing errors for me when `customFiltersP` or `customFiltersC` is null, so I added an additional check. 

Relevant issue: Unknown
Tested in browser: Yes


larsjohnsen:
Thanks!

PR #4828, opened 2018-08-26
Fix account switcher and other ajax requests to Reddit API from new.reddit.com

andytuba:
new.reddit.com always routes to the new reddit web desktop service (desktop2x). Unlike www.reddit.com, api (.json) requests are not routed to the reddit api public service. Thus, sending ajax requests to, e.g. `/api/login` from new.reddit.com will 404.  To fix this, when sending an ajax request to the reddit api from new.reddit.com, that request will now be sent to www.reddit.com and sent via the background page, including credentials.

Tested on Chrome 68 and Firefox 61. Was not tested with Firefox's third-party cookies option disabled.

nb: new.reddit.com/....json?embedded=true currently selects r2, i.e. the public reddit API. (This is currently leveraged by first-party ajax api requests from new.reddit.com/message/inbox). However, I decided against it for RES as that routing may be deprecated unexpectedly.

edit(erikdesjardins): closes #4819

### DIFF
--- a/lib/environment/foreground/ajax.js
+++ b/lib/environment/foreground/ajax.js
@@ -119,6 +119,9 @@ function buildRequestParams({ method = 'GET', url, query = {}, headers = {}, dat
 	}
 	let sameOrigin = sameSite;
 	if (sameSite && urlObj.hostname.startsWith('new.')) {
+		// new.reddit.com specifically selects the reddit desktop web frontend service.
+		// AJAX requests should be sent to www.reddit.com, which will select the correct
+		// service based on the path.
 		sameOrigin = false;
 		urlObj.hostname = urlObj.hostname.replace(/new\./, 'www.');
 	}
--- a/lib/environment/foreground/ajax.js
+++ b/lib/environment/foreground/ajax.js
@@ -49,7 +49,7 @@ declare function ajax(opt: AjaxOptions<'raw'>): Promise<{
  * `cacheFor` is a TTL, in milliseconds.
  */
 export async function ajax(options: AjaxOptions<*>) {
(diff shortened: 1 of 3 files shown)


lib/environment/foreground/ajax.js
```diff
@@ -112,11 +112,16 @@ function buildRequestParams({ method = 'GET', url, query = {}, headers = {}, dat
 		urlObj.searchParams.set(key, String(val));
 	}
 
-	const sameOrigin = urlObj.hostname.includes(location.hostname.split('.').slice(-2).join('.'));
-	if (sameOrigin) {
+	const sameSite = urlObj.hostname.includes(location.hostname.split('.').slice(-2).join('.'));
```
andytuba: With that `.slice(-2)` here, this may as well read `urlObj.hostname.endsWith('.reddit.com')`. Maybe pull the `'.reddit.com'` magic value from the manifest.

lib/utils/user.js
```diff
@@ -39,8 +39,7 @@ export const loggedInUserHash = _.once(async (): Promise<?string> => {
 });
 
 export const loggedInUserInfo = _.once((): Promise<CurrentRedditUser | void> =>
-	!isLoggedIn() ? Promise.resolve() : fetch('/api/me.json', { credentials: 'include' })
-		.then(r => r.json())
+	!isLoggedIn() ? Promise.resolve() : ajax({ url: '/api/me.json', type: 'json' })
```
andytuba: There appear to be no other instances of `fetch` in `lib/modules/**` and I'm sorry I didn't use `ajax` in the first place.

PR #4831, opened 2018-08-26
Revert "Temporarily disable FF integration tests"

erikdesjardins:
Reverts honestbleeps/Reddit-Enhancement-Suite#4830

To test if it started working again.


erikdesjardins:
Note to self: this now builds everything twice, see if we can pass in the target browser

PR #4835, opened 2018-08-28
Fix faulty XKCD alt text

kdreagle:
Relevant issue: N/A
Tested in browser: Chrome 70

Fixes [this bug](https://www.reddit.com/r/RESissues/comments/9asuvg/res_doesnt_display_xkcd_alt_texts_with_emojis/) found by [/u/DonaldLovesJong](https://www.reddit.com/user/DonaldLovesJong) and [/u/LevelSevenLaserLotus](https://www.reddit.com/user/LevelSevenLaserLotus).

Examples:
image
image

### DIFF
--- a/lib/modules/hosts/xkcd.js
+++ b/lib/modules/hosts/xkcd.js
@@ -23,7 +23,15 @@ export default new Host('xkcd', {
 		return {
 			type: 'IMAGE',
 			title,
-			caption: escapeHTML(decodeURIComponent(escape(alt))),
+			caption: escapeHTML(
+				// The XKCD API is broken for unicode characters which are multibyte in utf8,
+				// in that it escapes each byte individually (bytes are not codepoints!).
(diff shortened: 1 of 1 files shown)


kdreagle:
Maybe I should've kept the escapeHTML?

kdreagle:
I put the escapeHTML function back in to be safe, but it looks like it isn't needed at least for recent browsers because I tried injecting those escape characters after the decodeURI function call just to see what would happen. The first example I gave even has quote characters in it without any problems.

And as for the escape function, it is deprecated (not strictly), but encodeURI and encodeURIComponent do not work at all in terms of fixing the original bug.

Either way I guess this is a fix for most cases right now.

kdreagle:
Yeah I painstakingly tried cooking up a hack that could decode it for a few hours until I came across a much easier solution on stackoverflow 🙄

erikdesjardins:
Looks good!

PR #4838, opened 2018-08-28
commandLine: Close on Escape also when blurred

larsjohnsen:
Relevant issue: https://www.reddit.com/r/RESissues/comments/941l0n/issue_closing_filter_dialog/
Tested in browser: Chrome 70

### DIFF
--- a/lib/modules/commandLine.js
+++ b/lib/modules/commandLine.js
@@ -84,14 +84,17 @@ const commandLine = _.once(() => {
 		}
 	});
 
-	input.addEventListener('keyup', (e: KeyboardEvent) => {
+	document.addEventListener('keyup', (e: KeyboardEvent) => {
 		if (e.key === NAMED_KEYS.Escape) {
 			// close prompt.
 			toggleCmdLine(false);
-		} else {
-			// auto suggest?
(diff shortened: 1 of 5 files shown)


lib/modules/commandLine.js
```diff
@@ -84,16 +84,19 @@ const commandLine = _.once(() => {
...
 
-	input.addEventListener('keyup', (e: KeyboardEvent) => {
+	document.addEventListener('keyup', (e: KeyboardEvent) => {
 		if (e.key === NAMED_KEYS.Escape) {
 			// close prompt.
 			toggleCmdLine(false);
-		} else {
-			// auto suggest?
-			cmdLineHelper($(e.target).val());
+			e.stopImmediatePropagation();
```
erikdesjardins: Hmmm, this listener never gets removed from the document, do we want to call this for all keypresses?

erikdesjardins: Hmm, this listener never gets removed, so this might prevent other "press escape to close" things from working?

larsjohnsen: Yeah, I've overlooked that. That must be fixed.

PR #4843, opened 2018-09-03
Preload Batch font

larsjohnsen:
- Prevents icons (tag icon, subscribe, filter etc) flashing
- Avoids having the browser reflow the entire page (often takeing > 100 ms) after it's finally loaded

Tested in browser: Firefox 63, Chrome 70


larsjohnsen:
I should probably have checked whether Edge supports `FontFace` first… as it does not. Seems like the alternative method (adding a `<link rel="preload" as="font">`) must be used.

edge/fonts.scss
```diff
@@ -0,0 +1,6 @@
+@font-face {
+	font-family: 'Batch';
+	src: url('../lib/fonts/batch-icons-webfont.woff') format('woff');
```
erikdesjardins: Does `!url-loader!../lib/fonts/batch-icons-webfont.woff` work here? (to avoid special-casing the build script)

PR #4845, opened 2018-09-03
appveyor: Set image to Visual Studio 2017

mc10:
This fixes the current AppVeyor build failures.


mc10:
Hmm, looks like the build failures were spurious, but upgrading the image doesn't hurt.

PR #4850, opened 2018-09-05
[WIP] Replace string.html runtime templates with nanohtml compile time templates

erikdesjardins:
A step towards removing all usage of innerHTML and friends.

Nearly all of this needs testing.
Also, it makes the bundle 50k larger (or something).
Also, it doesn't generate great code, leaving in a bunch of strings containing unnecessary whitespace.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     ]
   },
   "dependencies": {
+    "nanohtml": "1.2.4",
     "fast-levenshtein": "2.0.6",
     "favico.js": "0.3.10",
     "jquery": "3.3.1",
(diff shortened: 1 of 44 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
     ]
   },
   "dependencies": {
+    "nanohtml": "1.2.4",
```
mc10: Nit: can you alphabetize this?

corylulu:
Do you worry at all that making the function just `html` might hurt it's searchability and might make it a bit less readable to those unfamiliar with the function? Perhaps something like `htmlParse` might be better.

larsjohnsen:
Some bugs I've noticed:
 - Wikipedia expando's `src` is empty
 - Command line state text is unstyled and not updated
 - Comment navigator buttons' HTML codes are escaped

PR #4856, opened 2018-09-12
 Makes video timeline control easier to click/use

corylulu:
I frequently misclicked when trying to move the timeline position on desktop, as it's only a 2px area that you have to get over. So I adjusted it so that the clickable space is larger, but still appears as a 2px bar. 

I also made the timeline slightly more visible.

The adjustments shouldn't interfere with any of the other controls or features of the video player.

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -375,8 +375,9 @@ img {
 
 		.video-advanced-progress {
 			width: 100%;
-			height: 2px;
-			background-color: rgba(256, 256, 256, .06);
+			height: 16px;
+			padding: 7px 0;
+			background-image: linear-gradient(rgba(0, 0, 0, 0) 0%, rgba(255, 255, 255, 0) 43.75%, rgba(255, 255, 255, 0.15) 43.75%, rgba(255, 255, 255, 0.15) 56.25%, rgba(255, 255, 255, 0) 56.25%, rgba(0, 0, 0, 0) 100%);
 			box-sizing: border-box;
 
 			.video-advanced-position-thumb {


corylulu:
@larsjohnsen Okay, I adjusted that, but kept 2 extra pixels of padding on the bottom since I think it's all too compact. 

Honestly, I think all the controls on that bar could be enlarged a bit and/or some increased padding (some better icons too). If nobody is opposed to that, maybe I'll tweak that a bit next. I was tempted to adjust it all on my first pass to make it all uniform and bit easier to use, but decided against it.

corylulu:
@larsjohnsen Okay, I adjusted that, but kept 2 extra pixels of padding on the bottom since I think it's all too compact. Honestly, I think all the controls on that bar could be enlarged a tiny bit and/or add some more padding (some better icons too). If nobody is opposed to that, maybe I'll tweak that a bit next. I was tempted to adjust it all on my first pass to make it all uniform and bit easier to use, but decided against it. 

Speaking of better icons, how would you guys feel about changing from the embedded "Batch" font to the ['Material Icons' font](http://google.github.io/material-design-icons/#icon-font-for-the-web)? The file size is very similar and it could be swapped out gradually since it's a different syntax, but the syntax is way more readable via code... It converts the whole name of the symbol into the symbol...

So to get a `►` symbol, instead of `content: '\f16c';` , it would just be `content: 'play_arrow'`

larsjohnsen:
Feel free to make the controls more accessible. :)

Having another font available and eventually just replacing Batch altogether would be great. I say go for it. Adding a new font is now a bit more complex than before though, see [here](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4843/files).

Issue #4556, opened 2017-11-19
React UI


wojtekmaj:
Hey,
as crazy as it may sound, I believe that RES could benefit a lot from being rewritten to React.

RES, being 6 years old (first commit 15.11.2011 - happy birthday! 🍰 ) has an architecture that offers far less than what modern libraries like React could give us.

Here's how I think RES could benefit:

* Much more structured architecture will help in project maintainability
* Increased code reusability. Having a clear component-based structure would make it easier to reuse code.
* Less bugs
    * Can you easily follow how a simple toggleButton works? It's hard - jQuery-based event handling, heavily reliant on things like document.querySelector, decentralized code.
    * Switching to component-based architecture can help writing good - and fast - unit tests.
* Performance - the amount of stuff that's going on currently in DOM is crazy. We render HTML oftenly using innerHTML, then we replace some containers with other innerHTMLs, then we find these very elements we've just rendered using querySelector, pushing them to global variable and applying events. To me, it sounds like we're doing twice as many DOM operations as we should. On Modern browsers, direct DOM operations are faster than innerHTML, too, but writing own API to do that does not make a lot of sense.
* In the process, we could naturally get rid of jQuery, which I find unnecessary in current state of JavaScript.

So I started experimenting with rewriting RES to React just to ensure that it is possible without freezing down the entire development process. And I think it's doable!

A repo is worth 1000 words though, so let me give you a sneak peek on what I'm into recently. It's on its very early stage, but worth a look.

https://github.com/wojtekmaj/Reddit-Enhancement-Suite/tree/react

### What works
* Displaying navigation
* Routing
* Displaying module
* Displaying most module options
* Toggling modules on/off
* Changing option values

### Components
* reusable
    * Input - for rendering any given input type - button, enum, text, etc.
    * Toggle
    * Table - for rendering tables with ability to add and remove rows
    * Button
    * Html - for rendering pure HTML into React elements
    * Markdown - for parsing markdown text
* ConsoleWrapper - to connect Console to Router and Store
    * ConsoleContainer
        * Search - for rendering search field
        * ModuleSelector - for displaying menu
        * ConfigPanelOptions - for displaying main content (currently selected module)
            * OptionContainer - for rendering option container with title, description and control
                * Option - for rendering specific option control (button, text, enum, ...)

### Technologies
* React-Router (hash-based)
* Redux store (monkey patched to get data from Modules.all())
    * Storing modules for display purposes
[...]

PR #4859, opened 2018-09-17
Settings console as embedded page

### LINKED ISSUE #2278: Show settings console in separate tab

larsjohnsen:
The settings console is now seamlessly embedded using an iframe. Other than it being a bit slower to open, there should not be any user impact.

- The URL hash of the embedded page and the parent stays in sync via messaging.
- Restricts running modules (`beforeLoad` and `go`) to those needed.
- settingsConsole is no longer included in the foreground bundle, and its CSS is separated.
- Added a context file including username, origin url, and modhash which is passed to the options page. This allows it to have URLs redirecting to correct Reddit subdomain, and make requests to Reddit from the background page.

TODO:

- [x] Closes the console unsaved if removing the hash (usually by going backwards in history a few too many times) without prompting
- [x] Links in settingsConsole (options / description) to Reddit pages must be absolute
- [x] All links must specify `_parent` (unless this is done in Javascipt)
- [x] When typing in search field, it is changed by outdated values routed via settingsNavigation  
- [x] Tests must be updated, preferably to use the standalone options page for simplicity
- [ ] The new options bundle is foreground + settingsConsole. There's probably some way to avoid duplicating the foreground bundle
- [x] Fix i18n (locale id cannot be retrieved, perhaps saved in the background)
- [x] Fix filteReddit subreddit sync (and other functionality which requires access to active username or Reddit APIs)
- [x] The open command line callback is broken as the module is not initialized.

Relevant issue: Fixes #2278, precursor for #4556 
Tested in browser: Chrome 70, Firefox 64


andytuba:
So exciting to see this!

lib/options/options.entry.js
```diff
@@ -0,0 +1,16 @@
+/* @flow */
+
+// include the LICENSE file
+// $FlowIgnore
+import 'file-loader?name=LICENSE!../../LICENSE';
```
erikdesjardins: This creates a file in the output dir so we only need it in `foreground.entry.js`

lib/core/init.js
```diff
@@ -18,6 +18,7 @@ import { migrate } from './migrate';
 let start;
 
 export function init() {
+	// TODO SessionStorage must be updated from iframe
```
larsjohnsen: An alternative could be checking whether the URL hash is `#res:disabled`.

lib/modules/nightMode.js
```diff
@@ -347,6 +347,7 @@ function timeStringToDate(timeString) {
 const className = () => {
 	switch (appType()) {
 		case 'r2':
+		case 'options':
```
larsjohnsen: Added due to nightMode otherwise adding bodyClass `res-d2x-nightmode`.

lib/modules/presets.js
```diff
@@ -47,36 +47,36 @@ module.options = {
...
 	},
 	cleanSlate: {
 		title: 'presetsCleanSlateTitle',
 		description: 'presetsCleanSlateDesc',
 		type: 'button',
 		text: 'apply preset',
-		callback: confirmPreset(() => {
-			toggleModules(undefined, '*');
-		}),
+		callback: confirmPreset(() => (
```
larsjohnsen: `toggleModules` is a bit slow, so returning a promise prevents the window be closed prematurely.

lib/modules/settingsNavigation.js
```diff
@@ -33,11 +32,59 @@ module.options = {
...
+const optionsPage = location.origin === getOptionsURL().origin;
+
+function checkStandaloneRedirect() {
+	// Testing requires access to the options page,
+	// and I couldn't figure out a way to get find the extension URL in nightwatch
+	if (location.hash.startsWith('#res:settings-redirect-standalone-options-page')) {
+		window.stop();
+		location.href = getOptionsURL(location.hash.replace('-redirect-standalone-options-page', '')).href;
+	}
+}
```
larsjohnsen: May be a bit slower and more unreliable than using checking webRequests, but redirecting with it would require 
 the `webRequestBlocker` permission.

lib/options/options.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
+<html res-options>
```
larsjohnsen: Added `res-options` to distinguishing the option page from other appTypes.

mc10:
I see both `#RESConsoleContainer` and `#console-container` used; what is the difference between the two?

lib/modules/settingsNavigation.js
```diff
@@ -79,96 +120,58 @@ export function makeUrlHash(moduleID?: string, optionKey?: string): string {
...
-		return;
+let iframe;
+
+export function open(moduleID?: string, optionKey?: string) {
+	if (iframe || isOptionsPage()) {
+		(iframe && iframe.contentWindow || window).postMessage({ load: { moduleID, optionKey } }, '*');
+	} else {
+		window.addEventListener('message', listener);
+		iframe = document.createElement('iframe');
+		iframe.id = 'console-container';
```
larsjohnsen: @mc10 `#console-container` is the iframe which embeds the options page, but should probably . I have not changed any of class names or ids on the options page, so all ids and class names there stays the same.

lib/modules/settingsNavigation.js
```diff
@@ -60,17 +60,17 @@ module.go = () => {
...
-			iframe.contentWindow.postMessage({ close: true }, '*');
-		}
-	};
-
-	window.addEventListener('popstate', updateConsole);
+	window.addEventListener('popstate', () => { updateConsole(); });
 	updateConsole();
+
+	document.body.addEventListener('click', (e: MouseEvent) => {
+		const url = e.target instanceof HTMLAnchorElement && new URL(e.target.href);
```
corylulu: `const url = !e.ctrlKey && e.target instanceof HTMLAnchorElement && new URL(e.target.href);`

would resolve Andy's concern about ignoring ctrl+clicks

PR #4865, opened 2018-09-28
Add quarantineHide module for hiding quarantine warnings

git-uname:
Add module to hide warnings for quarantined subreddits.

### DIFF
--- a/lib/modules/quarantineHide.js
+++ b/lib/modules/quarantineHide.js
@@ -1,0 +1,62 @@
+/* @flow */
+
+import { Module } from '../core/module';
+import { $ } from '../vendor';
+
+export const module: Module<*> = new Module('quarantineHide');
+
+module.moduleName = 'qhideName'; // i18n
+module.category = 'appearanceCategory';
+module.description = 'qhideDesc'; // i18n
+module.options = {
+	hideFlair: {
+		title: 'qhideFlairTitle', // i18n
+		type: 'boolean',
+		value: true,
+		description: 'qhideFlairDesc', // i18n
+	},
+	hideSidebar: {
+		title: 'qhideSidebarTitle', // i18n
+		type: 'boolean',
+		value: true,
+		description: 'qhideSidebarDesc', // i18n
+	},
+	fixHeaderImg: {
+		title: 'qhideFixHeaderImgTitle', // i18n
+		type: 'boolean',
+		value: true,
+		description: 'qhideFixHeaderImgDesc', // i18n
+	},
+	hideWarning: {
+		title: 'qhideWarningTitle', // i18n
+		type: 'boolean',
+		value: false,
+		description: 'qhideWarningDesc', // i18n
+	},
+};
+
+module.include = ['linklist'];
+
+module.go = () => {
+	if (module.options.hideFlair.value) {
+		removeFlair();
+		window.addEventListener('neverEndingLoad', removeFlair);
+	}
+	if (module.options.hideSidebar.value) {
+		const $notice = $('.quarantine-notice');
+		if (module.options.fixHeaderImg && $notice.length !== 0) {
+			$('body').removeClass('quarantine');
+		}
+		$notice.hide();
+	}
+	if (module.options.hideWarning.value) {
+		const $interimg = $('.interstitial-image');
+		if ($interimg.length !== 0 && $interimg.attr('alt') === 'quarantined') {
+			$('.c-btn').click();
+		}
+	}
+};
+
+function removeFlair() {
+	$('.quarantine-stamp').parent().remove();
+}
--- a/lib/modules/quarantineHide.js
+++ b/lib/modules/quarantineHide.js
@@ -32,7 +32,7 @@ module.go = () => {
 		removeStamp();
 		window.addEventListener('neverEndingLoad', removeStamp);
 	}
-	if (inQuarantinedSubreddit && module.options.hideQuarantinedInSub) {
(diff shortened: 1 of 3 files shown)


locales/locales/en.json
```diff
@@ -4381,5 +4381,35 @@
 	},
 	"faviconUseLegacyDesc": {
 		"message": "If you don't like the new orange Reddit tab icon, revert to using the old Reddit icon instead."
+	},
+	"qhideName": {
```
andytuba: Could you update these to `quarantineHideDesc`, `quarantineHideFlair`, etc? consistency makes for easier Ctrl-F'ing.

lib/modules/quarantineHide.js
```diff
@@ -0,0 +1,62 @@
...
+	if (module.options.hideSidebar.value) {
+		const $notice = $('.quarantine-notice');
+		if (module.options.fixHeaderImg && $notice.length !== 0) {
+			$('body').removeClass('quarantine');
+		}
+		$notice.hide();
+	}
+	if (module.options.hideWarning.value) {
+		const $interimg = $('.interstitial-image');
+		if ($interimg.length !== 0 && $interimg.attr('alt') === 'quarantined') {
```
andytuba: The `alt` text is localized (translated), is there a different property or technique you can use to detect "on the quarantine page"? (something that doesn't rely on user-facing text) Perhaps look in the url of the `action` of the form?

lib/modules/quarantineHide.js
```diff
@@ -0,0 +1,62 @@
...
+
+module.go = () => {
+	if (module.options.hideFlair.value) {
+		removeFlair();
+		window.addEventListener('neverEndingLoad', removeFlair);
+	}
+	if (module.options.hideSidebar.value) {
+		const $notice = $('.quarantine-notice');
+		if (module.options.fixHeaderImg && $notice.length !== 0) {
+			$('body').removeClass('quarantine');
```
andytuba: Could you use `BodyClasses.toggle` for this? You can `import { BodyClasses } from 'lib/utils';`

lib/modules/quarantineHide.js
```diff
@@ -0,0 +1,62 @@
...
+module.include = ['linklist'];
+
+module.go = () => {
+	if (module.options.hideFlair.value) {
+		removeFlair();
+		window.addEventListener('neverEndingLoad', removeFlair);
+	}
+	if (module.options.hideSidebar.value) {
+		const $notice = $('.quarantine-notice');
+		if (module.options.fixHeaderImg && $notice.length !== 0) {
```
andytuba: Could you put this `if fixHeaderImg` at the same level as `if hideSidebar`? Since these are unrelated chunks of functionality, they should be in separate blocks. 

It looks like you combined them to take advantage of `$notice.length` to determine "does this page show quarantined UI?"  It would be clearer to add something like `const inQuarantinedSubreddit = !!(document.querySelector('.quarantine-notice'))` to the top of `go()`. 
It would be even nicer to add `inQuarantinedSubreddit = _.once(() => { look for hints of quarantine });` to `lib/utils/location` (and import it into `lib/utils` for importing into modules.) You could then use that function in these `if` predicates.

lib/modules/quarantineHide.js
```diff
@@ -0,0 +1,62 @@
...
+
+module.include = ['linklist'];
+
+module.go = () => {
+	if (module.options.hideFlair.value) {
+		removeFlair();
+		window.addEventListener('neverEndingLoad', removeFlair);
+	}
+	if (module.options.hideSidebar.value) {
+		const $notice = $('.quarantine-notice');
```
andytuba: To determine if this is a quarantined subreddit, it would probably be cleaner and more robust to check if `document.body.classList.has('quarantine')`

lib/modules/quarantineHide.js
```diff
@@ -0,0 +1,62 @@
...
+	}
+	if (module.options.hideWarning.value) {
+		const $interimg = $('.interstitial-image');
+		if ($interimg.length !== 0 && $interimg.attr('alt') === 'quarantined') {
+			$('.c-btn').click();
+		}
+	}
+};
+
+function removeFlair() {
```
andytuba: nit - could you name this `stamp` instead of `flair`?

lib/modules/quarantineHide.js
```diff
@@ -0,0 +1,62 @@
...
+	},
+	hideWarning: {
+		title: 'qhideWarningTitle', // i18n
+		type: 'boolean',
+		value: false,
+		description: 'qhideWarningDesc', // i18n
+	},
+};
+
+module.include = ['linklist'];
```
larsjohnsen: Also 'comments' and 'wiki' contains this warning.

lib/modules/quarantineHide.js
```diff
@@ -0,0 +1,43 @@
...
+};
+
+module.include = ['linklist', 'comments', 'wiki'];
+
+module.go = () => {
+	if (module.options.hideFlair.value) {
+		removeStamp();
+		window.addEventListener('neverEndingLoad', removeStamp);
+	}
+	if (inQuarantinedSubreddit && module.options.hideQuarantinedInSub) {
```
larsjohnsen: `inQuarantinedSubreddit` is a function; use `inQuarantinedSubreddit()`.
`module.options.hideQuarantinedInSub` is the option object; the value is stored in `module.options.hideQuarantinedInSub.value`

larsjohnsen:
Thanks!

PR #4877, opened 2018-10-11
Adding links to sort options in subredditInfo card to resolve #1431

### LINKED ISSUE #1431: add /new, /top, etc links to subreddit info hover
and render them before the subreddit metadata

http://www.reddit.com/r/Enhancement/comments/2a03ic/feature_request_when_hovering_over_a_subreddit/

andrewboerema:
Relevant issue: resolves #1431 

### DIFF
--- a/lib/modules/subredditInfo.js
+++ b/lib/modules/subredditInfo.js
@@ -89,7 +89,17 @@ function handleMouseOver(e: Event) {
 }
 
 async function showSubredditInfo(card, subreddit) {
-	const header = string.html`<div><a href="/r/${subreddit}">/r/${subreddit}</a></div>`;
+	const header = string.html`
+		<div class="subredditInfoToolTipHeader">
+			<a href="/r/${subreddit}" class="subredditTitleLink">/r/${subreddit}</a>
+		</div>`;
+
+	const makeSortLink = (subreddit, sort) =>
+		$('<a />').attr('href', `/r/${subreddit}/${sort}/`).addClass('subredditSortLink').text(sort);
+
+	['hot', 'new', 'controversial', 'top', 'rising'].map(sort => {
+		header.appendChild(makeSortLink(subreddit, sort)[0]);
+	});
 
 	if (loggedInUser()) {
 		const subscribeToggle = $('<span />')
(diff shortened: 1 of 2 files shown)


lib/modules/subredditInfo.js
```diff
@@ -89,7 +89,17 @@ function handleMouseOver(e: Event) {
...
 
 async function showSubredditInfo(card, subreddit) {
-	const header = string.html`<div><a href="/r/${subreddit}">/r/${subreddit}</a></div>`;
+	const header = string.html`
+		<div class="subredditInfoToolTipHeader">
+			<a href="/r/${subreddit}" class="subredditTitleLink">/r/${subreddit}</a>
+		</div>`;
+
+	const makeSortLink = (subreddit, sort) =>
+		$('<a />').attr('href', `/r/${subreddit}/${sort}/`).addClass('subredditSortLink').text(sort);
```
andytuba: Could you convert this to a `string.html` template tag like `header`? Much more readable.

lib/modules/subredditInfo.js
```diff
@@ -89,7 +89,17 @@ function handleMouseOver(e: Event) {
...
+
+	['hot', 'new', 'controversial', 'top', 'rising'].map(sort => {
+		header.appendChild(makeSortLink(subreddit, sort)[0]);
```
andytuba: You can also embed this directly into the above template.

lib/modules/subredditInfo.js
```diff
@@ -89,7 +89,17 @@ function handleMouseOver(e: Event) {
...
-	const header = string.html`<div><a href="/r/${subreddit}">/r/${subreddit}</a></div>`;
+	const header = string.html`
+		<div class="subredditInfoToolTipHeader">
+			<a href="/r/${subreddit}" class="subredditTitleLink">/r/${subreddit}</a>
+		</div>`;
+
+	const makeSortLink = (subreddit, sort) =>
+		$('<a />').attr('href', `/r/${subreddit}/${sort}/`).addClass('subredditSortLink').text(sort);
+
+	['hot', 'new', 'controversial', 'top', 'rising'].map(sort => {
```
Brcrwilliams: Bit of a nitpick, but you should use `forEach` instead of `map` here. Map is intended for making a copy of an array with changes to each element.

andrewboerema:
@andytuba - do you have an example of the kind of dropdown you were thinking? I haven't contributed to RES before and I would like to stay in line with the components you use.

PR #4878, opened 2018-10-12
pageNavigator showLink: Fix ad sometimes being shown instead of submission

larsjohnsen:
Appears that Reddit sometimes inserts ads along with the submission on comment pages.

https://www.reddit.com/r/Enhancement/comments/9nha5n/why_am_i_getting_an_ad_in_the_page_navigator/

Tested in browser: Firefox 63

### DIFF
--- a/lib/modules/pageNavigator.js
+++ b/lib/modules/pageNavigator.js
@@ -3,7 +3,7 @@
 import _ from 'lodash';
 import { $ } from '../vendor';
 import { Module } from '../core/module';
-import { getHeaderOffset, isPageType, scrollTo, Thing, frameThrottle, string } from '../utils';
+import { getHeaderOffset, isPageType, scrollTo, frameThrottle, string, watchForThings } from '../utils';
 import { i18n } from '../environment';
 import * as Floater from './floater';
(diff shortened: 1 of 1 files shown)


lib/modules/pageNavigator.js
```diff
@@ -95,9 +98,8 @@ function backToNewCommentArea() {
 	Floater.addElement($backToNewCommentArea);
 }
 
-function showLinkTitle() {
+const showLinkTitle = _.once(submissionThing => {
```
andytuba: Instead of `_.once`, how about validating that `submissionThing` is a child of `#siteTable`?

larsjohnsen: The `_.once` construction is also used by hideChildComments to determine the submission thing, so I figure we might as well use it. While I didn't see the ad myself, the screenshot in the issue shows that the `hide all child comments` button was placed correctly so I think solution is sufficient.

andytuba: Gets the job done! If the ads team happens to add another ad unit that's sooner in the DOM (unlikely) we can revisit it.

PR #4893, opened 2018-10-25
Better intergrate floaters in redesign

larsjohnsen:
I am a little unsure about this solution's robustness, but this works by pushing the element `.header-user-dropdown` to the left by adding a margin. The margin may at some point be overwritten, which would cause to floater to overlay the profile button, though I didn't see that happen at any time while testing. It may be desirable to get a assigned element up there though.

The `belowFixedNavbar` style is kept as fallback in case the class name is changed.

image

Tested in browser: Chrome 70


larsjohnsen:
I'm merging this and hoping that this won't cause conflict any time soon. If any issue should arise, the `inNavbar` container can easily be removed and `belowFixedNavbar` will automatically be used instead.

Issue #2697, opened 2016-01-30
Add link to reddit "preferred comment sort" on comment sort dropdown


andytuba:
https://www.reddit.com/r/Enhancement/comments/43f342/feature_request_permanently_override_suggest_sort/czhqmsw?context=3

gdtrombetti:
I will add this link

johnnybigoode:
Heya, I think this is still open? I'd love to take this.

andytuba:
@johnnybigoode  Go for it!

johnnybigoode:
Thank you. I'm still working on it. It's been a crazy week but I'm sure I can do it.

radhakri555:
Can i take up this issue?

ofekp:
@radhakri555 are you on it?

andytuba:
Thanks for adding that to the redesign, @ofekp! 

If anyone still wants to implement this on old.reddit.com, it would be neat for you to extend #4898 with also append the gear link to the legacy frontend's sort menu.

PR #4898, opened 2018-10-27
Adds quick link to comment sort-by preference menu

ofekp:
Relevant issue: https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/2697
Tested in browser:  Google Chrome 70.0.3538.77

### DIFF
--- a/lib/modules/commentSortBy.js
+++ b/lib/modules/commentSortBy.js
@@ -5,21 +5,8 @@ import { $ } from '../vendor';
 import { Module } from '../core/module';
 import { rawLocale } from '../utils/localization';
 import {
-	loggedInUser,
-	watchForThings,
-	Alert,
-	CreateElement,
-	Thing,
-	batch,
-	click,
-	downcast,
-	isCurrentSubreddit,
-	isPageType,
 	string,
-	watchForElements,
-	empty,
 	watchForRedditEvents,
-	isAppType,
 } from '../utils';
 
 export const module: Module<*> = new Module('commentSortBy');
@@ -27,8 +14,8 @@ export const module: Module<*> = new Module('commentSortBy');
 export const commentSortSelector = "#CommentSort--SortPicker";
 
 module.beforeLoad = () => {
-	watchForRedditEvents('postModTools', (element) => {
-		applyToPost(element);
+	watchForRedditEvents('postModTools', () => {
+		applyToPost();
 	});
 };
 
@@ -61,7 +48,7 @@ export class SortByPrefLink {
 	}
 }
 
-function applyToPost(element: HTMLAnchorElement | HTMLElement) {
+function applyToPost() {
 	SortByPrefLink.render();
 }
 
--- a/lib/modules/commentSortBy.js
+++ b/lib/modules/commentSortBy.js
@@ -1,6 +1,5 @@
 /* @flow */
 
-import _ from 'lodash';
 import { $ } from '../vendor';
 import { Module } from '../core/module';
 import { i18n } from '../environment';
--- a/lib/modules/commentSortBy.js
+++ b/lib/modules/commentSortBy.js
@@ -12,7 +12,7 @@ import {
 export const module: Module<*> = new Module('commentSortBy');
 
 module.beforeLoad = () => {
-	watchForRedditEvents('postModTools', (element, { id, _: { update } }) => {
+	watchForRedditEvents('postModTools', (element, { _: { update } }) => {
 		if (update) {
 			return;
 		}
--- a/lib/modules/commentSortBy.js
+++ b/lib/modules/commentSortBy.js
@@ -31,6 +31,7 @@ export class SortByPrefLink {
 				style="opacity: 0.3;"
 				title="${i18n('addCommentSortPrefMenuItemLinkTitle')}"
 				href="https://old.reddit.com/prefs/#default_comment_sort"
+                                target="_blank"
 			>${'\u00A0'/* nbsp */}</a>
 		`;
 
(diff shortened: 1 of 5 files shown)


lib/modules/commentSortBy.js
```diff
@@ -0,0 +1,52 @@
...
+import { rawLocale } from '../utils/localization';
+import {
+	string,
+	watchForRedditEvents,
+} from '../utils';
+
+export const module: Module<*> = new Module('commentSortBy');
+
+module.beforeLoad = () => {
+	watchForRedditEvents('postModTools', () => {
```
andytuba: Could you pass along the `postModTools` JSAPI container to `applyToPost`?

lib/modules/commentSortBy.js
```diff
@@ -0,0 +1,52 @@
...
+	});
+};
+
+module.moduleName = 'commentSortBy';
+module.category = 'commentsCategory';
+module.disabledByDefault = false;
+module.description = 'addCommentSortPrefMenuItemDesc';
+
+export class SortByPrefLink {
+	static defaultLinkElement = (e => () => e().cloneNode(true))(_.once(() => SortByPrefLink.buildTagElement()));
```
andytuba: This class is overkill for creating a little link. Can you keep `function buildTagElement` and delete the rest of this class and `afterLoad`?

lib/modules/commentSortBy.js
```diff
@@ -0,0 +1,52 @@
...
+					style="opacity: 0.3;"
+					title="Set default sort by preferences"
+					href="https://${rawLocale()}.reddit.com/prefs/#default_comment_sort"
+				>${'\u00A0'/* nbsp */}</a>
+			</span>
+		`;
+
+	static render() {
+		const bttn = SortByPrefLink.defaultLinkElement();
+		if ($('#CommentSort--SortPicker').parent().find('.sortByPrefLink').length === 0) {
```
andytuba: Can you test this inside `applyToPost`, and add a flag to the `postModTools` container instead to verify "this element has been handled"? or you could check the postmodtools `.updates` value.

lib/utils/watchers.js
```diff
@@ -109,6 +109,7 @@ declare function watchForRedditEvents(type: 'postAuthor', callback: (HTMLElement
 declare function watchForRedditEvents(type: 'post', callback: (HTMLElement, PostEventData) => void | Promise<void>): void;
 declare function watchForRedditEvents(type: 'userHovercard', callback: (HTMLElement, UserHovercardEventData) => void | Promise<void>): void;
 declare function watchForRedditEvents(type: 'commentAuthor', callback: (HTMLElement, CommentAuthorEventData) => void | Promise<void>): void;
+declare function watchForRedditEvents(type: 'postModTools', callback: () => void | Promise<void>): void;
```
andytuba: Can you pass along the container here?

andytuba:
Thanks for jumping in on this task! I was just hoping for old reddit, so it's neat to see contributions on the redesign!

lib/modules/commentSortBy.js
```diff
@@ -0,0 +1,52 @@
...
+module.moduleName = 'commentSortBy';
+module.category = 'commentsCategory';
+module.disabledByDefault = false;
+module.description = 'addCommentSortPrefMenuItemDesc';
+
+export class SortByPrefLink {
+	static defaultLinkElement = (e => () => e().cloneNode(true))(_.once(() => SortByPrefLink.buildTagElement()));
+	static buildTagElement = () => string.html`
+			<span class="RESUserTag">
+				<a
```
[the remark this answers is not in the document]
ofekp: Do you mean `${string.safe(SettingsNavigation.makeUrlHashLink(module.moduleID, 'hardIgnore', ' configure ', 'gearIcon'))}`
It only links to the extension settings, not to the Reddit settings. Also my gear has an opacity of 0.3 otherwise it is too obnoxious over the white background.

lib/modules/commentSortBy.js
```diff
@@ -0,0 +1,52 @@
...
+
+module.moduleName = 'commentSortBy';
+module.category = 'commentsCategory';
+module.disabledByDefault = false;
+module.description = 'addCommentSortPrefMenuItemDesc';
+
+export class SortByPrefLink {
+	static defaultLinkElement = (e => () => e().cloneNode(true))(_.once(() => SortByPrefLink.buildTagElement()));
+	static buildTagElement = () => string.html`
+			<span class="RESUserTag">
```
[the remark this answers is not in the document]
ofekp: 🏁 
No need for that whatsoever, I removed it entirely, hope it is OK.

lib/modules/commentSortBy.js
ofekp: 🏁

ofekp:
@andytuba Please review my changes.
Thank you 😺

ofekp:
@andytuba Kind reminder.

lib/modules/commentSortBy.js
andytuba: Sounds good! Definitely want to ensure there's only ever one

andytuba: Just curious, when you say "not bulletproof" -- can you describe what you mean, and do you have any suspects what caused it? I'm wondering what unexpected pitfalls we need to learn about when working with the new reddit frontend.

andytuba:
Thanks for adding this in!  It's great to see a ticket closed out -- and on the new frontend!

lib/modules/commentSortBy.js
ofekp: Of course, I meant that without it, and contrary to what I thought should happen, I occasionally saw more than one setting gears. This means the id that the update event is receiving is not consistent enough to ensure that there is really only one setting gear. Unfortunately, I could not find the reason for that. Perhaps, and this is really shooting out of my hip, Reddit are reusing DOMs and therefore we're getting the same id in the event...

PR #4905, opened 2018-10-31
Added a new basis for french translation

nclsppr:
Relevant issue: 
Tested in browser:


BenMcGarry:
Thank you for the contribution however we do not accept translations within the core language files. 

Please see our translation document here: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/locales/locales/README.md

PR #4915, opened 2018-11-07
 redesign: Add expandos in comments

larsjohnsen:
- Adds expandos to comments and selftext when the comment page / lightbox
- The catch-all expando (`video`) button is not native to d2x, so I just added a placeholder button

Tested in browser: Chrome 71

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -35,6 +35,7 @@ import {
 	waitForEvent,
 	watchForElements,
 	watchForThings,
+	watchForRedditEvents,
 	getPercentageVisibleYAxis,
 	getViewportSize,
 } from '../utils';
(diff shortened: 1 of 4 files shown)


andytuba:
> The video expando CSS is not available in redesign, so those expando buttons don't have a background image.

We should go ahead and add that CSS to RES, with selector like `.res-app-d2x .expando-button.video { ... }`

lib/modules/showImages.js
```diff
@@ -488,30 +489,22 @@ module.beforeLoad = () => {
 		`);
 	}
 
-	watchForElements(['siteTable'], getSelector(), checkElementForMedia);
-	watchForElements(['selfText', 'newComments'], getSelector(true), checkElementForMedia);
-};
+	watchForElements(['selfText'], null, scanBody);
+	watchForThings(['comment', 'message'], thing => scanBody(thing.getTextBody()));
```
andytuba: May as well add `post` in here, so RES can show selftext expandos in lightbox and comments full-page views.  

I'll make a note that the redesign JSAPI add a property like "expando showing" which is updated when self-text expandos are opened.

PR #4918, opened 2018-11-12
Add navigation links to neverEndingReddit feature

fenuks:
Hello!
I'm very fond of neverEndingReddit feature, but I find it problematic that prev/next page links are not available in GUI boxes. My workaround so far was to disable autoloading, and then click ‘open next page’ URL. I think it makes sense to always show prev/next, and this PR adds this.

### DIFF
--- a/lib/modules/neverEndingReddit.js
+++ b/lib/modules/neverEndingReddit.js
@@ -103,6 +103,7 @@ const dupeSet = new Set();
 const siteTable = _.once(() => Thing.thingsContainer());
 let currentPageNumber = 1;
 let nextPageUrl;
+let prevPageUrl;
 let $NREPause, isPaused, pauseReason, pauseAfterPages, nextPausePage;
 
 let initAutoNextPage: ?() => void;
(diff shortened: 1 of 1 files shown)


andytuba:
What's the point of showing the prev/next buttons when you have infinite scroll?  This doesn't make sense to me, so I imagine I'm missing some context.

fenuks:
Hello, to answer your question, for example I have few multireddits that I browse once a week, which means that I have many, many pages to catch up, and sometimes I want to make a break, or move with browsing to mobile device, etc. As side note, I think that most often infinite scrolls give little power for user to control it, or return to the point where browsing has ended previous time.

In case of Never-Ending Reddit workaround to ‘extract’ next URL is, as noted above, to pause it, but it's a bit tedious, doesn't work instantly (user has to scroll through next page to have box with next page link), and I simply think it's convenient to have this information available ‘out-of-box’.

larsjohnsen:
Added in 7ea56c39280fe8bfc3497001dc246d1939d2a9ab

xenius:
Ive setup NER to pause when i reach 3th page. But then when i click next page which opens a new page, but it doesn't unpause NER. Is it possible to renable this?

PR #4919, opened 2018-11-13
Leave NP mode when navigating away using keyboard shortcuts

### LINKED ISSUE #2731: gomode should leave noparticipation

ackinc:
Relevant issue: #2731 

This PR fixes #2731, and does a bit more: all go-mode shortcuts (except next page, prev page, and subreddit front page) will take the user out of NP even when "useGoMode" is false. Previous discussion about this is [here](https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4724#discussion_r178501595). It made sense to me that a user be taken out of NP when they visit their inbox etc., regardless of the useGoMode setting's status.

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -789,7 +789,7 @@ module.options = {
 		title: 'keyboardNavsSubredditFrontPageTitle',
 		callback() {
 			const sub = currentSubreddit();
-			if (sub) navigateTo(`/r/${sub}`, { leaveNP: true });
+			if (sub) navigateTo(`/r/${sub}`, { leaveNP: false });
 		},
 		goMode: true,
 	},


lib/modules/keyboardNav.js
```diff
@@ -1219,7 +1219,18 @@ function reply(selected = getSelected()) {
 	));
 }
 
-function navigateTo(newWindow, href) {
+function navigateTo(newWindow, href, leaveNP = false) {
+	if (leaveNP) {
+		// convert relative links to absolute (needed for the call to nonNpLocation below)
+		if (href.startsWith('/')) {
```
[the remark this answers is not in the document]
ackinc: Calling new URL with a relative link throws a TypeError, so I had to put this entire block in

lib/modules/keyboardNav.js
```diff
@@ -1219,7 +1219,18 @@ function reply(selected = getSelected()) {
...
+			const url = new URL(location.href);
+			url.pathname = href;
+			href = url.href;
+		}
+
+		href = NoParticipation.nonNpLocation(href);
```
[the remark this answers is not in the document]
ackinc: Yes, that is a problem. I will try to figure out how to store the subdomain the user was on before entering np, so it can be restored in NoParticipation.nonNpLocation

lib/modules/keyboardNav.js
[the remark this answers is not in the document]
ackinc: Would it make sense to spin this block off into a separate "relativeToAbsUrl" function, and put it somewhere in util?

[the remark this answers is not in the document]
larsjohnsen: Sorry, forgot that you'll also have provide the base for relative URLs: `(new URL(href, location.origin)).href`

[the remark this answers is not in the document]
ackinc: Thank you, I've made the change

lib/modules/keyboardNav.js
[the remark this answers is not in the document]
ackinc: Pushed a fix. Using goMode no longer forces users onto www, unless they were in NP.

lib/modules/keyboardNav.js
```diff
@@ -1219,7 +1219,13 @@ function reply(selected = getSelected()) {
 	));
 }
 
-function navigateTo(newWindow, href) {
+function navigateTo(newWindow, href, leaveNP = false) {
```
andytuba: Since I suspect this function is headed towards adding more options, it would be nice to reorder the function signature into an options dictionary (keyword args) pattern. 

So, if you write it like this (or with flow typing and drop the false default)
```javascript
function navigateTo(href, { newWindow: false, leaveNP: false })
```

Then you can invoke it like this:


```javascript
 navigateTo('/example', { leaveNP: true })
```

andytuba:
Agreed on your assessment of next/prev. It seems appropriate to keep np on while you're still in the same listing (subreddit, multi, whatever).

lib/modules/keyboardNav.js
```diff
@@ -1219,7 +1219,13 @@ function reply(selected = getSelected()) {
...
+	if (leaveNP) {
```
andytuba: Could you also verify that the NoParticipation module and its `escapeNP` option are both enabled? There should be a helper like "get opaque option value", there are some other modules which check options from other modules. (Let me know if you'd like a pointer.)

andytuba: Even better, could you add a new `escapeNPKeyboardNav` option right after it, and add a new human-readable title/description to `en.json`?  You can just copy paste and add a qualifier like "when using keyboard navigation to go to another page."

And for the coup de gras: could you also add a new "data migration" to copy the current value of the `escapeNP` option into the new option. `migrate.js` is full of examples.

lib/modules/keyboardNav.js
ackinc: Commit #59e628 changes the signature as recommended

lib/modules/keyboardNav.js
ackinc: If we are introducing a new escapeNPKeyboardNav option:

Some users might be confused that the escapeNP option has no effect on keyboard nav. Should I edit its description to say it only affects link clicks?

ackinc: > Could you also verify that the NoParticipation module and its `escapeNP` option are both enabled? There should be a helper like "get opaque option value", there are some other modules which check options from other modules. (Let me know if you'd like a pointer.)

Added this in commit #4789f18

lib/core/modules/modules.js
```diff
@@ -132,3 +132,7 @@ export function getByCategory(category: string): Array<Module<any>> {
 			return (i18n(a.moduleName).toLowerCase() > i18n(b.moduleName).toLowerCase()) ? 1 : -1;
 		});
 }
+
+export function getOptionValue(opaqueId: OpaqueModuleId, option: string) {
```
andytuba: I could have sworn there was already some pattern for doing this, like ...

```javascript
import SomeModule from './someModule.js';

// snip

if (Module.isEnabled(SomeModule) && SomeModule.options.someOption.value) {
```

lib/modules/keyboardNav.js
andytuba: Hmm, they might be confused. Updating a description is tricky, though, since the other translations won't mention the link clicks. I think it's an acceptable compromise to just add the new option and leave the old one alone.

It would probably help to make `escapeNPKeyboardNav` depend on `escapeNP` being enabled. There's a pattern like `enabled => module.options.escapeNP.value`.

lib/modules/keyboardNav.js
```diff
@@ -1219,8 +1219,16 @@ function reply(selected = getSelected()) {
 	));
 }
 
-function navigateTo(newWindow, href) {
-	if (newWindow) {
+function navigateTo(href, options: { newWindow?: boolean, leaveNP?: boolean }) {
+	if (options.leaveNP &&
+	Modules.isEnabled(NoParticipation) &&
+	Modules.getOptionValue(NoParticipation, 'escapeNP')) {
```
andytuba: You can simply test `NoParticipation.options.escapeNP.value` here.

lib/core/modules/modules.js
andytuba: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4919/files#r233343392

lib/modules/keyboardNav.js
andytuba: Let's not bother with a new option for this PR.

lib/core/modules/modules.js
ackinc: Made this change and got rid of getOptionValue in the latest commits

lib/modules/keyboardNav.js
```diff
@@ -1219,8 +1219,16 @@ function reply(selected = getSelected()) {
 	));
 }
 
-function navigateTo(newWindow, href) {
-	if (newWindow) {
+function navigateTo(href, options: { newWindow?: boolean, leaveNP?: boolean }) {
+	if (options.leaveNP &&
```
larsjohnsen: Nit: Newline before/after parentheses on multi-line conditionals

lib/modules/keyboardNav.js
```diff
@@ -1219,8 +1219,16 @@ function reply(selected = getSelected()) {
 	));
 }
 
-function navigateTo(newWindow, href) {
-	if (newWindow) {
+function navigateTo(href, options: { newWindow?: boolean, leaveNP?: boolean }) {
```
larsjohnsen: `options` can be deconstructed to `{ newWindow = false, leaveNP = false }`

lib/modules/keyboardNav.js
```diff
@@ -1219,8 +1219,16 @@ function reply(selected = getSelected()) {
...
+	if (options.leaveNP &&
+	Modules.isEnabled(NoParticipation) &&
+	NoParticipation.module.options.escapeNP.value) {
+		// convert relative links to absolute (needed for the call to nonNpLocation below)
```
larsjohnsen: It would be cleaner to pass the relative URL to `nonNpLocation` …

lib/modules/noParticipation.js
```diff
@@ -108,9 +112,9 @@ function isSubscriber() {
 	return document.body.classList.contains('subscriber');
 }
 
-function nonNpLocation() {
-	const nonNpUrl = new URL(location.href);
-	nonNpUrl.hostname = 'www.reddit.com';
+export function nonNpLocation(href: string | void) {
+	const nonNpUrl = new URL(href || location.href);
```
larsjohnsen: … and change this to `new URL(href, location.href)`

lib/modules/keyboardNav.js
ackinc: Yes, but then each new option we add support for in navigateTo would have to be specified twice - once in the deconstruction, and once in the type definition. If we want that, I'll make the change

lib/modules/keyboardNav.js
ackinc: Changed in commit #0c385ee

lib/modules/keyboardNav.js
mc10: I think @larsjohnsen is suggesting to combine the deconstruction and type definition by replacing `options: { newWindow?: boolean, leaveNP?: boolean }` entirely with that snippet.

larsjohnsen: Both ways are fine with me.

It would also be worthwhile to make the object type exact (`{| newWindow?: boolean, leaveNP?: boolean |}`) to prevent typos when invoking the function.

ackinc: Replacing ```options: { newWindow?: boolean, leaveNP?: boolean }``` entirely with ```{ newWindow = false, leaveNP = false }``` results in flow errors

**Cannot call navigateTo with object literal bound to the second parameter because property leaveNP is missing in object literal**

for every time navigateTo is called with one of the options missing (like ```navigateTo('/', {newWindow: true})```). I'll try to figure out how to make this work.

ackinc: > Both ways are fine with me.
> 
> It would also be worthwhile to make the object type exact (`{| newWindow?: boolean, leaveNP?: boolean |}`) to prevent typos when invoking the function.

Making function param destructuring work with flow types is a bit harder than I expected. I've made the 'object type exact' change for now.

lib/modules/noParticipation.js
```diff
@@ -108,9 +112,9 @@ function isSubscriber() {
 	return document.body.classList.contains('subscriber');
 }
 
-function nonNpLocation() {
-	const nonNpUrl = new URL(location.href);
-	nonNpUrl.hostname = 'www.reddit.com';
+export function nonNpLocation(href: string = location.href) {
+	const nonNpUrl = new URL(href, location.origin);
```
mc10: Is `location.origin` supported on Edge? The [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/origin#Browser_compatibility) says that compatibility info is unknown.

ackinc: The page linked above is about a global 'origin' property (window.origin)

window.location.origin [seems](https://developer.mozilla.org/en-US/docs/Web/API/Window/location) to be supported in all browsers except Opera

larsjohnsen: Opera is based on Chromium, so generally will code which works on Chrome also work on Opera. In this use case it does not seem to matter whether `origin` or `href` is used, though.

ackinc: It was interesting to see though that location.origin always returns undefined in Opera. Leaving the PR as-is would have broken go-mode for our Opera users.

I'll be more careful about cross-browser incompatibility issues for future PRs. Thanks to mc10 for raising the issue.

mc10:
I have not tested this feature; I can probably get around to doing so tomorrow.

lib/modules/keyboardNav.js
```diff
@@ -789,7 +789,7 @@ module.options = {
 		title: 'keyboardNavsSubredditFrontPageTitle',
 		callback() {
 			const sub = currentSubreddit();
-			if (sub) navigateTo(false, `/r/${sub}`);
+			if (sub) navigateTo(`/r/${sub}`, { leaveNP: true });
```
mc10: This is the only shortcut where I might consider setting `leaveNP` to `false`, since we're still browsing the same subreddit.

PR #4925, opened 2018-11-16
Flair emojis as text option #4924

### LINKED ISSUE #4924: Optionally convert redesign flair emoji to text
https://www.reddit.com/r/Enhancement/comments/9xmmg7/force_user_flair_to_be_text_only/ Ctrl-f "css-only"

In my opinion, this should be part of the "Style Tweaks" module, and should be gated by a "Flair Emoji as Text"  option with enum value like one of "When subreddit style is disabled" (default), "Always", "Never".    (By the way, RES adds a CSS class `body.res-srstyle-disabled`.)

overdodactyl:
Relevant issue: fixes #4924
Relevant Reddit post:  https://old.reddit.com/r/Enhancement/comments/9xmmg7/force_user_flair_to_be_text_only/
Tested in browser: Firefox (63.0.3)

Option was placed in the Style Tweaks module:

screen shot 2018-11-16 at 3 38 45 pm

Before and after example from r/soccer:

before
after

### DIFF
--- a/lib/css/modules/_styleTweaks.scss
+++ b/lib/css/modules/_styleTweaks.scss
@@ -152,7 +152,7 @@ blockquote.twitter-tweet {
 		width: auto;
 		vertical-align: initial;
 	}
-	
+
 	.flairemoji::after {
 		content: attr(title);
 	}
--- a/lib/css/modules/_styleTweaks.scss
+++ b/lib/css/modules/_styleTweaks.scss
@@ -146,13 +146,13 @@ blockquote.twitter-tweet {
 }
 
 .res-styleTweaks-flairEmojiAsText-always,
-.res-styleTweaks-flairEmojiAsText-nosubstyle body.res-srstyle-disabled {
-    .flairemoji[style]  {
-        background-image: none !important;
-        width: auto;
-        vertical-align: initial;
-    }
-    .flairemoji:after {
-        content: attr(title);
-    }
+.res-styleTweaks-flairEmojiAsText-nosubstyle.res-srstyle-disabled {
+	.flairemoji[style]  {
+		background-image: none !important;
+		width: auto;
+		vertical-align: initial;
+	}
+	.flairemoji:after {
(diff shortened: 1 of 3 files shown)


lib/css/modules/_styleTweaks.scss
```diff
@@ -144,3 +144,15 @@ blockquote.twitter-tweet {
 		display: none;
 	}
 }
+
+.res-styleTweaks-flairEmojiAsText-always,
+.res-styleTweaks-flairEmojiAsText-nosubstyle body.res-srstyle-disabled {
```
andytuba: You can squish this together, since they're both listed inside `<body class="...">`

```css
.res-styleTweaks-flairEmojiAsText-nosubstyle.res-srstyle-disabled
```

lib/modules/styleTweaks.js
```diff
@@ -171,6 +171,23 @@ module.options = {
 		description: 'styleTweaksHideDomainLinkDesc',
 		bodyClass: true,
 	},
+	flairEmojiAsText: {
+		title: 'styleTweaksflairEmojiAsText',
+		description: 'styleTweaksflairEmojiAsTextDesc',
+		type: 'enum',
+		value: 'never',
+		values: [{
+			name: 'Never',
```
andytuba: Could you also move the `name` values into `en.json` with (gnarly long) keys like `styleTweaksflairEmojiAsText`?  Then you can load them in here as `i18n('styleTweaksflairEmojiAsTextNever')`

andytuba: Oh, yes, this sounds right.

erikdesjardins:
Tested in Chrome 78. Some subreddits (like /r/soccer) slaughter the default flair CSS, and then this only works with subreddit style disabled. But I don't think there's a good way to solve that in general, and it can be fixed with CSS snippets for specific cases.

erikdesjardins:
Thanks! Sorry about the delay

PR #4926, opened 2018-11-17
Fixes relatively addressed anchor links causing page refreshes

corylulu:
Reddit seems to have introduced some link handling that causes relatively addressed anchor urls like:
`[Link To RES Setting](#res:settings/showImages/autoplayVideo)`
to invoke a page refresh instead of just jumping to the anchor or opening the RES settings.

In addition, this change also makes all #res:settings links open on the current page, even if it's absolute path isn't the same.

Doing this in the `showImages` module might not be an ideal spot for it, but links are already being monitored and processed at that point, so it limits the required overhead.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -908,6 +908,16 @@ function shouldLinkBeDeferred(element) {
 const inText = element => !!element.closest('.md, .search-result-footer');
 
 async function checkElementForMedia(element) {
+	// Prevents relatively addressed anchor link urls from refreshing the page,
+	// while also opening all #res:settings urls on the current page
+	let hash = /^#.*|#res:settings.*/gi.exec(element.getAttribute('href'));
+	if (hash) {
+		element.addEventListener('click', function(e) {
+			e.preventDefault();
+			document.location.href = hash[0];
+		});
+	}
+
 	if (shouldLinkBeDeferred(element)) {
 		await new Promise(resolve => deferredLinks.set(element, resolve));
 		deferredLinks.delete(element);
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -910,9 +910,9 @@ const inText = element => !!element.closest('.md, .search-result-footer');
 async function checkElementForMedia(element) {
 	// Prevents relatively addressed anchor link urls from refreshing the page,
 	// while also opening all #res:settings urls on the current page
-	let hash = /^#.*|#res:settings.*/gi.exec(element.getAttribute('href'));
+	const hash = (/^#.*|#res:settings.*/gi).exec(element.getAttribute('href'));
 	if (hash) {
-		element.addEventListener('click', function(e) {
+		element.addEventListener('click', e => {
 			e.preventDefault();
 			document.location.href = hash[0];
 		});


lib/modules/showImages.js
```diff
@@ -908,6 +908,16 @@ function shouldLinkBeDeferred(element) {
 const inText = element => !!element.closest('.md, .search-result-footer');
 
 async function checkElementForMedia(element) {
+	// Prevents relatively addressed anchor link urls from refreshing the page,
```
andytuba: Could you move this to the settings module which handles these clicks? If `showImages` is disabled, then this won't run.  The overhead won't be that bad.

lib/modules/showImages.js
```diff
@@ -908,6 +908,16 @@ function shouldLinkBeDeferred(element) {
...
+	// while also opening all #res:settings urls on the current page
+	const hash = (/^#.*|#res:settings.*/gi).exec(element.getAttribute('href'));
```
andytuba: It would be clearer to test this with something like `linkElement.hash.startsWith('#res:settings')`

lib/modules/showImages.js
```diff
@@ -908,6 +908,16 @@ function shouldLinkBeDeferred(element) {
...
+	if (hash) {
+		element.addEventListener('click', e => {
+			e.preventDefault();
```
andytuba: If the user middle-clicks, right-clicks, ctrl-clicks, etc. this function shouldn't run.

lib/modules/showImages.js
corylulu: The regex is doing a couple of addition checks here, though.  

The expression is basically  
`if ((element.getAttribute('href').startsWith("#") || element.hash.toLowerCase().startsWith("#res:settings"))`

**Note:** _For some reason, the `element.href` always returns a fully qualified link, even when the actual `href` attribute is only "#anchorLink", so I needed to compare it to the ".getAttribute('href')" to know if it's a fully qualified link or just an anchor link by itself. The fully qualified links are not meant to be handled unless it's a #res:settings link._ 

So if you guys just prefer less regex's, I can use that, but if you just assumed it was doing the equivalent to `linkElement.hash.startsWith('#res:settings')`, perhaps you might prefer the regex since it's doing a few things checks at once.

lib/modules/showImages.js
corylulu: You're right, the ctrl-click would run, incorrectly, I'll fix that. Middle and Right clicks won't fire the event though, so those will work.

corylulu:
@andytuba 
Old Reddit is where I've seen the issue. 

Previous to the change: 
https://gfycat.com/HairyThisEelelephant
(notice the page fully refreshes when clicking the link)

Changed:
https://gfycat.com/QueasyThoughtfulGadwall

larsjohnsen:
I've included this fix in #4859 now.

https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/13bfbc3588d81781885941f885d861a505fb8d75/lib/modules/settingsNavigation.js#L66-L73

corylulu:
@larsjohnsen Great, I wasn't sure if Wiki anchor links were handled the same way, but I guess they aren't, so it's fine to only handled the #res:settings links.

larsjohnsen:
Implemented in #4859. Thanks, @corylulu

PR #4975, opened 2018-11-20
On-save-settings hook

ackinc:
Relevant issue: NA

This PR adds a module-level hook (onSaveSettings) that runs when the module's options are changes in RES settings console

### DIFF
--- a/lib/core/module.js
+++ b/lib/core/module.js
@@ -22,6 +22,7 @@ export class Module<RawOpt: { [string]: any }, Opt: { [string]: ModuleOption<Raw
 	exclude: Array<PageType | AppType | RegExp> = [];
 	shouldRun: () => boolean = () => true;
 	onToggle: (enabling: boolean) => void = () => {};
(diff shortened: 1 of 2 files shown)


lib/core/module.js
```diff
@@ -22,6 +22,7 @@ export class Module<RawOpt: { [string]: any }, Opt: { [string]: ModuleOption<Raw
 	exclude: Array<PageType | AppType | RegExp> = [];
 	shouldRun: () => boolean = () => true;
 	onToggle: (enabling: boolean) => void = () => {};
+	onSaveSettings: (savedSettings: any) => void = () => {}; // savedSettings will only contain options that were just changed
```
andytuba: How about renaming `savedSettings` to `changedSettings`, and dropping the comment?

ackinc: Well. Why didn't I think of that? Change coming up

andytuba:
This will be useful for a variety of workflows:

* refreshing UI elements or data that was set up on pageload
* multicasting updates to other tabs
* showing usage instructions
etc.

PR #4976, opened 2018-11-20
Added a pre-push hook that will run flow, lint, and test scripts

ackinc:
Relevant issue: NA
Tested in browser: NA

Runs the flow, lint, and unit test commands before every `git push`
Mainly useful for WIP PRs - simple errors will now be caught locally instead of in the CI stage

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,11 @@
       "@babel/register"
     ]
   },
+  "husky": {
+    "hooks": {
+      "pre-push": "npm run flow && npm run lint && npm test"
+    }
+  },
   "nyc": {
     "all": true,
     "exclude": [
@@ -108,6 +113,7 @@
     "firefox-extension-deploy": "1.1.1",
     "flow-bin": "0.82.0",
     "html-loader": "0.5.5",
+    "husky": "^1.1.4",
     "inert-entry-webpack-plugin": "4.0.2",
     "interpolate-loader": "2.0.0",
     "jscodeshift": "0.5.1",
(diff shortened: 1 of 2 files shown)


andytuba:
Haven't read the docs, but I assume husky will add a prepush git hook when it's installed by npm/yarn?

ackinc:
Yes; and husky can add other hooks too - pre-commit etc. - depending on what we put in package.json

ackinc:
I don't think the hook is added when husky is installed, since we're able to change the hook (in package.json or the husky config file) long after installation.

I'll try to dig into the library a bit to see how it works

ackinc:
Looked into it - husky does add git hooks (all hooks) at installation time

package.json
```diff
@@ -108,6 +113,7 @@
     "firefox-extension-deploy": "1.1.1",
     "flow-bin": "0.82.0",
     "html-loader": "0.5.5",
+    "husky": "^1.1.4",
```
mc10: This should be a fixed version.

ackinc: I'll make that change. Does it also make sense to add a line to CONTRIBUTING.md, at the top-level or as a sub-section under the 'Adding New Files' section, saying something like "When adding a new dependency or devDependency, make sure to fix the version in package.json"?

andytuba: We should make a linter that runs in CI and pre push :grin:

andytuba: And also document it to contributing.md

Issue #1740, opened 2014-11-01
"add a subreddit" on /me/m/{{multi}} sidebar layered on top of expando pics


jewel-andraia:
https://www.reddit.com/r/RESissues/comments/2kz8da/bug_add_a_subreddit_box_on_multis_hovers_over/

![](http://i.gyazo.com/6e53b5c9bd5ac2e8bee7a02b46aed109.png)
![screen shot 2014-11-01 at 11 47 30 am](https://cloud.githubusercontent.com/assets/455632/4872902/8b459c76-61f7-11e4-969e-3c559aa0f848.png)
## What's causing it?

reddit's CSS rule:

```
#sr-autocomplete-area { 
    position: relative;
    z-index: 100;
}
```
## But why?

Why `z-index: 100`? Protection against other nearby elements?
## Potential fix

RES overrides that with `.side #sr-autocomplete-area  { z-index: auto; }`

The dropdown appears to work fine still.

andytuba:
Is this still a problem?

erikdesjardins:
Appears to be an issue; when in nightmode it's set to `z-index: 1` for...some reason, which fixes it in that situation.

PR #4985, opened 2018-12-01
Fixed "add subreddit" control overlay on top of expanded images

jaysc:
It's a simple fix to a bug that keeps annoying me.
z-index: 1 is chosen to match night mode

Relevant issue: #1740
Tested in browser: Chrome


larsjohnsen:
Thanks!

PR #4986, opened 2018-12-02
v5.13.8

andytuba:


package.json
```diff
@@ -58,6 +58,7 @@
     ]
   },
   "dependencies": {
+    "dompurify": "^1.0.8",
```
mc10: Can you remove the `^` here?

andytuba: Wouldn't it be ironic if RES suffered a vulnerability thanks to an invisible upgrade to a compromised version of a security package.

I added the integrity sha to yarn.lock, too.

andytuba:
I squashed in @larsjohnsen's improvements to v5.13.3.md on that force-push (rebase).

PR #4987, opened 2018-12-02
Remove defunct hosts: vidlit, oddshot

larsjohnsen:


mc10:
At a brief glance it seems like iloopit is still around?

BenMcGarry:
> At a brief glance it seems like iloopit is still around?

Was it always what is it now? Looks like it might have changed.

mc10:
I think the URL format is still the same, and it [looks like it's still being used (NSFW)](https://www.reddit.com/domain/iloopit.net/).

larsjohnsen:
You're right, @mc10. Seems that Loopit is only broken on the test URL.

PR #5003, opened 2018-12-12
Make permission 'history' optional

### LINKED ISSUE #4769: /u/andytuba on 'Read history' permission enabled by default when we install this extension
<div class="md">
<p>!GitHub Move "read history" permission to optional permissions</p>
</div>
<br>
<br>
https://ift.tt/2IloTRC

larsjohnsen:
- The permission must be granted when opening an (by default non-selftext) expando or using the visited filter.
- The `markVisited` option is simplified to a simple on/off, to make it easy to disable the feature from the notification.

peek 2018-12-12 01-05

Closes #4769


honestbleeps:
I think I agree on that one! Unfortunately there's just no good way to
message this without the fearmongering tone of Chrome's dialog - which,
while technically true, doesn't reflect what RES is actually doing. :-\

diswardeepak:
> I realize now that this only changes the declared permission "Read **and change** your browsing history" to "Read your browsing history", which may not be worthwhile for the added trouble this makes.

I think it's worthwhile

PR #5009, opened 2018-12-17
Fixed dependency's version in package.json

ackinc:
Relevant issue: NA
Tested in browser: NA

Based on suggestion here: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/4976#discussion_r242148344


andytuba:
Thanks for jumping on that :smile:

PR #5012, opened 2018-12-22
hosts gfycat: Add support for non-muted videos

larsjohnsen:
https://www.reddit.com/r/Enhancement/comments/a8ij8d/gfycats_now_have_a_sound_option/

Tested in browser: Chrome 71


uppusaikiran:
@larsjohnsen  To see this Update reflected in Release, how many days we should wait ?
Is there any way to custom build the chrome extension ?

BenMcGarry:
@uppusaikiran We release when we are happy with the updates within the release and stability. We do not guarantee release dates or have ETAs.

PR #5019, opened 2019-01-10
Fix for faviconUseLegacy

corylulu:
They updated the static url for the old favicon, so I just replaced it with a base64 string instead to avoid the issue in the future.

Relevant issue: https://www.reddit.com/r/RESissues/comments/ae9rv6/legacy_favicon_not_working/
Tested in browser: Chrome


larsjohnsen:
To avoid having it inlined in the source, could you save the icon in `/lib/images/`, like is done for the Giphy logo?
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/57a3fffd19e8a7f0b748da0d02c0a65909cd5ad9/lib/modules/hosts/giphy.js#L5

larsjohnsen:
Added in #5040

corylulu:
Oh, mb, forgot about this. Thanks for adding. 👍

PR #5049, opened 2019-01-23
Host privacy policy

### LINKED ISSUE #4983: Media host privacy policy URLs
In the interest of fairness to RES users and all the platforms RES draws on, the RES team should add options to each media host linking to their privacy policy and noting "telemetry may be collected by this host for analytics."  

This could be implemented as simply adding for each media host `{hostName}privacyPolicy`: button option: (Read privacy Policy) - This media host may be collecting analytical data on user activity while viewing their content.

Atrolantra:
Relevant issue: fixes #4983 
Relevant Reddit post: N.A.
Tested in browser: Firefox (64.0.2)

### DIFF
--- a/lib/modules/hosts/aarli.js
+++ b/lib/modules/hosts/aarli.js
@@ -8,6 +8,15 @@ export default new Host('aarli', {
 	name: 'AAR Library',
 	domains: ['aar.li'],
 	logo: 'https://aar.li/favicon.ico',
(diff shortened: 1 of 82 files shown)


larsjohnsen:
I think having a button after host could be a bit noisy. What do you think about integrating it with the toggle to look something like this?
image

PR #5057, opened 2019-01-24
Fix 3rd party cookie blocking issue on Firefox [Externally blocked]

larsjohnsen:
Firefox 66 (nightly) [includes a fix](https://bugzilla.mozilla.org/show_bug.cgi?id=1509112) which allows users who have "Block all third-party cookies" enabled, to make requests to (manifest) permitted URLs from the background. Those requests includes cookies, and as such makes it unnecessary for the user to whitelist third-party domains to be able to use neverEndingReddit etc.

- [ ] Since the fix won't be LTS before in July, requests must still, by default, be sent in foreground.  A way to check whether cookies have been blocked, and only then try the background, is necessary.
- [ ] A fallback method may also be used to work around uMatrix etc blocking requests

Tested in browser: Firefox 66 (works), Firefox 65 (doesn't work)

### DIFF
--- a/lib/environment/background/ajax.js
+++ b/lib/environment/background/ajax.js
@@ -3,18 +3,16 @@
 import _ from 'lodash';
 import { addListener } from './messaging';
 
-addListener('ajax', async ({ method, url, headers, data, credentials }) => {
-	const rawResponse = await fetch(url, {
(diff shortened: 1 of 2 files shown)


larsjohnsen:
Cloning this since the workaround is no longer feasible.

Issue #5046, opened 2019-01-22
Media download breaks on special characters


larsjohnsen:
"Save button not working when certain characters are in the post title, such as ? and "

https://www.reddit.com/r/RESissues/comments/ai1d0d/cannot_use_save_button_on_artstation_images/eeps61u/

PR #5060, opened 2019-01-27
Fix broken media download

kdreagle:
Relevant issue: [#5046](https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/5046)
Tested in browser: Chrome 71

This fixes 2 things that broke the download media button:
1. Some media url's had a query after the file name, for example 'picture.jpg?12345'. This fix will strip off the query.
2. Some post titles had illegal filename characters. This fix removes those characters from the filename.


larsjohnsen:
Thanks!

PR #5062, opened 2019-01-27
Filter list optimization

larsjohnsen:
The time spent creating and evaluating a list filter of 40000 users on 100 posts is reduced from multiple seconds to a few hundreds ms.

- Reconciles entries of same type at an earlier stage, to avoid creating a `Case` object for every entry
- Class `PatternCase` which extends `Case`, since many cases have a lot in common
- By deferring the string-to-regex build, it is easier to build more more performent regexes
  - The resulting `/^(a|b|c)$/` is must faster than the previous `/^a$|^b$|^c$/`

Tested in browser:  Chrome 71, Firefox 65

### DIFF
--- a/lib/modules/filteReddit.js
+++ b/lib/modules/filteReddit.js
@@ -780,7 +780,15 @@ export async function saveFilterlineStateAsDefault(type: string) {
 }
 
 function getStringMatchConditions(source, caseType, additionalCriteria) {
-	const [matchString = '', applyTo = 'everywhere', applyList = '', except = ''] = Array.isArray(source) ? source : [source];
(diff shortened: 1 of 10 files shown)


mc10:
Is it possible to add a few tests to check that the generated regexes are what we expect? This will probably be helpful to avoid regressions.

larsjohnsen:
@mc10 There are some [tests](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/0fd11ba613d53c11cc03191d900100ef46cec9ee/tests/filteReddit.js#L178) which I think will catch most issues.

PR #5065, opened 2019-01-31
Add peertube

aliceinwire:
Adding peertube

### DIFF
--- a/lib/modules/hosts/peertube.js
+++ b/lib/modules/hosts/peertube.js
@@ -29,15 +29,13 @@ export default new Host('peertube', {
 		'framatube.org',
 	],
 	attribution: false,
-	detect: ({ pathname, hostname }) => {
-		// split path excluding first /
-		const split = pathname.substring(1).split('/');
-		if (split[0] === 'videos') {
-			return [split[2], hostname];
+	detect: ({ hostname, pathname }) => {
+		const [, route, , id] = pathname.split('/');
+		if (route === 'videos') {
+			return [hostname, id];
 		}
 	},
-
-	handleLink(href, [id, hostname]) {
+	handleLink(href, [hostname, id]) {
 		const embed = `https://${hostname}/videos/embed/${id}`;
 		return {
 			type: 'IFRAME',
--- a/lib/modules/hosts/peertube.js
+++ b/lib/modules/hosts/peertube.js
@@ -2,7 +2,6 @@
 
 import { Host } from '../../core/host';
 
-
 export default new Host('peertube', {
 	name: 'peertube',
 	domains: [
@@ -48,4 +47,3 @@ export default new Host('peertube', {
 		};
 	},
 });
-
--- a/lib/modules/hosts/peertube.js
+++ b/lib/modules/hosts/peertube.js
@@ -1,0 +1,51 @@
+/* @flow */
+
+import { Host } from '../../core/host';
+
+
+export default new Host('peertube', {
+	name: 'peertube',
+	domains: [
+		'peervideo.net',
+		'peertube.social',
+		'peertube.mastodon.host',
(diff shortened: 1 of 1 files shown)


aliceinwire:
working screenshot:
https://screenshots.firefox.com/BA6L2Ttj2hgCtXUH/www.reddit.com

aliceinwire:
Would be nice to add a setting in the UI for adding new domain for Peertube.
But I have no idea if is possible or accepted.
Anyway it will probably be in a next pull request, as I'm already enough out of my competence sphere.

lib/core/migrate/migrate.js
```diff
@@ -372,6 +372,7 @@ const migrations = [
 			await Migrators.moveOption('showImages', 'display memedad', 'showImages', 'display_memedad');
 			await Migrators.moveOption('showImages', 'display ridewithgps', 'showImages', 'display_ridewithgps');
 			await Migrators.moveOption('showImages', 'display photobucket', 'showImages', 'display_photobucket');
+			await Migrators.moveOption('showImages', 'display peertube', 'showImages', 'display_peertube');
```
larsjohnsen: @BenMcGarry, yeah, the option `display_peertube` is automatically generated, so this isn't necessary.

lib/modules/hosts/peertube.js
```diff
@@ -0,0 +1,49 @@
...
+		'watching.cypherpunk.observer',
+		'queertube.org',
+		'exode.me',
+		'framatube.org',
+	],
+	attribution: false,
+	detect: ({ pathname, hostname }) => {
+		// split path excluding first /
+		const split = pathname.substring(1).split('/');
+		return [split[2], hostname];
```
larsjohnsen: Are any way to check that a given link points to a video, to avoid that a link to `https://peertube.openstreetmap.fr/about` generates an expando?

aliceinwire: oh yes.
sending update

lib/core/migrate/migrate.js
aliceinwire: ok thanks
I have to remove all the line ? is it correct ?

aliceinwire:
tested here:
https://screenshots.firefox.com/51zDsLk99btKVqYA/www.reddit.com

SoniEx2:
any reason this uses a domain whitelist?

Nutomic:
Maybe you could check for the [Peertube meta tag](https://github.com/Chocobozzz/PeerTube/pull/1591) instead of hardcoding an instance list? However it is not released yet.

SoniEx2:
it's actually possible to detect the URL layouts of peertube and the peertube API. altho the tag does make it easier I guess.

larsjohnsen:
@Nutomic That would require loading all links, correct? I think that is infeasible, both due to bandwidth and privacy concerns.

SoniEx2:
Hm. p2p-sourced link data?

bjuergens:
This may be a little off-topic, but shouldn't the og-tags always be loaded for unkown urls? After all `og:video` and `og:image` exists exactly for situations like this.

aliceinwire:
@SoniEx2 @Nutomic the problem is doing it from just the URL.
If you have a better way please send fix commit here.

SoniEx2:
that pissed me off with youtube-dl and it pisses me off here too. altho youtube-dl downloads the full page anyway so they don't have an excuse for it. does RES not?

larsjohnsen:
Thanks, @aliceinwire!

aliceinwire:
would be nice to use a peertube instance list like https://fediverse.network/about
unfortunatly their api is not ready yet

SoniEx2:
I don't think that handles non-federating instances

PR #5075, opened 2019-02-13
Replace momentjs with (much leaner) dayjs

larsjohnsen:
Dayjs is modular and extendable. By being selective with what methods and locales are imported, we get a dependency that is less than 20 kB (in contrast to momentjs' 300 kB)

Tested in browser: Chrome 73

### DIFF
--- a/lib/utils/localization.js
+++ b/lib/utils/localization.js
@@ -1,24 +1,44 @@
 /* @flow */
 
-import moment from 'moment';
+import _ from 'lodash';
 import { locale } from '../environment';
+
+// Avoid initializing dayjs before it's necessary
+const Dayjs = _.once(() => {
+	require('dayjs/locale/de'); // eslint-disable-line global-require
+	require('dayjs/locale/el'); // eslint-disable-line global-require
+	require('dayjs/locale/es'); // eslint-disable-line global-require
+	require('dayjs/locale/he'); // eslint-disable-line global-require
+	require('dayjs/locale/it'); // eslint-disable-line global-require
+	require('dayjs/locale/nl'); // eslint-disable-line global-require
+	require('dayjs/locale/pl'); // eslint-disable-line global-require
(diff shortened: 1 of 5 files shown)


xxyuk:
Can't personally vouch for it but have read about `Day.Js` as an alternative

⏰ Day.js 2KB immutable date library alternative to Moment.js with the same modern API

https://github.com/iamkun/dayjs

reference: [moment/moment#2373 (comment)](https://github.com/moment/moment/issues/2373#issuecomment-406087130)

larsjohnsen:
Thanks for the suggestion, @xxyuk. That seems to fit our use case well!

lib/utils/localization.js
```diff
@@ -1,24 +1,44 @@
...
+	require('dayjs/locale/he'); // eslint-disable-line global-require
+	require('dayjs/locale/it'); // eslint-disable-line global-require
+	require('dayjs/locale/nl'); // eslint-disable-line global-require
+	require('dayjs/locale/pl'); // eslint-disable-line global-require
+	require('dayjs/locale/pt-br'); // eslint-disable-line global-require
+	require('dayjs/locale/pt'); // eslint-disable-line global-require
+	const dayjs = require('dayjs'); // eslint-disable-line global-require
+	dayjs.extend(require('dayjs/plugin/relativeTime')); // eslint-disable-line global-require
+	dayjs.extend(require('dayjs/plugin/localizedFormat')); // eslint-disable-line global-require
+	dayjs.locale(locale());
```
kevinji: Hmm, since we can get the locale via `locale`, maybe we can replace all of the `require` lines with a single <code>require(&#96;dayjs/locale/${locale()}&#96;);</code> call?

larsjohnsen: I don't think Webpack will be able to resolve that without also bundling all locales.

PR #5076, opened 2019-02-13
Replace jquery-sortable with sortablejs

larsjohnsen:
jquery-sortable is broken in the settings console, and is no longer maintained. Sortablejs appear to work fine with minimal configuration.

- [x] Fix issue in Firefox where it does not initialize

Tested in browser: Chrome 73, Firefox 65

### DIFF
--- a/lib/css/res.scss
+++ b/lib/css/res.scss
@@ -511,22 +511,8 @@ span.multi-count {
 	background-color: lightblue;
 }
 
-.handle,
-.dragging {
+.handle {
 	cursor: grab !important;
-}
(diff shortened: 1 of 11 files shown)


erikdesjardins:
After updating and rebasing it seems to work in Chrome 78 and Firefox 68, at least in the settings console. Do you remember where it wasn't working in FF before?

larsjohnsen:
The issue was apparent on the dashboard and in Filterline complex groups. I did update it to the latest version now, and everything has started working fine however.

PR #5077, opened 2019-02-17
storage: Remove PrefixWrapper#getMultiple(Nullable)

larsjohnsen:
Loading multiple items together can be so slow that RES doesn't initialize (adding option body classes, custom styles etc) before Firefox decides to paint the first frame, causing a FOUC. Benchmarking this, the total time spent loading from storage is a little (about one third) larger, but the removal of complexity and said Firefox issue makes this worth it.

Tested in browser: Firefox 66, Chrome 74


larsjohnsen:
I've merged the most important part of this PR (the option fetching), and the rest probably is not worthwhile to change due to lowered performance.

PR #5078, opened 2019-02-20
Redesign: Implement loggedInUser

larsjohnsen:
The first text node in the user button contains the username.

Tested in browser: Chrome 74

### DIFF
--- a/lib/utils/user.js
+++ b/lib/utils/user.js
@@ -4,7 +4,7 @@ import _ from 'lodash';
 import { ajax } from '../environment';
 import type { CurrentRedditUser, RedditAccount } from '../types/reddit';
 import { MINUTE } from './time';
-import { regexes } from './location';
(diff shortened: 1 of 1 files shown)


lib/utils/user.js
```diff
@@ -18,6 +18,14 @@ export const isLoggedIn = _.once(() => {
 export const loggedInUser = _.once((): string | void => documentLoggedInUser(document));
 
 export const documentLoggedInUser = (document: Document | Element): string | void => {
+	if (isAppType('d2x')) {
+		// The first text node in the user button contains the username
+		const findFirstTextNode = e => [...e.childNodes].filter(v => v.nodeType === 3).concat(...[...e.children].map(findFirstTextNode));
```
andytuba: this seems useful to throw in a "utils" / "library" function

PR #5087, opened 2019-03-12
Prevent duplication of user tags when navigating posts

### LINKED ISSUE #5068: User tagger appears multiple times for postAuthor in lightbox when N/P
1. https://new.reddit.com
2. Click the "comments" button on a post.
3. Observe a usertag button to the right of the author link, above the post.
4. Press `N` on your keyboard to open the next post, keeping the lightbox open.

Expected: only one user tag button (user tag is updated or replaced)
Actual: tags accumulate

This is probably because React is reusing the postAuthor JSAPI container and not wiping the contents. RES should be aware that a container may get reused and update it accordingly. RES should probably remove any existing user tags in the JSAPI container before adding a new user tag.

prakhar1912:
*Before*
Before

*After*
UserTagger

Relevant issue: fixes #5068
Tested in browser: Chrome 72, Firefox 65

### DIFF
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -146,11 +146,10 @@ module.beforeLoad = () => {
 
 	watchForRedditEvents('postAuthor', (element, { author, _: { update } }) => {
 		if (update) return;
-		// Remove the previously added tags
-		const previousTags = element.getElementsByClassName('RESUserTag');
-		if (previousTags.length) {
-			for (const tag of previousTags) tag.remove();
-		}
+
+		// Remove previously added tags, as Reddit may not empty the container when navigating posts
+		for (const tag of element.getElementsByClassName('RESUserTag')) tag.remove();
+
 		applyToUser(element, { username: author });
 	});
 	watchForRedditEvents('commentAuthor', (element, { author, _: { update } }) => {
--- a/lib/modules/userTagger.js
+++ b/lib/modules/userTagger.js
@@ -146,6 +146,11 @@ module.beforeLoad = () => {
 
 	watchForRedditEvents('postAuthor', (element, { author, _: { update } }) => {
 		if (update) return;
+		// Remove the previously added tags
+		const previousTags = element.getElementsByClassName('RESUserTag');
+		if (previousTags.length) {
+			for (const tag of previousTags) tag.remove();
+		}
 		applyToUser(element, { username: author });
 	});
 	watchForRedditEvents('commentAuthor', (element, { author, _: { update } }) => {


lib/modules/userTagger.js
```diff
@@ -381,6 +381,8 @@ export class Tag {
 			// `defaultTagElement` is a lot fast than `buildTagElement`
 			instance.tagger = (this.text || this.color) ? Tag.buildTagElement(this) : Tag.defaultTagElement();
 			instance.tagger.addEventListener('click', () => this.openPrompt(instance));
+			// Remove the previously added tags
+			instance.element.innerHTML = '';
```
andytuba: Instead of "blanking out" the element, it would be better to remove the element entirely. e.g. `instance.element.parentNode.removeElement(instance.element)`

andytuba: That said, this doesn't feel quite right .. I assumed that a _previous_ instance's element is sticking around, and that's what needs to be removed?

prakhar1912: Yeah, you are right. The previous instance's element has a tag that is not removed when a new post is loaded.

prakhar1912: @andytuba Removing the element won't work because every instance refers to the same container (`instance.element`) to insert the tag. I tried removing the element, and the tags did not show up. Please advise.

andytuba: Sorry, I should have said "some other instance's `.tagger`" element. 

`instance.element` is the JSAPI container, `<span data-name="reddit-enhancement-suite">...</span>`.  All RES's modules share that `instance.element`. So, UserTagger shouldn't clear out the entire `instance.element` because that might clobber content added by other RES modules.  Instead, UserTagger should only remove `.tagger` elements from this `instance.element`

prakhar1912: Turns out all instances are actually appending children in the same `<span data-name="reddit-enhancement-suite">...</span>`, because every instance's element points to it.

*User tags repeated inside the span*
Screenshot from 2019-03-13 22-39-13

andytuba: Exactly :)

andytuba: So, what I'm hoping you can help with, is to removing the old `.RESUserTag` (`instance.tagger`) when adding a new `tagger`.

prakhar1912: Yeah, so I need to clear the `instance.element` before appending more tags.

`if (instance.tagger) instance.tagger.remove();` : This check already exists

Do keep in mind that the tags being repeated are of different users and not the same user.

andytuba: ... clear `.RESUserTag` elements from `instance.element` before appending more tags.  (Don't clear it out entirely.)

Yeah, that `if instance.tagger` check is checking for "same user".  The extra `.RESUserTag` items are from different users -- that's why they should be removed.

prakhar1912: Okay, got it.

andytuba:
Looks like we're almost there! It looks like the user tagger pop-up stays open when the lightbox changes. Does that automatically go away when the tag is removed, or does that need more work?

andytuba:
Oh, another PR sounds fine.

lib/modules/userTagger.js
```diff
@@ -381,6 +381,11 @@ export class Tag {
 			// `defaultTagElement` is a lot fast than `buildTagElement`
 			instance.tagger = (this.text || this.color) ? Tag.buildTagElement(this) : Tag.defaultTagElement();
 			instance.tagger.addEventListener('click', () => this.openPrompt(instance));
+			// Remove the previously added tags
+			const previousTags = instance.element.getElementsByClassName(instance.tagger.className);
+			if (previousTags.length) {
+				previousTags.forEach(tag => tag.remove());
```
larsjohnsen: Note that this misses the voteWeight element. It will also only remove previous tags added with `append`, not `after` (see beneath).

Since this is special behavior is only necessary for the (buggy?) postAuthor container, it might be cleaner if the workaround is moved up to the `watchForRedditEvents` listener.

prakhar1912: Okay. Thank you for pointing that out. I will make the change.

larsjohnsen:
Thanks!

PR #5089, opened 2019-03-14
Make uncheckSendRepliesToInbox compatible with redesign

### LINKED ISSUE #4956: Submission Helper - uncheck send replies - redesign compatibility
Look for aria-labelledby="send-replies" and figure out how to untick it

prakhar1912:
SendReplyNotifications

Relevant issue: fixes #4956
Tested in browser: Chrome 73, Firefox 65

### DIFF
--- a/lib/modules/submitHelper.js
+++ b/lib/modules/submitHelper.js
@@ -6,8 +6,11 @@ import { Module } from '../core/module';
 import { ajax, i18n } from '../environment';
 import {
 	formatDateDiff,
+	isAppType,
 	isPageType,
 	string,
+	waitForDescendant,
+	waitForEvent,
 } from '../utils';
 import type { RedditListing, RedditLink } from '../types/reddit';
 import * as CommandLine from './commandLine';
@@ -51,7 +54,22 @@ const $repostWarning = _.once(() => $(string.html`
 let urlField, srField;
 
 module.go = () => {
-	if (isPageType('submit') && module.options.warnAlreadySubmitted.value) {
+	implementOptions();
+	// Register submit CLI if not on the submit page
+	if (!isPageType('submit')) {
+		registerCommandLine();
+	}
+};
+
+async function implementOptions() {
+	if (isAppType('d2x')) {
+		waitForEvent(document, 'reddit.urlChanged').then(implementOptions);
+	}
+	if (!isPageType('submit')) {
+		return;
+	}
+	// Implement submitHelper preferences if on the submit page
+	if (module.options.warnAlreadySubmitted.value) {
 		const urlFieldDiv = document.querySelector('#url-field');
 		if (urlFieldDiv) {
 			$(urlFieldDiv).parent().after($repostWarning());
@@ -71,21 +89,18 @@ module.go = () => {
 		}
 	}
 
-	if (isPageType('submit') && module.options.uncheckSendRepliesToInbox.value) {
-		const sendReplies: ?HTMLInputElement = (document.querySelector('#sendreplies'): any);
+	if (module.options.uncheckSendRepliesToInbox.value) {
+		const selector = isAppType('d2x') ? '[aria-labelledby="send-replies"]' : '#sendreplies';
+		const sendReplies: ?HTMLElement = await waitForDescendant(document.documentElement, selector);
 		if (sendReplies) {
-			sendReplies.checked = false;
+			sendReplies.click();
 		}
 	}
 
-	if (isPageType('submit') && module.options.focusFormOnLoad.value) {
+	if (module.options.focusFormOnLoad.value) {
 		$('form.submit [name=url], form.submit [name=title]').filter(':visible').first().focus();
 	}
-
-	if (!isPageType('submit')) {
-		registerCommandLine();
-	}
-};
+}
 
 function registerCommandLine() {
 	const trailingUrl = /(?:\s+(\w+:\/\/.+))$/;
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -45,9 +45,9 @@ export {
 export {
 	EdgeScroll,
 	addCSS,
-	empty,
 	click,
 	elementInViewport,
+	empty,
 	getViewportSize,
 	getHeaderOffset,
 	getD2xBodyOffset,
--- a/lib/utils/location.js
+++ b/lib/utils/location.js
@@ -1,6 +1,7 @@
 /* @flow */
 
 import _ from 'lodash';
+import { waitForEvent } from './dom';
 
 export function matchesPageLocation(includes: Array<PageType | AppType | RegExp>, excludes: Array<PageType | AppType | RegExp> = []): boolean {
 	const [includeStrings, includeRegExps]: any = _.partition(includes, (x: string | RegExp): boolean => typeof x === 'string');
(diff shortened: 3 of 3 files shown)


lib/modules/submitHelper.js
```diff
@@ -51,41 +57,49 @@ const $repostWarning = _.once(() => $(string.html`
...
 		}
-	}
 
-	if (isPageType('submit') && module.options.focusFormOnLoad.value) {
-		$('form.submit [name=url], form.submit [name=title]').filter(':visible').first().focus();
-	}
+		if (module.options.uncheckSendRepliesToInbox.value) {
+			const selector = isAppType('d2x') ? '[aria-labelledby="send-replies"]' : '#sendreplies';
+			const sendReplies: ?HTMLElement = document.querySelector(selector);
+			if (sendReplies) {
```
andytuba: As you mentioned, this will probably be null immediately after `urlChanged` jsapi event fires since Reddit's react component hasn't mounted yet. In this case, we could adopt the selenium approach of "await element".

If RES already had a helper function, I would look in the `dom` utils. If not, here's an idea:

```javascript
function awaitElement(selector, container = document.documentElement, timeout = 5000) {
    const rejectAfter = Date.now() + timeout;
    return new Promise ((resolve, reject => {
        f();
        function f() {
              const element = container.querySelector(selector);
                if (element) {
                     resolve(element);
                 } else if (Date.now() >= rejectAfter) {
                      reject("timeout");
                 } else {
                      setTimeout(f, 100);
                 }
            }
    });
}
```

lib/utils/index.js
```diff
@@ -60,6 +60,7 @@ export {
 	watchForDescendants,
 	watchForFutureDescendants,
 	waitForEvent,
+	findElement,
```
andytuba: could you alphabetize this (and `empty` too)?

prakhar1912: Sure.

lib/modules/submitHelper.js
```diff
@@ -51,41 +58,53 @@ const $repostWarning = _.once(() => $(string.html`
...
+		registerCommandLine();
 	}
+};
 
-	if (isPageType('submit') && module.options.uncheckSendRepliesToInbox.value) {
-		const sendReplies: ?HTMLInputElement = (document.querySelector('#sendreplies'): any);
-		if (sendReplies) {
-			sendReplies.checked = false;
+async function implementOptions() {
+	if (isPageType('submit')) {
```
larsjohnsen: To avoid unnecessary many levels of nesting, return early (`if (!isPageType('submit')) return;`).

lib/modules/submitHelper.js
```diff
@@ -51,41 +58,53 @@ const $repostWarning = _.once(() => $(string.html`
...
 		}
-	}
 
-	if (isPageType('submit') && module.options.focusFormOnLoad.value) {
-		$('form.submit [name=url], form.submit [name=title]').filter(':visible').first().focus();
-	}
+		if (module.options.uncheckSendRepliesToInbox.value) {
+			const selector = isAppType('d2x') ? '[aria-labelledby="send-replies"]' : '#sendreplies';
+			try {
+				const sendReplies: ?HTMLElement = await findElement(selector);
```
larsjohnsen: Also consider using `waitForDescendent`.

lib/utils/location.js
```diff
@@ -75,17 +75,17 @@ const appPageTypes: { [AppType]: {| default?: PageType, pageTypes: PageType[] |}
...
 	d2x: {
-		pageTypes: ['profile2x', 'profileCommentsPage'],
+		pageTypes: ['profile2x', 'profileCommentsPage', 'submit'],
 	},
 	options: {
 		pageTypes: [],
 	},
 };
 
-export const pageType = _.once((): ?PageType => {
```
larsjohnsen: To avoid making this expensive (as `isPageType` is invoked under many circumstances), could `pageType` be made a variable which is updated on `reddit.urlChanged`? For example: 

```
expert let pageType;
(function updatePageType() { 
  pageType =  ...; 
  waitForEvent('reddit.urlChanged').then(updatePageType);
})()
```

lib/modules/submitHelper.js
prakhar1912: Yeah, good point. Thanks!

lib/modules/submitHelper.js
prakhar1912: @larsjohnsen You mean change the function name?

prakhar1912: Tried it. `waitForDescendants` adds another observer, which is causing the Send Reply Notifications toggle to switch whenever the React component for Submit is mounted. So, I think I'll stick to `findElement`, since it only fetches the element once.

lib/utils/location.js
prakhar1912: Yeah, good catch. Thanks!

lib/modules/submitHelper.js
larsjohnsen: You might be confusing `waitForDescendant` and `watchForDescendants`. `waitForDescendant` will stop observing once it has found a match, but it may be a good idea to have it optionally time out.

prakhar1912: Okay, got it. My bad. Thanks!

lib/utils/location.js
prakhar1912: Okay, so I have been trying to get this to work. The code as above works fine in the browser. But, the tests fail. What I found out:
1) jQuery ($) is being imported in `utils/dom.js`, and the error `window is not defined` pops up when `waitForEvent` is imported.

2) If I emulate the functionality in `waitForEvent` in `utils/location.js`, I get this error,
Screenshot from 2019-03-19 02-05-40

I am pushing the state of the code as at point 2. I'd appreciate help in debugging this.

prakhar1912: This will actually break the tests for `utils/location.js` because getting the page type will dive into the DOM. And since the function is an IIFE, it will be run when importing functions from the file.

larsjohnsen: In that case, maybe this pattern could be used: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/e52dd41196b0cbd597d137ed1703230341b886af/lib/utils/dom.js#L142-L143

lib/utils/location.js
```diff
@@ -75,18 +75,27 @@ const appPageTypes: { [AppType]: {| default?: PageType, pageTypes: PageType[] |}
...
+export const pageType = _.memoize((): ?PageType => {
+	// Using waitForEvent from utils/dom.js would throw an error
+	// during the test for this module as $ is defined on
+	// the window in that module. Emulating functionality here
```
larsjohnsen: I didn't realize that. I have removed the $ dependency f78d299d5840281dc92a5c8f48502166952f5c80, so I think `waitForEvent` can be used from `location` now without breaking the test.

lib/utils/location.js
```diff
@@ -75,18 +76,24 @@ const appPageTypes: { [AppType]: {| default?: PageType, pageTypes: PageType[] |}
...
 
-export const pageType = _.once((): ?PageType => {
+export const pageType = _.memoize((): ?PageType => {
+	clearPageTypeCache();
 	const spec = appPageTypes[appType()];
 	return spec.pageTypes.find(pageType => regexes[pageType].test(location.pathname)) || spec.default;
 });
 
+function clearPageTypeCache() {
+	waitForEvent(document, 'reddit.urlChanged').then(clearPageTypeCache);
```
larsjohnsen: Nit: It would be cleaner to reset the cache inlined in the function above, like in the example I provided above.

larsjohnsen:
Thanks!

PR #5090, opened 2019-03-17
Add VLive to media hosts

sporksparks:
VLive is a livestreaming service (with VODs too) for Korean-based celebrities (Kpop, makeup/beauty, etc...)

This should add embed support for VLive videos as an expando. You can test [here](https://www.reddit.com/domain/vlive.tv/) 
Tested in browser: Chrome 72


larsjohnsen:
Thanks!

PR #5091, opened 2019-03-25
Add dashboard iframe on redesign

### LINKED ISSUE #4974: Dashboard - redesign compatibility
If user visits r/dashboard on redesign, then load an iframe with /r/dashboard?embedded=true and fix up the little issues with gear icon rendering weirdly, kill the sidebar properly.

prakhar1912:
Dashboard

Relevant issue: fixes #4974
Tested in browser: Chrome 73

### DIFF
--- a/lib/css/modules/_dashboard.scss
+++ b/lib/css/modules/_dashboard.scss
@@ -388,3 +388,11 @@ ul.widgetStateButtons {
 		&.selected a:first-child { pointer-events: none; }
 	}
 }
+
+iframe.dashboard {
+	position: fixed;
+	z-index: 100;
+	width: 100%;
+	height: 100%;
+	display: none;
+}
--- a/lib/modules/dashboard.js
+++ b/lib/modules/dashboard.js
@@ -96,12 +96,12 @@ module.options = {
 		advanced: true,
 	},
 };
-module.shouldRun = () => isCurrentSubreddit('dashboard');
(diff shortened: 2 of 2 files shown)


larsjohnsen:
I don't think this needs to be embedded as an iframe. It would should suffice to replace `$tabPage` with the an element on the page, and probably update some of the queried URLs so that they loads from `old.reddit.com`.

prakhar1912:
> I don't think this needs to be embedded as an iframe. It should suffice to replace `$tabPage` with an element on the page, and probably update some of the queried URLs so that they loads from `old.reddit.com`.

@

prakhar1912:
> I don't think this needs to be embedded as an iframe. It should suffice to replace `$tabPage` with an element on the page, and probably update some of the queried URLs so that they loads from `old.reddit.com`.

@larsjohnsen Do you mean something like this?
Screenshot from 2019-03-28 02-24-47

larsjohnsen:
Yes. Though the element probably ought to be positioned to take up the whole viewport / lightbox except for header.

andytuba:
hmm, i was thinking iframe so that we could rely on old.reddit.com css/javascript  ... and of course RES would need to actually load the dashboard widgets inside the iframe.

andytuba:
Hmm, I see significant work needed for styling posts in embedded view.  It might be better for RES to build out its own post view with minimally: 

* post title
* byline (posted to subreddit by username at time)
* expando button.

PR #5092, opened 2019-03-26
Close user tagger when navigating between posts

### LINKED ISSUE #5088: User tagger not closing
- My browser: Chrome 73 on Linux
- My RES version: 5.14.5

*What's the problem?*
The User Tagger dialog remains open when navigating between posts.

Bug

*Are there any related posts in https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues or /r/RESissues?* No

prakhar1912:
Peek 2019-03-30 03-26

*Note: The pop up in the GIF is not removed when there are 2 user tags, because the user tag is not removed when navigating posts. PR for that issue is #5087*

Relevant issue: fixes #5088 
Tested in browser: Chrome 73, Firefox 66


larsjohnsen:
This was also changed in 5.14, but was reverted due to a bug in Firefox, in addition to many people  preferring it not automatically closing.

Another solution could be to have the Hover observe the target and close dialog if target is removed, or manually close the hover when removing the element.

prakhar1912:
> This was also changed in 5.14, but was reverted due to a bug in Firefox, in addition to many people preferring it not automatically closing.
> 
> Another solution could be to have `Hover` observe the target element and close dialog if it is removed, or manually close the hover when removing the element.

Okay, I think I will add an observer on the parent.

lib/utils/dom.js
```diff
@@ -412,3 +412,21 @@ function addStyle(css) {
...
+					if (node === el || node.contains(el)) {
+						observer.disconnect();
+						resolve();
+						return;
+					}
+				}
+			}
+		});
+
+		observer.observe(document.body, { subtree: true, childList: true });
```
larsjohnsen: Observing `document.body` would be expensive. Instead observing `ele.parentElement` would probably suffice, and then the `subtree` option can be removed.

lib/modules/hover.js
```diff
@@ -243,6 +250,11 @@ class Hover {
 	begin() {
 		if (!this._enabled) return false;
 
+		if (this._options.closeOnTargetUnmount) {
+			const target = this.getCheckedTarget();
+			waitForRemoval(target).then(() => this.close());
```
larsjohnsen: Since `target` is mutable, `waitForRemoval` may resolve after `target` has changed.

lib/modules/hover.js
```diff
@@ -69,6 +75,7 @@ module.beforeLoad = () => {
 		fadeSpeed: parseFloat(module.options.fadeSpeed.value),
 		width: parseFloat(module.options.width.value),
 		closeOnMouseOut: module.options.closeOnMouseOut.value,
+		closeOnTargetUnmount: module.options.closeOnMouseOut.value,
```
larsjohnsen: I don't see a need to have an option for this.

prakhar1912: I did that because of the `if` check.

lib/utils/dom.js
prakhar1912: `parentElement` does not actually work.  This is because the JSAPI container is not removed when navigating between posts. I did try it before `document.body`.

larsjohnsen: It am not uncertain why it would not work in this case. Since `target` is not moved (but is at all times the child of the same JSAPI container), it being manually removed in `userTagger` should trigger the observer when observing `target.parentElement`.

lib/modules/hover.js
larsjohnsen: What I mean here, is this hover instance can be triggered from another tag icon and mutate the `target` variable. This creates an additional observer, this time for for the new `target`. If the old `target` element then were to removed (for whatever some reason), the new hover instance would be closed.

lib/modules/hover.js
larsjohnsen: We probably don't need that conditional, since it seems useful to apply this everywhere.

lib/utils/dom.js
prakhar1912: `target.parentElement` does work, but the `MutationObserver` fires only once. Looking into why it does that.

lib/modules/hover.js
prakhar1912: Okay, got it. Will a check like
```js
() => target === this.getCheckedTarget() && this.close()
```
be better?

larsjohnsen: Yeah, that should be sufficent.

lib/utils/dom.js
prakhar1912: Update: `target.parentElement` issue could not be resolved. Listening on `document.body` is the only element that fires the MutationObserver correctly.

prakhar1912: @larsjohnsen Thoughts on what I could be missing?

larsjohnsen: Using `target.parentElement` appears to work in the instances where the old tag icon is manually removed (#5087). However, when Reddit generates a new JSAPI container (which it seemingly does at random) only `document.body` will catch that.

Since we are stuck with using `document.body`, one thing that will make it less expensive would be to add the removal listener at a later stage, and potentially remove it when the popup is closed. The popup becomes visible here https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/f78d299d5840281dc92a5c8f48502166952f5c80/lib/modules/hover.js#L273

larsjohnsen:
Thanks!

PR #5093, opened 2019-03-29
Hide subredditInfo if no data

prakhar1912:
This also hides the pop up if the subreddit is a fake subreddit like r/all//

*Before:*
Befire

*After*
after

Tested in browser: Chrome 73, Firefox 66

### DIFF
--- a/lib/modules/subredditInfo.js
+++ b/lib/modules/subredditInfo.js
@@ -104,8 +104,6 @@ async function showSubredditInfo(card, subreddit) {
 			cacheFor: HOUR,
 		}): RedditSubreddit);
 	} catch (e) {
-		// Don't show pop up if no subreddit info
-		card.remove();
 		return [null, i18n('subredditInfoErrorLoadingSubredditInfo')];
 	}
 
--- a/lib/modules/subredditInfo.js
+++ b/lib/modules/subredditInfo.js
@@ -10,6 +10,7 @@ import {
 	formatDate,
 	formatDateDiff,
 	formatNumber,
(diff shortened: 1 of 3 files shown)


lib/modules/subredditInfo.js
```diff
@@ -100,6 +104,8 @@ async function showSubredditInfo(card, subreddit) {
 			cacheFor: HOUR,
 		}): RedditSubreddit);
 	} catch (e) {
+		// Don't show pop up if no subreddit info
+		card.remove();
```
larsjohnsen: Note that this makes `subredditInfoErrorLoadingSubredditInfo` (beneath) never appear.

I think this can also can be archived by simply removing `try...catch`, and Hover should automatically close the popup if an error is thrown.

prakhar1912: That is true. But I thought the purpose of the code would be clearer if I caught the AJAX exception.

erikdesjardins: I think we do still want to display the error here if reddit is down / having api issues. Otherwise I think it'll be confusing if the popup just disappears.

But, hmm, do we need this line if we have the fake subreddit changes? Off the top of my head I can't think of any other error cases (sub names with invalid characters like `/r/res@tests` are handled by the regex, and valid names that don't exist return a search page, which is handled below). Maybe just remove this part for now?

erikdesjardins:
I think I'll take the `fakeSubreddit` changes for now cause they look good (and fix the issue in the post). If there's another situation where we need to hide the popup, let me know. Thanks!

PR #5100, opened 2019-04-08
Remove EdgeHTML from codebase.

BenMcGarry:
This will need second review to make sure i didnt mess anything up.

https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/456c0bb00b8f436b1150c5dd1b7bf801cdc5d492/lib/core/init.js#L45 Has a tag for edge but may need checking if anything needs to be removed/changed..


webpack.config.babel.js
```diff
@@ -137,13 +131,6 @@ export default (env = {}, argv = {}) => {
 				use: [
 					{ loader: 'url-loader' },
 				],
-			}, {
-				test: /\.woff$/,
-				use: [
```
erikdesjardins: Keep this, and just change `conf.target === 'edge' ?
						{ loader: 'url-loader' } :
						{ loader: 'file-loader', options: { name: '[name].[ext]' } }`
to
`{ loader: 'file-loader', options: { name: '[name].[ext]' } }`

lib/core/init.js
```diff
@@ -110,14 +110,3 @@ export const afterLoad: Promise<void> = Promise.all([go, watchers, loadComplete]
 
 // BodyClasses may have been added before document.body was ready
 bodyStart.then(BodyClasses.addMissing);
-
-// Preload custom fonts to avoid FOUT as the browser otherwise wouldn't start loading them before encountering them in `go`
-if (process.env.BUILD_TARGET !== 'edge') { // Edge lacks these interfaces, and must loads fonts inlined in CSS; see webpack configuration
```
erikdesjardins: Stuff inside this `if` should be kept

lib/modules/neverEndingReddit.js
```diff
@@ -317,8 +317,7 @@ function createLoaderWidget(startPromise, donePromise) {
 	});
 
 	const removeWidget = () => widget.remove();
-	if (process.env.BUILD_TARGET !== 'edge') donePromise.finally(removeWidget);
```
kevinji: I think this line should be copied, not the bottom line.

BenMcGarry: Ah missed the !. Will fix now.

PR #5119, opened 2019-05-24
keyboardNav: Scroll to top of inline gallery on browse

jeroonk:
Adds an option to scroll back to the top of an inline gallery image when using keyboard navigation.

This has always been a small annoyance for me when when browsing galleries of large-ish images (e.g. comics). Hopefully this will be useful for other users as well.

Maybe a few issues to consider:
- Should this default to *on* as it is now? Because it only scrolls *up*, most users will probably not experience any change in behavior for fully visible galleries.
It may cause some confusion if a user has a post w/ gallery selected, but then continues reading further down the page, before attempting to browse a gallery (without selecting the correct comment first).
- I have chosen to scroll to the top of the `.res-gallery-pieces` container, which contains the image + caption (if the user has changed *`captionsPosition`* to show it above), but not the gallery title or navigation controls. Maybe it should scroll to include the controls?
- The call to `getMostVisibleElementInThingByQuery` was changed from the navigation buttons to the whole gallery, maybe this can cause some issues with multiple galleries of different sizes in the same post? (Not sure).

Please let me know if I missed anything.

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -21,6 +21,7 @@ import {
 	loggedInUser,
 	matchesPageLocation,
 	niceKeyCode,
+	scrollToElement,
 	string,
 	watchForElements,
 	waitForEvent,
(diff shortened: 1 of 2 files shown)


larsjohnsen:
Look good. Could you move the callback to a unified function (e.g. `navigateGallery(direction: 'prev' | 'next')`)?

larsjohnsen:
Thanks!

PR #5124, opened 2019-06-01
Fix nightmode on /report and tweaks to link.

BenMcGarry:
Relevant issue: 
Tested in browser: Edge Version 76.0.172.0

Fixes: https://www.reddit.com/comments/bvelql

### DIFF
--- a/lib/css/modules/_nightMode.scss
+++ b/lib/css/modules/_nightMode.scss
@@ -32,7 +32,6 @@ $title-link-visited-color: hsl(0, 0%, 65%);
 	.link,
 	.helpcenter-form .section-content {
 		background-color: $background-color;
-		border: none;
 	}
 
 	.helpcenter-form .helpcenter-bottom-panel .helpcenter-content-policy p {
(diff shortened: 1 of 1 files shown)


lib/css/modules/_nightMode.scss
```diff
@@ -24,12 +24,21 @@ $title-link-visited-color: hsl(0, 0%, 65%);
...
-	.link {
+	.crosspost-thing-preview {
 		background-color: $background-color;
 		border-color: $border-color;
 	}
 
+	.link,
+	.helpcenter-form .section-content {
+		background-color: $background-color;
+		border: none;
```
larsjohnsen: Is `border: none;` doing anything useful here? If not, I think we might as well forego it as to not break custom styles which has specified a border.

BenMcGarry: A weird yellow border existed on the element, could potentially leave it in. Let me revisit and and check it again.

BenMcGarry: Removed in latest commit.

PR #5127, opened 2019-06-13
Activate RES in redesign embedded iframes. Fixes  #5126

### LINKED ISSUE #5126: Activate RES in redesign embedded iframes.
Reddit redesign makes use of iframes to include some old reddit functionality like the user's inbox. This is done by appended `?embedded=true` to the url. When going there directly RES features appear to work (fairly) well. For example visiting: 

- https://www.reddit.com/r/enhancement?embedded=true

Shows neverending reddit, usertags, etc. 

RES however **does not** activate when the same url is opened in an iframe. It would be nice if it did. 

For complete clarity, I am mostly asking this because I am working on [new toolbox functionality](https://github.com/toolbox-team/reddit-moderator-toolbox-redesign/issues/86) where RES being active would be nice.

creesch:
Relevant issue:  #5126
Tested in browser: Firefox and Chrome

Does exactly what it says in the title. Only activates in redesign embedded pages to prevent issues with other iframes.

I have tested a variety of functions: 
- Tagging
- User hover popup
- Neverending reddit

All of those appear to be working fine. 

image

As is visible in the screenshot there is one small visual issue with `.blueButton` which I haven't fixed yet as I don't know if there are guidelines about specificity. When changing the selector to `.res .blueButton` it did show up fine.

### DIFF
--- a/chrome/beta/manifest.json
+++ b/chrome/beta/manifest.json
@@ -30,6 +30,7 @@
 		"matches": [
 			"https://*.reddit.com/*"
 		],
+		"all_frames": true,
 		"exclude_matches": [
 			"https://mod.reddit.com/*",
 			"https://ads.reddit.com/*",
--- a/chrome/manifest.json
+++ b/chrome/manifest.json
@@ -30,6 +30,7 @@
 		"matches": [
 			"https://*.reddit.com/*"
 		],
+		"all_frames": true,
 		"exclude_matches": [
 			"https://mod.reddit.com/*",
 			"https://ads.reddit.com/*",
--- a/firefox/beta/manifest.json
+++ b/firefox/beta/manifest.json
@@ -34,6 +34,7 @@
 		"matches": [
 			"https://*.reddit.com/*"
 		],
+		"all_frames": true,
 		"exclude_matches": [
 			"https://mod.reddit.com/*",
 			"https://ads.reddit.com/*",
--- a/firefox/manifest.json
+++ b/firefox/manifest.json
@@ -34,6 +34,7 @@
 		"matches": [
 			"https://*.reddit.com/*"
 		],
+		"all_frames": true,
 		"exclude_matches": [
 			"https://mod.reddit.com/*",
 			"https://ads.reddit.com/*",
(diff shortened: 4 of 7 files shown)


creesch:
It was brought to my attention that the way I initially did the check also breaks the option page. I have changed the check, but on IRC it was suggested I could also move the check to foreground. Which I assume means to `foreground.entry.js` before `init()` which is I am also okay with but didn't seem the right place initially as it doesn't contain logic in itself.

lib/core/init.js
```diff
@@ -25,6 +25,9 @@ let start;
 export function init() {
 	if (location.hash === RES_DISABLED_HASH) return;
 
+	// Only open RES in iframes that are reddit embedded pages.
+	if (window.location !== window.parent.location && !(window.location.search.includes('embedded=true') || window.location.pathname === '/options.html')) return;
```
larsjohnsen: I've moved the init logic to foreground.entry. https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/ee0c18d06db840fcd3562bd731ad155f48036e3e

`(new URL(location.href)).searchParams.get('embedded') === 'true'` would be more accurate than `window.location.search.includes('embedded`

creesch: Fair enough, I have changed it to follow that. Also because I feel that it doesn't need to throw an error (it not loading on other iframes is the goal) I wrapped `init()` in the check. 



Do you have a suggestion for the css issue? I could just use 

```css 
.res .blueButton
```

To make it more specific but wanted to check that first.

larsjohnsen: Yeah, throwing errors is a bit too much. I've tweaked the init code in 968914f484b1f6cf2fda0259c319200142557528.

> `window.location !== window.parent.location`

Any reason to check this way instead of simply `window !== window.parent`?

> Do you have a suggestion for the css issue?

Probably simplest to just add `!important` to the `color` rule.

creesch: > Any reason to check this way instead of simply window !== window.parent?

It did seem like a more concrete check but other than that I don't really have a reason. 

> Probably simplest to just add !important to the color rule.

Alright, will do.

creesch: Alright, all fixed.

creesch:
Got everything implemented. Fixed the css specificity issue and also made sure it works nicely with RES nightmode.

creesch:
Is there anything else needed for this to be considered for merging?

larsjohnsen:
There might be some issues with messaging from embedded instances, but I think this will work fine for the most part.

PR #5141, opened 2019-08-11
Set Firefox min version to 69 for ResizeObserver

erikdesjardins:
It seems that ResizeObserver was pushed back to 69 due to regressions: https://bugzilla.mozilla.org/show_bug.cgi?id=1555786

We could also revert #5132 but we can probably just wait 3 weeks for FF69

### DIFF
--- a/firefox/beta/manifest.json
+++ b/firefox/beta/manifest.json
@@ -8,7 +8,7 @@
 	"applications": {
 		"gecko": {
 			"id": "jid1-xUfzOsOFlzSOXg@jetpack",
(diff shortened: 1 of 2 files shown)


BenMcGarry:
Just because i noticed it, is it worth us stripping FF beta info since we dont do them anymore?

erikdesjardins:
Good catch. Although I kinda like having a firefox build with sourcemaps published to github for testing old releases, so I want to keep it for now.

larsjohnsen:
Might as well wait for Firefox 69, agreed.

PR #5146, opened 2019-09-16
added peervideo.net to known peertube domains

bjuergens:
addition to PR #5065 

note: I have not tested this change in a browser, because it is only a single line. If you want me to, I'll gladly test it.


larsjohnsen:
Seems to be fine, thanks.

Issue #1621, opened 2014-09-15
Use textarea instead of input[type=text] in subreddit shortcut editor


jewel-andraia:
for people with ridiculously long multi-subreddit shortcuts, like @farow

abc516:
Hi @andytuba , I was wondering if I could take on this issue please. I am participating in hacktobefest, and saw it was still open. Thanks!

andytuba:
@abc516 Sure, you can go for either the simple `textarea` solution and/or add in a subreddit multi-picker!

edmundho:
hey there, i've opened a PR against this commit for hacktoberfest. please take a look! https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/5154

PR #5154, opened 2019-10-15
use textarea for subreddit shortcut editor

edmundho:
Relevant issue: #1621 
Tested in browser: 

Before:
Screen Shot 2019-10-14 at 10 17 04 PM

After:
Screen Shot 2019-10-15 at 4 02 05 PM

### DIFF
--- a/lib/css/modules/_subredditManager.scss
+++ b/lib/css/modules/_subredditManager.scss
@@ -104,7 +104,7 @@
 	}
 
 	#shortcut-subreddit {
-		width: 240px;
+		width: 234px;
 	}
 
 	#sortButton {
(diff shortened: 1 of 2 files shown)


larsjohnsen:
I get an error:

```
Uncaught TypeError: Cannot read property 'addEventListener' of undefined
    at editSubredditShortcut
```

It seems that `subredditInput` must be updated.

edmundho:
> I get an error:
> 
> ```
> Uncaught TypeError: Cannot read property 'addEventListener' of undefined
>     at editSubredditShortcut
> ```
> 
> It seems that `subredditInput` must be updated.

dang ok. i'll look into that and re-submit.

PR #5170, opened 2019-11-28
Added visual indication of account switching

FrankSalmick:
Currently, the page just sits and appears to be doing nothing when you choose an account to switch to. This change makes the cursor spin while the ajax request is being made and resets the cursor after the ajax request is complete. 

Tested in chrome 78

### DIFF
--- a/lib/modules/accountSwitcher.js
+++ b/lib/modules/accountSwitcher.js
@@ -226,6 +226,10 @@ const switchTo = mutex(async (username: string) => {
 	}
 
 	await logoutPromise;
+	
+	// Provide some form of indication that the account switch is being processed
+	let oldCursorStyle = document.body.style.cursor;
+	document.body.style.cursor = "wait";
 
 	const { success, jquery } = await ajax({
 		method: 'POST',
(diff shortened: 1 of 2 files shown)


lib/modules/accountSwitcher.js
```diff
@@ -245,7 +249,9 @@ const switchTo = mutex(async (username: string) => {
 
 		throw e;
 	});
-
+	
+	document.body.style.cursor = oldCursorStyle;
```
larsjohnsen: This might not be executed if any of the above `await ajax` fails.

larsjohnsen: It might be better to add this change here: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/5170/files#diff-f0f9061059f0e11ef81314d9f20d428cR200

E.g: 
```
element.addEventListener('click', () => {
  const oldCursorStyle = document.body.style.cursor;
  document.body.style.cursor = 'wait';
  switchTo(username).catch(console.error);
  document.body.style.cursor = oldCursorStyle;
})
```

FrankSalmick: That makes sense, I updated the code to match yours. By the way I really appreciate you walking me through all these changes, I'll keep these coding patterns in mind for the next pull request.

larsjohnsen:
Thanks for taking care of the last issues.

PR #5171, opened 2019-11-28
NaverTV host

sporksparks:
Relevant issue: Similar to #5090, this is also used in Korea (kind of like the YouTube of Korea) - used commonly in kpop subreddits as seen here https://www.reddit.com/domain/tv.naver.com/ 
Tested in browser: Chrome 78.0.3904.108


BenMcGarry:
Tested on Edge Version 80.0.361.69 (Official build) (64-bit). Thanks!

PR #5174, opened 2019-12-19
Added RedGIFs as a Host to enable embedding video player

redg-dev:
This pull request adds support for redgifs.com for the embedded video player. This was requested in r/Enhancement last week (https://www.reddit.com/r/Enhancement/comments/e9zn7s/feature_request_please_add_expando_support_for/).

RedGIFs is the new site created by gfycat as the place for all NSFW/adult gifs and short videos.

Tested in browsers: Chrome (Version 79.0.3945.88), Firefox (71.0)


lib/core/migrate/migrate.js
```diff
@@ -382,6 +382,7 @@ const migrations = [
 			await Migrators.moveOption('showImages', 'display Oddshot', 'showImages', 'display_oddshot');
 			await Migrators.moveOption('showImages', 'display Miiverse', 'showImages', 'display_miiverse');
 			await Migrators.moveOption('showImages', 'display swirl', 'showImages', 'display_swirl');
+			await Migrators.moveOption('showImages', 'display redgifs', 'showImages', 'display_redgifs');
```
honestbleeps: I could be mistaken, as it's been a while since I've looked at anything with the migrations, but I don't think this line is necessary.

BenMcGarry: Yeah we can remove the migration as its a new host.

redg-dev: @honestbleeps @BenMcGarry thanks, removed the line. Does the rest look good?

PR #5176, opened 2019-12-28
Liquipedia.net Host Support

iMarbot:
Added Liquipedia.net hosts file and some scss for it. Based on existing Wikipedia hosts file using Mediawiki API. I have also confirmed with Liquipedia that this is acceptable use of the API.


BenMcGarry:
@iMarbot This host appears to now be offline, can you confirm?

BenMcGarry:
@iMarbot The site wasn't responding to any requests when I looked at it but now appears to be back.

BenMcGarry:
From testing we get rate limited from a single page load.

image

BenMcGarry:
After lockout lgtm. Thanks!

PR #5190, opened 2020-03-21
Liquipedia.net: Adding futureproof and some css changes

iMarbot:
Adding future-proof removal of elements using custom class. Added some recently added css for items which are commonly included in the included content. Follow up to #5176

### DIFF
--- a/lib/css/modules/_showImages.scss
+++ b/lib/css/modules/_showImages.scss
@@ -665,53 +665,91 @@ body:not(.res-showImages-displayImageCaptions) {
 }
 
 .res-expando-box[data-host='liquipedia'] {
-	.md .fo-nttax-infobox {
-		width: 100%;
-		max-width: 300px;
-		border: none;
-		background: #eee;
-		margin-left: 8px;
-		float: right;
+	.md {
+		.fo-nttax-infobox {
+			width: 100%;
+			max-width: 300px;
+			border: none;
+			background: #eee;
+			margin-left: 8px;
+			float: right;
 
-		.infobox-cell-2 {
-			width: 46%;
-		}
+			.infobox-cell-2 {
+				width: 46%;
+			}
 
-		.res-nightmode & {
-			color: rgb(221, 221, 221);
+			.res-nightmode & {
+				color: rgb(221, 221, 221);
+
+				> div:nth-child(2n) {
+					background-color: rgb(60, 60, 60);
+				}
+
+				> div:nth-child(2n+1) {
+					background-color: rgb(57, 57, 57);
+				}
+			}
 
 			> div:nth-child(2n) {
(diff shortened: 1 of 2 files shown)


BenMcGarry:
If the browser pre-loads URLs yes.

larsjohnsen:
The `handleLink` function is executed when the viewport is close to the expando. This causes the liquipeida html to load regardless of whether the users actually opens the expando.

larsjohnsen:
The processing is managed in `utils/watchers.js`.  There's no easy way to override this behavior, though the current system certainly could be optimized to avoid unnecessarily preloading.

BenMcGarry:
@larsjohnsen Is it possible to disable hosts out the box for opt-in? Not sure if we have that functionality.

@iMarbot If we can disable as standard would you want it to stay or still remove? Ill add a PR now to remove pending lars.

honestbleeps:
Hi all, I've not chimed in in quite some time due to being busy but if I'm
understanding this behavior accurately I'd like to change it for sure.

We should have the extension be respectful of external hosts and not
preload excessively. I can put in a pr if ben can't get to it.

I feel like we shouldn't preload any non reddit stuff at all, let alone
before it's in the viewport.

BenMcGarry:
For the next release we could set a migration to disable the module to allow it opt-in?

BenMcGarry:
Ok, I shall remove it now and add the option of preventing excessive API requests to the todo list.

notr1ch:
I'm the above mentioned server admin for Liquipedia. I've had to temporarily block the RES requests by responding with a 403 to the CORS preflight requests as there's currently a thread on the top of /r/leagueoflegends that is causing very high load. 

The core problem is already mentioned above - RES shouldn't be requesting the API data until the user clicks the expando. Some tournament pages on Reddit have Liquipedia links to every team / series which is playing, this can cause RES to send 50+ API requests per page load and the user ends up temporarily blocked from Liquipedia due to our rate limiting setup. The likelihood of a user interacting with more than one Liquipedia expando is quite low, even more so if they're refreshing Reddit only to see new comments (Liquipedia links are often used for live events and sometimes hidden behind spoiler blocks).

Feel free to ping me if you'd like any feedback on any new Liquipedia expando implementation in the future!

BenMcGarry:
We are looking at pushing out a hotfix to remove your module to lower load on your servers. Apologies for this.

notr1ch:
The requests are easily identified and dropped on our frontend servers so we're not seeing any further issues. Take your time - don't feel pressured to rush out a hotfix on our behalf.

BenMcGarry:
We've just pushed v5.18.13 with the module removed. Should be live on all platforms within 24hours.

PR #5191, opened 2020-03-22
Migrate somes files from "lib/modules/hosts" to flow .0.85.0

fedemp:
I noticed that in https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/5002 it's mentioned that you are stuck in flow 0.84.0 due to the new warnings that appear in 0.85.0.

I've checking it and I was able to fix the errors for some files in `"lib/modules/hosts"`. But before going on with the rest, I wanna check if this would be an accepted contribution.

This PR does not upgrade flow yet cause there are a lot of other issues, but it's a beggining.

### DIFF
--- a/lib/core/host.js
+++ b/lib/core/host.js
@@ -3,7 +3,7 @@
 import type { Thing } from '../utils';
 import type { ModuleOption } from './module'; // eslint-disable-line no-unused-vars
 
-export class Host<Detect, Opt: { [string]: ModuleOption<any> }> {
+export class Host<Detect, Opt: { [string]: ModuleOption<any> } | void> {
 	moduleID: string;
 
 	name: string;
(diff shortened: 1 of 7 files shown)


lib/core/host.js
```diff
@@ -3,7 +3,7 @@
 import type { Thing } from '../utils';
 import type { ModuleOption } from './module'; // eslint-disable-line no-unused-vars
 
-export class Host<Detect, Opt: { [string]: ModuleOption<any> }> {
+export class Host<Detect, Opt: { [string]: ModuleOption<any> } | void> {
```
fedemp: Not all hosts have `options` for the constructor.

lib/core/host.js
```diff
@@ -17,6 +17,7 @@ export class Host<Detect, Opt: { [string]: ModuleOption<any> }> {
 	options: Opt | void;
 
 	detect: (a: URL, context: ?Thing) => void | null | false | Detect;
+	// handleLink: (href: string, detectResult: Detect | null) => ExpandoMedia | Promise<ExpandoMedia>;
```
fedemp: In some cases, `detect` tests a string against a RegExp that may return `null` as mentioned in line #19. Given so, `handleLink` must be made aware of that; that `detectResult` may be `null`. I haven't changed it yet cause it breaks all other hosts but added the comment of what would be right.

larsjohnsen: I'm not sure that this would be necessary. The only time `handleLink` is invoked, is after this conditional: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/46208aab875e262f24b0923e559625453682598d/lib/modules/showImages.js#L856-L857

PR #5195, opened 2020-04-05
Use getRedirectURL in Firefox to fix backups in 75 and later

### LINKED ISSUE #5184: OAuth redirection URI must be updated for Firefox
Breaking change in Firefox 75. All OAuth providers must allow `https://f6be6623b0b4ed4e84464ff8b07afff3c421e497.extensions.allizom.org` as redirect URIs.

https://bugzilla.mozilla.org/show_bug.cgi?id=1617885

Updated docs https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/identity#Getting_the_redirect_URL:
> Starting with Firefox 75, you must use the redirect URL returned by identity.getRedirectURL().  Earlier versions allowed you to supply any redirect url.

erikdesjardins:
Fixes #5184

After this change, OneDrive backups in Firefox are separate from OneDrive backups in other browsers. Microsoft requires that all redirect URIs for an application are on the same domain. In Firefox 75 and later we are forced to use a Mozilla domain like f6be6623b0b4ed4e84464ff8b07afff3c421e497.extensions.allizom.org, which of course is different from redditenhancementsuite.com. So there is now a separate "Reddit Enhancement Suite (for Firefox)" OneDrive application used in Firefox.

Tested in browser: Firefox 75

### DIFF
--- a/lib/environment/background/auth.js
+++ b/lib/environment/background/auth.js
@@ -3,7 +3,7 @@ import { apiToPromise } from '../utils/api';
 import { addListener } from './messaging';
 
 addListener('authFlow', ({ domain, clientId, scope, interactive }) => {
-	const redirectUri = 'https://redditenhancementsuite.com/oauth';
+	const redirectUri = process.env.BUILD_TARGET === 'chrome' ? 'https://redditenhancementsuite.com/oauth' : chrome.identity.getRedirectURL();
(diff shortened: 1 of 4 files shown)


BenMcGarry:
Do we need this for Dropbox/G Drive as well?

PR #5204, opened 2020-04-19
Remove Host at request of original author due to API requests.

BenMcGarry:
image


larsjohnsen:
Associated CSS should also be removed.

Issue #2305, opened 2015-06-22
Add "block user" buttons


andytuba:
https://www.reddit.com/r/Enhancement/comments/3ao08f/notifications_from_blocked_users/

reddit only shows the "block" button in the messages inbox. It would be nice to also show this button on the user info / tagger dialogs and in /prefs/blocked.

---

This is probably only worth doing if blocking a user prevents them replying to you in public, too.

erikdesjardins:
https://www.reddit.com/dev/api/#POST_api_block_user

prech:
Would highly appreciate this feature, please.

Currently resorting to manually inserting CSS snippets to hide those users and avoid harassment, so something more convenient/native would be greatly appreciated!

johnnybigoode:
I'd like to tackle this if it's still open eh?

johnnybigoode:
@andytuba 
I'd ask you to confirm the block options for me please:
>user info / tagger dialogs and in /prefs/blocked.

**The prefs/blocked seems easy enough:**
![image](https://user-images.githubusercontent.com/2698584/79743414-ea15cc80-82da-11ea-8875-e657843b1131.png)

Reminding that this already exists when using the redesign, but ew, who uses the new redesign:
![image](https://user-images.githubusercontent.com/2698584/79743521-17fb1100-82db-11ea-88ef-b5613bd8e1bd.png)

**User info dialog would be this one:**
![image](https://user-images.githubusercontent.com/2698584/79743826-a96a8300-82db-11ea-890b-d3cf24d02cc1.png)

**And the tag dialog is this one**
![image](https://user-images.githubusercontent.com/2698584/79744003-ffd7c180-82db-11ea-8504-9ab99f480957.png)

Is all of this correct?

andytuba:
@johnnybigoode yeah, go for it.  did you delete your mockup screenshots?

johnnybigoode:
> 
> 
> @johnnybigoode yeah, go for it. did you delete your mockup screenshots?

I did because I figured that making the prefs/blocked would mean a new module, or at least, it wasn't clear how to implement. 

The tag part seems like a hassle, since anywhere in Reddit where you can tag a user you can hoover over their name and get the userInfoCard

Oh, by the way, I think I'm done...?

Sadly, wasn't able to do the integration tests locally thou.

![image](https://user-images.githubusercontent.com/2698584/79804424-65f33180-833a-11ea-991a-c8737481600a.png)

[...]

PR #5211, opened 2020-04-21
Implements #2305 - add block user button 

johnnybigoode-zz:
Implements #2305 
No local integration tests

filteReddit.js was changed so toggleString would accept an extra parameter 'saveOptions'
If called the func it would remove all posts related to the string **and also add it to the filter list**
This way, if you need to make some posts hide because you just blocked a new user, you can, without adding the blocked user to the filter list. 

userInfo.js has a new if(userCard) section which adds blocking, it follows the patterns, I would kindly ask for someone to review the ajax call, I think it could be a little more cleaner 

en.json has a new field for the text of the block button

### DIFF
--- a/lib/modules/userInfo.js
+++ b/lib/modules/userInfo.js
@@ -226,7 +226,8 @@ async function showUserInfo(card) {
 							api_type: 'json',
 							name: data.name,
 							uh: String(hash) },
-					}).then(() => FilteReddit.listFilters.users.toggleString(data.name, true, false));
+							//undo changes to FilteReddit.listFilters.users.toggleString(userTag.id, true, false));
+    					}).then(() => addCSS(`.thing[data-author="${data.name}"] { display: none !important; }`));
 				},
 			);
 		});


lib/modules/userInfo.js
```diff
@@ -209,6 +212,26 @@ async function showUserInfo(card) {
 		userHTML += string.escape`<div class="${isHighlight ? 'blueButton' : 'redButton'}" id="highlightUser" data-userid="${data.id}">${isHighlight ? i18n('userInfoHighlight') : i18n('userInfoUnhighlight')}</div>`;
 	}
 
+	if (userTag) {
```
larsjohnsen, line 215: `userTag` is not required, if `data.name` is used instead of `userTag.id`.

lib/modules/userInfo.js
```diff
@@ -209,6 +212,26 @@ async function showUserInfo(card) {
...
+				hash => {
+					ajax({
+						method: 'POST',
+						url: '/api/block_user',
+						data: {
+							account_id: `t2_${userTag.id}`,
+							api_type: 'json',
+							name: userTag.id,
+							uh: String(hash) },
+					}).then(() => FilteReddit.listFilters.users.toggleString(userTag.id, true, false));
```
larsjohnsen, line 229: This may be inadvertently be saved at a later stage, e.g. if another user is ignored. It might be simpler to emulate the blocking behavior by adding a CSS snippet:

```addCSS(`.thing[data-author="${data.name}"] { display: none !important; }`)```

File: lib/modules/userInfo.js [hunk 3]
```diff
@@ -209,6 +212,27 @@ async function showUserInfo(card) {
 		userHTML += string.escape`<div class="${isHighlight ? 'blueButton' : 'redButton'}" id="highlightUser" data-userid="${data.id}">${isHighlight ? i18n('userInfoHighlight') : i18n('userInfoUnhighlight')}</div>`;
 	}
 
+	if (data.name) {
```
larsjohnsen, line 215: Note that `data.name` always is truish at this stage, so this conditional is not needed.

lib/modules/userInfo.js
```diff
@@ -209,6 +212,27 @@ async function showUserInfo(card) {
...
+		userHTML += string.escape`<a target="_blank" rel="noopener noreferer" id="blockUser" class="blueButton">${i18n('userInfoBlock')}</a>`;
+
+		$body.on('click', '#blockUser', () => {
+			loggedInUserHash().then(
```
larsjohnsen, line 219: If the `loggedInUserHash` doesn't a return hash (like when you're not logged in), the button will useless. It might be better to only add the button if `hash` is truish.
```
const hash = await loggedInUserHash();
if (hash) { /* ... add button */ }
```

PR #5213, opened 2020-04-24
backupAndRestore: Fix Google Drive token retrieval

### LINKED ISSUE #5212: [Chrome] Google Drive backup module not recieving access token.
As of last night it appears that during the oauth call-back no access token is being returned from google blocking backups/restores of RES settings. G Drive API Dev console shows no major errors.

No console errors from the google network requests.

image

larsjohnsen:
The Google OAuth token retrieval did an additional redirection before landing at `redditenhancementsuite.com`, which was unnecessary and not support by our system.

Relevant issue: Fixes #5212
Tested in browser: Chrome 84, Chrome 75

### DIFF
--- a/chrome/beta/manifest.json
+++ b/chrome/beta/manifest.json
@@ -82,7 +82,7 @@
 		"https://www.flickr.com/services/oembed",
 
 		"https://redditenhancementsuite.com/oauth",
-		"https://accounts.google.com/o/oauth2/v2/auth",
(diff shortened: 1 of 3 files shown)


BenMcGarry:
Tested on Firefox 74 which wasn't impacted originally and still works good.

BenMcGarry:
Tested on Chrome 81 and FF 76 - Working

PR #5232, opened 2020-06-12
Update Twitch with latest API changes.

### LINKED ISSUE #5123: No Next/Previous Page navigation after update to 5.16.9
- My browser: Firefox ESR 60.7.0
- My RES version: 5.16.9
- Am I in reddit beta (yes/no):  no

*What's the problem?*
After updating to 5.16.9 I do not have any page navigation elements anymore (previous/next at bottom of page). Possibly because of commit https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/9a890d37f5ae8843fe589f0dca359e634f3be590
I am using old.reddit.com and have never ending reddit disabled. I tried clearing the res cache and also resetting to factory default which didnt help. Also reset all cookies. Never Ending Reddit does not seem to work either. Downgrading to 5.16.7 fixes this.

None that I could find. 

uBlock Origin, uMatrix, https everywhere, vim vixen, keepassxc, redirect amp to html

BenMcGarry:
Fixes #5123 

Twitch changed their API requirng the domain location in the URL. This adds www and old. We dont touch redesign so can ignore new as Reddit handle that natively. They said they won't support wildcard (*.) so if anyone uses subdomains like np. they are out of luck.

https://discuss.dev.twitch.tv/t/twitch-embedded-player-updates-in-2020/23956

### DIFF
--- a/lib/modules/hosts/twitch.js
+++ b/lib/modules/hosts/twitch.js
@@ -12,7 +12,7 @@ export default new Host('twitch', {
 			`video=${typeId.replace('b', 'a')}${videoId}` :
 			`channel=${channel}`;
 
-		const embed = `https://player.twitch.tv/?${channelOrVideo}&time=${+h || 0}h${+m || 0}m${+s || 0}s`;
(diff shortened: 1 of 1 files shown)


larsjohnsen:
Would using `parent=${location.hostname}` work?

BenMcGarry:
Hmm didnt think of that, let me try.

PR #5235, opened 2020-06-13
Changed redgifs module to video instead of iframe

borderline232:
- Used similar functionality found in the gfycat.js host file and
  implemented similar video options to have redgifs display as a video
  rather than iframe allowing for video size and volume adjustments
- fixed api url to reflect new api url

Relevant issue: None, better overall functionality
Tested in browser: Firefox


BenMcGarry:
The API appears to have stability issues. We appear to hit a block pretty quick. I'm reaching out to the devs to see what their views on this are as they gave us a iframe implementation.

PR #5239, opened 2020-07-01
Add visitOnLoad to neverEndingReddit

a8f:
Add options in neverEndingReddit to "visit" all links on the previous page when a new page is loaded


larsjohnsen:
I think this would be better suited for the betteReddit module. 

Also take a look at the `watchForThings` interface, and the `Thing` methods. By using those, you could do something like this (not tested):

```
watchForThings('post', thing => {
  if (
    (module.options.visitOnLoad.value === 'links' && !thing.isLinkPost()) ||
    (module.options.visitOnLoad.value === 'self' && !thing.isSelfPost())
  ) return;
  addURLToHistory(thing.getPostUrl())
});
```

PR #5240, opened 2020-07-06
Try redgifs if gfycat fails

fjkda:
Relevant issue: Older gfycat links fail to expand if they've moved to redgifs
e.g. https://www.reddit.com/r/NSFW_GIF/top/?sort=top&t=all (NSFW)
Tested in browser: Firefox


fjkda:
Ack, this issue was due to some content blockers in Firefox and the fix was really to add white lists.

PR #5245, opened 2020-07-22
Add Reddit Poll support

BenMcGarry:
Adds support for Reddit Poll on old reddit.

Tested on Edge 84.0.522.40.

On main feed:
image

In Post:
image


BenMcGarry:
Looks like it causes a internal server error when loading from main page listing. In post is fine.

BenMcGarry:
@larsjohnsen As its domain Reddit it takes over all expandos and causes internal server errors on feeds. Thoughts?

BenMcGarry:
Never mind, stole your check from the gallery module. As follows same format and works.

PR #5261, opened 2020-11-03
Added support for screen readers to gallery view and tables

FrankSalmick:
Relevant issue: None, I just did this without creating one
Tested in browser: Chrome 85 on Windows 10


larsjohnsen:
Thanks!

Issue #5274, opened 2021-01-15
Edge 87: Attempting to backup/restore RES settings with Google Drive/OneDrive/Dropbox fails


kirazy:
- My browser: https://www.whatismybrowser.com/w/GF6ZK9N
- My RES version: v5.20.12
- Am I in reddit beta (yes/no): No

*What's the problem?*
Attempting to create or restore backups using any of Google Drive, Dropbox or One Drive results in an error: "Cannot read property 'getRedirectURL' of undefined".

*Are there any related posts in https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues or /r/RESissues?*
https://www.reddit.com/r/RESissues/comments/krzcut/unable_to_restore_saved_settings/

*What other browser extensions are installed?*
uBlock Origin, NoScript, LastPass, Picture-in-Picture

alysson-silva:
This only happens for me if I use the extension from the Microsoft Store. The one from Chrome works fine.

BenMcGarry:
Fixes in #5279

PR #5279, opened 2021-02-01
Fix build target on certain modules

BenMcGarry:
Fixes issue where redirectUri build target wasn't hitting Edge since the build changes. As well as replicating the fix to other relevant areas.

Relevant issue: #5274
Tested in browser: Edge 88

@larsjohnsen - are you able to review please?


larsjohnsen:
The same is probably required for the `opera` target. Might be simpler to modify the conditionals to check that `BUILD_TARGET` is not `firefox`, or define a new env variable in Babel which defines the browser engine (e.g. `'process.env.BUILD_TARGET_ENGINE': conf.target_engine`).

PR #5281, opened 2021-02-12
Update twitchclips regex for new clip names.

kcyk01:
Twitch has started appending hyphen + random string to the end of their auto-generated clip names, causing the twitchclips regex to not match anymore.

Fixes the issue mentioned in /r/Enhancement:  
[[Request] Twitch recently changed the format of their clips URLs which has affected functionality.](https://reddit.com/r/Enhancement/comments/lhk8hq/request_twitch_recently_changed_the_format_of/)

Tested in browsers:
* Firefox 85.0.1
* Chrome 88.0.4324.150

### DIFF
--- a/lib/environment/background/auth.js
+++ b/lib/environment/background/auth.js
@@ -3,7 +3,7 @@ import { apiToPromise } from '../utils/api';
 import { addListener } from './messaging';
 
 addListener('authFlow', ({ domain, clientId, scope, interactive }) => {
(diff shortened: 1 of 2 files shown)


larsjohnsen:
Thanks!

Saxayone:
There is still an edge case that was missed with this fix.
Where it has a double hyphen, example:
https://clips.twitch.tv/ResourcefulCrazyNarwhalMau5--lX7NFdp8DZQTyMV

PR #5290, opened 2021-03-21
Move Geolocation to Permissions

BenMcGarry:
Geolocation does not work in optional causing it to be ignored.

image

Tested in Chrome/Edge.


larsjohnsen:
Having `geolocation` as an optional permission is allowed in Firefox at least. I'm still uncertain whether this will cause an additional prompt.

Does the issue only occur in the settings console? In that case, we might rather just have RES not prompt for geolocation when loading it.

BenMcGarry:
Do you mean an additional prompt at extension load?

The API not being available breaks in settings as well as page load if the user has their nightmode setting to be set to location.

BenMcGarry:
Playing around with this, Chrome does require main permissions and the below trigger fires when the API is called. Not before.

image

Firefox triggers this when the API is called:

image

However I have found an bug with it. If permission is granted, then nightmode is turned OFF and the permission removed. When you enable it again the permission prompt will not fire resulting in the feature being bricked.

larsjohnsen:
> Playing around with this, Chrome does require main permissions and the below trigger fires when the API is called. Not before.

Are we sure about this? Typically will Chrome disable the extension and require that the user explicitly grant the new permission when it is upgraded.

> The API not being available breaks in settings as well as page load if the user has their nightmode setting to be set to location.

I am able to only replicate this issue on the settings page. I believe that as long as the user has granted permission for the web page use to use geolocation, RES will be able to retrieve it via the content script.

BenMcGarry:
I ran the extension unpacked without the geolocation API, then rebuilt with the perm and updated in Chrome. Extension page showed the geolocation permission but I was not re-prompted.

larsjohnsen:
I don't think the user is prompted for permissions at all when adding the extension unpacked though. But we can test this in the beta branch.

BenMcGarry:
With chrome not allowing this as an optional permission do we have a choice with it? It looks like from the original PR the API was never put in so the features never worked?

https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/3806

larsjohnsen:
I've never had it an issue with it not working. Some users might however have decided to not grant permissions for geolocation to reddit.com, which indeed breaks it.

PR #5291, opened 2021-03-25
nightmode: Silence geolocation permission issue in settings console

larsjohnsen:
The settings console doesn't run in the reddit.com-context, and requires that an additional permission to geolocation is provided in the manifest-file and granted by the user. Since automatic nightmode isn't really that useful for the settings console, the easiest solution is to just ignore the issue.

Relevant issue: #5290
Tested in browser: Firefox 87, Chrome 88

### DIFF
--- a/lib/modules/nightMode.js
+++ b/lib/modules/nightMode.js
@@ -254,6 +254,10 @@ async function getNightModeTimes() {
 
 				switch (err.code) {
 					case err.PERMISSION_DENIED:
+						// Ignore permission issues when the user is in the settings console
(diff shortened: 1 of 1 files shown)


larsjohnsen:
@BenMcGarry, could you check if this solution is sufficient?

BenMcGarry:
@larsjohnsen Yep this appears to work, do we want to remove the geolocation permission from all manifests?

larsjohnsen:
@BenMcGarry Yeah, I think we might as well do that.

PR #5303, opened 2021-07-13
showParent: Fix Shift-P not working

gurnec:
The <kbd>Shift</kbd>-<kbd>P</kbd> shortcut which shows parent posts is currently broken—it does nothing when pressed.

Luckily the fix was simple. `Hover`'s `watch` method doesn't return anything, and the `hover` variable was always assigned `undefined` instead of the `HoverInfoCard` object that was intended.

(Chromium: 91.0.4472.124)


larsjohnsen:
Thanks!

PR #5340, opened 2021-07-27
Remove clear: both from live previews

ElijahPepe:
clear: left is already used for usertext and wiki live previews. This fixes an issue when providing ban notes; the live preview is shifted extremely far down due to how floating elements work.

Tested in browser: Firefox 90 on Windows 10 Pro.


ElijahPepe:
Live preview content is shifted extremely far down while using RES on the ban user page.

PR #5355, opened 2021-11-06
Disable mouse selecting on subreddit info buttons

tankorsmash:
Relevant issue: No issue, but quite a few times I've clicked on 'Filter' and accidentally selected the text on the button, rather just clicking it. This resolves that entirely!

Tested in browser: Edited the CSS manually in Chrome to verify that `.subredditInfoToolTip .bottomButtons` is a selector and `user-select: none;` is the rule.


larsjohnsen:
Thanks!

PR #5369, opened 2022-01-15
RES Test Pipeline

BenMcGarry:
Pipeline for build tests, no deployment to prevent issues incase the workflow is broken.


ZephrFish:
it's broken @BenMcGarry

PR #5371, opened 2022-01-15
First pipeline test

BenMcGarry:
GH actions pipeline.

TODO:

- [ ] Fix Coveralls
- [ ] Fix SauceLabs
- [ ] Add Deploy back in

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -36,7 +36,7 @@ jobs:
       env:
         SAUCE_USERNAME: RES-Auto
         SELENIUM_HOST: ondemand.saucelabs.com
-        SELENIUM_PORT: 443
+        SELENIUM_PORT: 80
         SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
 
   windows:
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -1,0 +1,57 @@
+name: RES Pipeline
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


BenMcGarry:
@erikdesjardins You wouldnt happen to know saucelabs wont connect? Just returns:

> ⚠ Error connecting to ondemand.saucelabs.com on port 80.
> View results: https://saucelabs.com/tests/undefined/

BenMcGarry:
GH actions doesnt send env vars to forked pr's. Merging to mainline after talking to bleeps on call to test the action. This will only merge testing capabilities. No deployment.

erikdesjardins:
@BenMcGarry yeah, we'll probably just have to stop running tests on PRs (this was a problem for us at $JOB too lol, we had to have everyone push to the main repo and stop using forks, until they [added an option](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/) to allow that--which is not safe on public repos so we can't use it here).

So there are a few options:
- have trusted contributors push wip branches to hb/RES instead of their forks, so at least we get test runs in that case
- use something like bors/mergify, which will push approved PRs to a staging branch in the main repo to run tests, then automatically merge them if it passes

BenMcGarry:
@erikdesjardins After merging the tests to master saucelabs still doesnt like to connect, we added the env variable to GitHub but still throws an error.

Since Browserstacks now own nightwatch going to try move there as hopefully its got nicer intergration. I think for the rate the project is now, testing on main merge will probably be the easiest option to prevent over-engineering a solution.

PR #5390, opened 2022-02-17
Reliably detect reddit gallery posts with a caption link

themacguffinman:
In old.reddit.com, a reddit gallery's title link can be misleadingly set to the source link in the gallery's caption instead of the normal reddit.com/gallery/\<id\> URL. The only reliable indication that a post is a reddit gallery is by checking the "data-is-gallery" attribute in the Thing element.

An example of reddit gallery post with a mangled title link: https://old.reddit.com/r/manga/comments/stv143/disc_sir_liebe_von_trapp_ch_4_5_by_isiyumi/

Tested in browser: Chrome

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -812,9 +812,8 @@ function resolveMediaUrl(element, thing) {
 	) {
 		// In old.reddit.com, a reddit gallery's title link can be misleadingly set
 		// to the source link in the gallery's caption instead of the normal reddit.com/gallery/<id> URL.
-		// The only reliable gallery detection is the "data-is-gallery" attribute in the Thing element.
-		const isGallery = thing.element.getAttribute('data-is-gallery') === 'true';
-		if (isGallery) {
+		// The only reliable way to detect a reddit gallery is to check the "data-is-gallery" attribute in the Thing element.
+		if (thing.element.dataset.isGallery === 'true') {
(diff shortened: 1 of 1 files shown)


lib/modules/showImages.js
```diff
@@ -810,6 +810,15 @@ function resolveMediaUrl(element, thing) {
 		thing &&
 		element.classList.contains('title')
 	) {
+		// In old.reddit.com, a reddit gallery's title link can be misleadingly set
+		// to the source link in the gallery's caption instead of the normal reddit.com/gallery/<id> URL.
+		// The only reliable gallery detection is the "data-is-gallery" attribute in the Thing element.
+		const isGallery = thing.element.getAttribute('data-is-gallery') === 'true';
+		if (isGallery) {
```
larsjohnsen, line 817: Could probably be simplified to `if (thing.element.dataset.isGallery)`, otherwise looking good!

themacguffinman: I inlined the boolean expression, but I can't do a truthy string check on `dataset.isGallery` alone because `dataset.isGallery` can be set to a `"false"` string.

themacguffinman:
@larsjohnsen is there anything else needed to get this merged?

BenMcGarry:
Thanks for the contribution! We will release this in the next version :)

PR #5394, opened 2022-04-29
Add syntax highlighting

ElijahPepe:
Tested in browser: Firefox 90.0.1/99.0.1

This PR adds in a new option to automatically highlight all code blocks on Reddit using [highlight.js](https://highlightjs.org/).

### DIFF
--- a/lib/modules/syntaxHighlighting.js
+++ b/lib/modules/syntaxHighlighting.js
@@ -1,0 +1,25 @@
+/* @flow */
+
+import hljs from 'highlight.js';
+import { Module } from '../core/module';
+
+export const module: Module<*> = new Module('syntaxHighlighting');
+
+module.moduleName = 'syntaxHighlightingName';
+module.category = 'appearanceCategory';
+module.description = 'syntaxHighlightingDesc';
+module.options = {
+	auto: {
+		title: 'syntaxHighlightingEnableTitle',
+		type: 'boolean',
+		value: true,
+		description: 'syntaxHighlightingEnableDesc',
+		bodyClass: true,
+	},
+};
+
+module.contentStart = () => {
+	if (module.options.auto.value) {
+		hljs.highlightAll()
+	}
+};
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "dompurify": "2.3.5",
     "fast-levenshtein": "3.0.0",
     "favico.js": "0.3.10",
+    "highlight.js": "^11.5.1",
     "jquery": "3.6.0",
     "lodash-es": "4.17.21",
     "snudown-js": "3.2.2",
--- a/lib/css/modules/_syntaxHighlighting.scss
+++ b/lib/css/modules/_syntaxHighlighting.scss
@@ -1,0 +1,176 @@
+pre code.hljs {
+	display: block;
+	overflow-x: auto;
+	padding: 1em;
+}
+code.hljs {
+	padding: 3px 5px;
+}
+.hljs {
+	background: #f3f3f3;
+	color: #444;
+}
+.hljs-comment {
+	color: #697070;
+}
+.hljs-punctuation,
+.hljs-tag {
+	color: #444a;
+}
+.hljs-tag .hljs-attr,
+.hljs-tag .hljs-name {
+	color: #444;
+}
+.hljs-attribute,
+.hljs-doctag,
+.hljs-keyword,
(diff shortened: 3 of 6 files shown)


lib/modules/syntaxHighlighting.js
```diff
@@ -0,0 +1,25 @@
...
+		type: 'boolean',
+		value: true,
+		description: 'syntaxHighlightingEnableDesc',
+		bodyClass: true,
+	},
+};
+
+module.contentStart = () => {
+	if (module.options.auto.value) {
+		hljs.highlightAll()
```
creesch, line 23: It probably isn't that big of a concern, but if this is done on a page with lots of comments and lots of code blocks it might lock up the page for a bit. 

It could be that hljs already takes care of that, but if that isn't the case it might be worthwhile to take a slightly different approach. Something along the lines of first selecting all visible things on page and highlighting them individually through [`forEachChunked`](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/7b7341cf2c897e7c7538180cd8a6e8b70b4c5671/lib/utils/async.js#L26) (there might be a better utility function for this but my RES is a bit rusty).


Another but also minor issue is comments that get dynamically loaded in later (load more comments), I don't think this will highlight those.

lib/modules/syntaxHighlighting.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const module: Module<*> = new Module('syntaxHighlighting');
+
+module.moduleName = 'syntaxHighlightingName';
+module.category = 'appearanceCategory';
+module.description = 'syntaxHighlightingDesc';
+module.disabledByDefault = false;
+
+module.beforeLoad = () => {
+	watchForThings(['post', 'comment', 'message'], highlight);
```
larsjohnsen, line 15: Instead of `watchForThings(['post'])`, use `watchForElements(['selfText'], null, ele => hljs.highlightElement(ele))` as this will look for code in posts when they become are expanded.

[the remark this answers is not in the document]
ElijahPepe: Doesn't seem like Highlight.js wants to take in a HTMLCollection element in this way from what I've seen. Not sure if it would work with the whole element since I tested it with jQuery and selecting the closest pre tag.

lib/modules/syntaxHighlighting.js
```diff
@@ -0,0 +1,17 @@
...
+export const module: Module<*> = new Module('syntaxHighlighting');
+
+module.moduleName = 'syntaxHighlightingName';
+module.category = 'appearanceCategory';
+module.description = 'syntaxHighlightingDesc';
+module.disabledByDefault = false;
+
+module.beforeLoad = () => {
+	watchForElements(['selfText'], null, ele => hljs.highlightElement(ele.querySelector('pre code')));
+	watchForThings(['comment', 'message'], thing => hljs.highlightElement(thing.entry.querySelector('pre code')));
```
larsjohnsen, line 16: After taking a closer look, it is probably possible to cover all cases by simply using:
`watchForElements(['selfText', 'page'],  'pre code', hljs.highlightElement)`

package.json
```diff
@@ -58,6 +58,7 @@
     "dompurify": "2.3.5",
     "fast-levenshtein": "3.0.0",
     "favico.js": "0.3.10",
+    "highlight.js": "^11.5.1",
```
larsjohnsen, line 61: Nit: Lock this by removing the `^`

lib/modules/syntaxHighlighting.js
```diff
@@ -0,0 +1,17 @@
+/* @flow */
+
+import hljs from 'highlight.js';
```
larsjohnsen, line 3: Since this is a fairly large library which is requires a little processing when loaded, we might want to lazy load it. Take a look at how the dash.js library is loaded:
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/5a36f5ab6dfbb48b661bc61b6b7a540ea17d9e78/lib/modules/showImages.js#L4
https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/5a36f5ab6dfbb48b661bc61b6b7a540ea17d9e78/lib/modules/showImages.js#L1972

ElijahPepe: I was originally going to lazy load Highlight.js, but unfortunately no dist files are provided in the NPM package.

Issue #5273, opened 2021-01-05
Feature request: Allow provider-neutral webdav sync


ShadowJonathan:
Allow the settings/data sync with a generic WebDAV endpoint alongside the existing three proprietary options (Dropbox, Onedrive, Gdrive)

This'd allow the settings to be backed up to a personal cloud like Nextcloud

ElijahPepe:
Resolved in #5398.

PR #5398, opened 2022-05-01
Add WebDAV support to backups

ElijahPepe:
Relevant issue: #5273
Tested in browser:

### DIFF
--- a/lib/modules/backupAndRestore.js
+++ b/lib/modules/backupAndRestore.js
@@ -103,6 +103,30 @@ module.options = {
 		title: 'backupAndRestoreGoogleAccountTitle',
 		advanced: true,
 	},
+	webDAV: {
+		type: 'table',
+		addRowText: 'backupAndRestoreAddWebDAVAccount',
+		fields: [{
+			key: 'username',
+			name: 'backupAndRestoreUsername',
+			type: 'text',
+		}, {
+			key: 'password',
+			name: 'backupAndRestorePassword',
+			type: 'password',
+		}, {
+			key: 'url',
+			name: 'backupAndRestoreUrl',
+			type: 'text',
+		}, {
+			key: 'digest',
+			name: 'backupAndRestoreDigest',
+			type: 'boolean',
+		}],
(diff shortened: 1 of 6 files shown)


BenMcGarry:
Thanks for the PR!

We are having some discussions on this among the team due to the concern of storing plaintext credentials in file store in RES. Please stand by.

ElijahPepe:
Passwords for accounts are stored in plaintext.

BenMcGarry:
Reddit account passwords are a lot less risk compared to someones file storage.

BenMcGarry:
Yeah I think having a strong warning might be the approach we want to go with. We have it in our privacy policy that we do store in plain text (https://redditenhancementsuite.com/privacy). But for this make it explicitly clear to say "Hey, these are the risks. Proceed with caution". Looking to see if we have a way we can flag a popup on user select to act as the warning.

ElijahPepe:
> Looking to see if we have a way we can flag a popup on user select to act as the warning.

I added a commit to trigger a `window.alert()` using `onChange()`. Hopefully that works.

lib/modules/backupAndRestore.js
```diff
@@ -103,6 +103,34 @@ module.options = {
...
+			key: 'digest',
+			name: 'backupAndRestoreDigest',
+			type: 'boolean',
+			value: false,
+		}],
+		value: ([]: Array<[string, string, string, boolean]>),
+		description: 'backupAndRestoreWebDAVDesc',
+		title: 'backupAndRestoreWebDAVTitle',
+		onChange() {
+			window.alert('RES stores WebDAV credentials in plaintext.');
```
larsjohnsen, line 131: Using `window.alert` in embedded cross-origin frames (which is the case for RES' option page) is [slated to be blocked](https://chromestatus.com/feature/5148698084376576).  As an alternative, you could use the `showNotification` function.

ElijahPepe: Was not aware that the options page was a cross-origin frame. I'll change this ASAP.

PR #5402, opened 2022-05-02
Add Reddit version to submitIssue

ElijahPepe:


larsjohnsen:
Hm, do you know if this still works? The input area isn't a textbox when I use it:
image

ElijahPepe:
I have been talking with @andytuba about new Reddit. IIRC the issue was pinpointed to some React validation, probably detecting a keydown or something like that.

PR #5405, opened 2022-06-04
feat(modules): showImages custom download filename

skyme5:
Relevant issue: 
Tested in browser:


skyme5:
Sorry wrong upstream. 😅

Issue #5344, opened 2021-08-19
[Request] Add expandos to youtube clips.


DavidPHH:
Youtube implemented a clipping feature for videos and livestreams. Would it be possible to add expando support on posts linking to these clips?  
[Example post/clip](https://www.reddit.com/r/LivestreamFail/comments/p7jb29/doc_calls_out_all_of_the_streamers/)

s3b-:
The problem with this is, that youtube doesn't offer the embedding of livestream clips. I checked the documentation page (https://support.google.com/youtube/answer/10332730?hl=en) and it says that you can - but when I tested in on Doc's livestream, it didn't offer me an embed-link, but only a direct link.

import-that:
could this be solved by using the direct link as a source for an iframe?

s3b-:
The direct link links to the actual live stream video. So when embedding the direct link, it will include all the unnecessary stuff like the youtube logo, subscribe button, etc. The video wouldn even be in focus, I guess?

Kryptortio:
> could this be solved by using the direct link as a source for an iframe?

YouTube sets [x-frame-options: SAMEORIGIN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) so unfortunately browsers will not allow it, you would need to [bypass](https://github.com/guilryder/chrome-extensions/tree/main/xframe_ignore) this security feature. Then I suppose you could make the iframe large enough to show the entire page and put it inside a [container that crops](https://jsfiddle.net/wf0sz6h4/) it so that it only shows the video. Would probably be a brittle  solution though (ui changes, add-ons and user preferences might break it), slow (needs to load everything) and maybe a security risk depending on how the bypass is done. So in short, likely not a viable solution.

TopFuel:
> The problem with this is, that youtube doesn't offer the embedding of livestream clips. I checked the documentation page (https://support.google.com/youtube/answer/10332730?hl=en) and it says that you can - but when I tested in on Doc's livestream, it didn't offer me an embed-link, but only a direct link.

Clips can actually be embedded (it works on Discord for example), if you check the meta tags `og:video:url`, `og:video:secure_url` and `twitter:player` on a clip page you can see the embed url for that clip.

Seems like it's a normal video embed url plus the parameters `clip` and `clipt`.
[...]

PR #5417, opened 2022-09-22
Add YouTube clips "expando"

thomasasfk:
- Can now expand and watch YouTube clips
- Supports full embedded player functionality

Image of the new expando option:
image

Video demo to see it in action;
https://user-images.githubusercontent.com/22506439/191857353-ee5bff89-7aef-4488-882c-73b3de544c99.mp4

I see there is an existing old implementation in #5358, which may still work.  It however uses `https://opengraph.io/` and requires an `app_id` query parameter to extract the embed link, in contrast, the implementation in this PR uses `https://web.scraper.workers.dev/` to do the same thing, just without any required authentication.  

With that being said, it is a third-party dependency and could stop working or change it's underlying functionality, it may also end up hitting issues given YouTube's underlying implementation changes, although it is working at this time of raising this PR.

Relevant issue: #5344 
Other implementations: #5358 

### DIFF
--- a/lib/modules/hosts/youtubeclips.js
+++ b/lib/modules/hosts/youtubeclips.js
@@ -13,13 +13,10 @@ export default new Host('youtubeclips', {
 		return clipId;
 	},
 	async handleLink(href, clipId) {
-		const youtubeResponseBody = await ajax({
-			url: `https://web.scraper.workers.dev/?url=https://www.youtube.com/clip/${clipId}&selector=body`,
+		const {iframeUrl} = await ajax({
+			url: `https://youtube-clip-embed-extractor.544146.workers.dev/?clipId=${clipId}`,
 			type: 'json',
 		});
-
-		const [ytInitialPlayerResponse] = youtubeResponseBody.result.body[0].match(/(?<=var ytInitialPlayerResponse = ).*?}(?=;)/);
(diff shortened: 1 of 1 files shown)


corylulu:
This API isn't doing anything other than being a proxy and returning the entire body of the uncached HTML body and searching for the link in the body. You could just as easily call Youtube directly. 

It's also easier to find in the `og:video:url` metadata tag. The reason [embed.ly](https://github.com/reddit-archive/reddit/blob/3db4066f73753546bfb3cdcf8ab06f05443b7636/r2/r2/lib/media.py) was used was because it cached the data so it wouldn't spam calls or have taxing responses. 

I have yet to find a better solution using the APIs so far, however.

thomasasfk:
> https://jsonlink.io/

Hey, good point.  With caching in mind, I forked the original worker repository and changed it to just specifically handle our use-case with caching.  

You can see the project here: https://github.com/thomasasfk/web.scraper.workers.dev
And I've added changed to support it in:

jonas1015119:
is there a reason why this isn't being added?

creesch:
> is there a reason why this isn't being added?

No specific reason that this isn't specifically added, but in general: 
https://www.reddit.com/r/RESAnnouncements/comments/sh83gx/announcement_life_of_reddit_enhancement_suite/

jonas1015119:
ah, thanks. I guess this is isn't something that could just be added via Tampermonkey by users?

PR #5441, opened 2022-12-21
Add fix backtick code block module

### LINKED ISSUE #5223: [Enhancement] support for markdown code blocks on old.reddit.com
So apparently the new reddit has support for code blocks using triple backticks like in regular markdown, but this has not been added on old.reddit.com. Maybe it would be nice to add something like this to RES? Perhaps even with syntax highlighting using something like [highlight.js](https://github.com/highlightjs/highlight.js).

Screenshot_20200519_093532

magnus-ISU:
fixes #5223 

One of the greatest pains to using old reddit ever

### DIFF
--- a/lib/modules/fixCodeBlock.js
+++ b/lib/modules/fixCodeBlock.js
@@ -4,7 +4,9 @@ import { markdown } from 'snudown-js';
 
 import { Module } from '../core/module';
 
-import { watchForThings } from '../utils';
+import { watchForThings, regexes } from '../utils';
+
+import { ajax } from '../environment';
 
 export const module: Module<*> = new Module('myModule');
 
@@ -19,16 +21,44 @@ module.beforeLoad = () => {
 	watchForThings(['post', 'comment', 'message'], rewriteCodeBlocks);
 };
 
-function rewriteCodeBlocks(thing) {
+async function rewriteCodeBlocks(thing) {
+	const path = thing.element.getAttribute('data-permalink');
 	// Find the <div class=md> and prepare to edit it
 	const markdownBody = thing.element.querySelector('.md');
 	const originalMarkdown: string = markdownBody.textContent;
-	// Small optimization? I suspect doing regex splits, then markdown reencoding on every comment will be more taxing than two passes on code comments
 	if (!originalMarkdown.includes('```')) {
 		return;
 	}
+	// Ideally, you would parse something (innerHTML?) and not have to go through the network at all.
+	// Now that we have the response, we can fix triple backticks
+	let sourceText;
+
+	const response = await ajax({
+		url: `${path}.json`,
+		query: { raw_json: 1 },
+		type: 'json',
+	});
+
+	console.log(response);
+
+	if (regexes.commentPermalink.test(path)) {
+		sourceText = response[1].data.children[0].data.body;
+	} else if (regexes.comments.test(path)) {
+		sourceText = response[0].data.children[0].data.selftext;
+	} else {
+		const postId: string = ((/\/(\w*)\/?$/).exec(path): any)[1];
+		const data = response.data.children[0].data;
+		if (data.id === postId) {
+			sourceText = data.body;
+		} else {
+			// The message we want is a reply to a PM/modmail, but reddit returns the whole thread.
+			// So, we have to dig into the replies to find the message we want.
+			sourceText = data.replies.data.children.find(({ data: { id } }) => id === postId).data.body;
+		}
+	}
+
 	// Magic regex magic from magic people on github
-	const reflowedMarkdown = originalMarkdown.replace(
+	const reflowedMarkdown = sourceText.replace(
 		/```([\s\S]+?)```/gm,
 		(_, code) => `\n${code.split('\n').map(l => `    ${l}`).join('\n')}\n`,
 	);
--- a/lib/modules/fixCodeBlock.js
+++ b/lib/modules/fixCodeBlock.js
@@ -19,7 +19,10 @@ module.beforeLoad = () => {
 };
 
 async function rewriteCodeBlocks(thing) {
-	const path = thing.element.getAttribute('data-permalink');
+	const path = thing.element.getAttribute('data-permalink'); // || thing.element.getAttribute('data-url'); // For post pages, but those need a listener on the expando div
+	if (!path) {
+		return;
+	}
 	// Find the <div class=md> and prepare to edit it
 	const markdownBody = thing.element.querySelector('.md');
 	if (!markdownBody) {
@@ -29,11 +32,11 @@ async function rewriteCodeBlocks(thing) {
 	if (!originalMarkdown.includes('```')) {
 		return;
 	}
+
 	// Ideally, you would parse something (innerHTML?) and not have to go through the network at all.
-	// Now that we have the response, we can fix triple backticks
+	// Code copied from sourceSnudown module
 	let sourceText;
 
-	// Code copied from sourceSnudown module
 	const response = await ajax({
 		url: `${path}.json`,
 		query: { raw_json: 1 },
@@ -56,11 +59,16 @@ async function rewriteCodeBlocks(thing) {
 		}
 	}
 
+	// Now that we have the response, we can fix triple backticks
 	// Magic regex magic from magic people on github
 	const reflowedMarkdown = sourceText.replace(
 		/```([\s\S]+?)```/gm,
 		(_, code) => `\n${code.split('\n').map(l => `    ${l}`).join('\n')}\n`,
 	);
+
+	// Turn md to html
 	const htmlMarkdown = markdown(reflowedMarkdown);
+
+	// Change the DOM
 	markdownBody.innerHTML = htmlMarkdown;
 }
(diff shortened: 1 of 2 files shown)


magnus-ISU:
Problems:

 [ ] The regex used changes some things, I will check this out and fix it. See [here](https://www.reddit.com/r/test/comments/zs3zzg/test_markdown_code_block_fix_for_res/?)
 [ ] What is the correct PageType for this to work on? Just `comments` or `commentsLinkList` or `commentsProfilePage` too? I don't know. Should I just remove it?

magnus-ISU:
hmm. The regex wasn't the problem. It works on some things, but not always. textContent is very crappy, maybe parsing the json does make more sense.

serg06:
Hello! I haven't worked with RES before but I'm ready to pitch in.

To clarify the `textContent` issue: You're trying to grab the raw post/comment text so that you can parse it with markdown?

If so: Why don't you grab the raw text from the `textarea` that follows the post?

image

magnus-ISU:
That appears only to exist if you wrote the comment I think? Or doesn't exist on a fresh browser in any case

serg06:
Ah good call. It appears when you click `source`, which grabs it from the json.

Good call about the json, that looks like the right move. Do you have a way to grab it from your code?

magnus-ISU:
Just need to find the correct link, just adding `.json` to the permalink appears to return json for the whole comments page which I don't think is right. The `sourceSnudown` module does some random shit, I'm trying to understand it.

magnus-ISU:
Got it working. I don't know exactly what the code is doing still to be honest, but it works for both comments and posts.

It doesn't work for post previews in a normal subreddit. Unsure why exactly, it seems `watchForThings` doesn't call my function for those at all. But sourcesnudown does show `source` buttons for them.

serg06:
> the permalink appears to return json for the whole comments page

You can add `limit: 1` to your query to limit your results.

---

> Got it working

Nice! How do I test it out, do I need to do the setup from [CONTRIBUTING.md](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/CONTRIBUTING.md)?

magnus-ISU:
Go to my repo, clone it,

```
NODE_OPTIONS=--openssl-legacy-provider yarn start --env browsers=firefox
```

then install the folder from `dist/` as a temporary addon.

Or using chrome, don't pass in `--env browsers=firefox` and figure out how to load the chrome extension

magnus-ISU:
Presently it only works on comment pages. I am investigating and it seems that on a subreddits main page (example: `reddit.com/r/test`) the entry will be visited. At this moment, the text of the post is not loaded yet. So it does not find any markdown container, and it is skipped.

Then when you expand it, in the `div class="expando"` the body of the post is loaded in there via network request. But it does not trigger an update from the extension. The extension should add some sort of listener to those expandos so it can reformat posts on subreddit pages too. But that can be an enhancement, at this moment I am happy for this to be merged.

serg06:
It's working! Woo!

serg06:
It doesn't work for styled code blocks, i.e.:

```
```js
console.log('test');
.``` (ignore the period)
```

Displays as

```
js
console.log('test');
```

When it should display as

```js
console.log('test');
```

magnus-ISU:
Dang. I knew I shouldn't trust regexes from random people which I don't understand

serg06:
Some ideas:

- We could use a markdown syntax parsing library to parse the text, then converting that to snudown format.

- We could use a markdown->html library and skip snudown altogether.

magnus-ISU:
- We could fix the regex. I don't want to add dependencies

serg06:
Any particular reason to avoid dependencies?

magnus-ISU:
Not my project, I'm just a random dude coming into it. If a dependency should be added that would probably have been brought up by more experienced maintainers discussing this.

Anyway, its not the regex's fault, it's that reddit changes 

````
\`\`\` some stuff
and some more
stuff\`\`\`
```

Into `<code>some stuff and some more stuff</code>` an inline code block.

I guess we can make more network requests for any comment/post which includes a `<code>` block at all.

serg06:
Aside: I don't see any reason why we'd need maintainer approval for that; they don't mention it in CONTRIBUTING.md either.

Also this regex should handle styled code blocks: `/```(?:[a-zA-Z0-9]*\n)?([\s\S]+?)```/gm`

magnus-ISU:
The current regex is fine - it accepts anything already. The trouble was as I described above. I pushed - you'll note though that multiple code blocks in a row look like trash. I will see what I can do about that.

magnus-ISU:
Hmm. backticks in four-space code lines are messed up by the regex. I am more and more tempted to just line-by-line parse it as in my first prototype.

serg06:
Nice job on the styled MD fix. Seems to match Github's markdown implementation.

serg06:
If I understand correctly, you'd like this:

```
    ```
    test
    ```
```

to display as

```
test
```
?

magnus-ISU:
Yes. Look at how new reddit renders my comment

serg06:
I see. How about enforcing that backticks must start at the beginning of a line?

`/^```([\s\S]+?)^```/gm`

magnus-ISU:
I do think that would be a good idea. I think I tried that regex in the past, but before anything worked, so it slipped my mind.

magnus-ISU:
Alright. It still does not render everything exactly the same as new reddit, but it seems to be related to behavior of triple backticks becoming single backticks by snudown, not anything I'm doing, which no one uses anyway, so should be fine. And/or is related to new reddit not requiring closing code fence which is a mistake on reddit's part, so I don't care.

serg06:
Not sure - whatever's more consistent is better I guess.

magnus-ISU:
That shouldn't be a code block.

> A [code fence](https://spec.commonmark.org/0.30/#code-fence) is a sequence of at least three consecutive backtick characters (`) or tildes (~). (Tildes and backticks cannot be mixed.) A [fenced code block](https://spec.commonmark.org/0.30/#fenced-code-block) begins with a code fence, preceded by up to three spaces of indentation.

> The line with the opening code fence may optionally contain some text following the code fence; this is trimmed of leading and trailing spaces or tabs and called the [info string](https://spec.commonmark.org/0.30/#info-string). If the [info string](https://spec.commonmark.org/0.30/#info-string) comes after a backtick fence, it may not contain any backtick characters. (The reason for this restriction is that otherwise some inline code would be incorrectly interpreted as the beginning of a fenced code block.)

> The content of the code block consists of all subsequent lines, until a closing [code fence](https://spec.commonmark.org/0.30/#code-fence) of the same type as the code block began with (backticks or tildes), and with at least as many backticks or tildes as the opening code fence. If the leading code fence is preceded by N spaces of indentation, then up to N spaces of indentation are removed from each line of the content (if present). (If a content line is not indented, it is preserved unchanged. If it is indented N spaces or less, all of the indentation is removed.)

> The closing code fence may be preceded by up to three spaces of indentation, and may be followed only by spaces or tabs, which are ignored.

I mean commonmark has so many backwards compatibility rules and weird things anyway that what it says isn't exactly too important, but there is no way that should become a code block.

magnus-ISU:
For example, I do not care that a code fence may be preceded by up to 3 spaces of indentation.

magnus-ISU:
Okay I think I'm happy now, going to wait to see what someone who can merge this thinks. @honestbleeps ?

serg06:
In terms of efficiency: If you visit a thread with 500 comments with code blocks, you'd make 500 ajax requests. That doesn't seem okay.

Step 1: Since the HTTP request returns data for a ton of comments, why don't we cache all that data and re-use it later?

Step 2: We could add an async mutex to prevent request/cache race conditions.

magnus-ISU:
Good point. The http request only returns data for 1 comment though, it is limiting it to 1 somehow which is how it always gets the right one. I think, if that is wrong oops.

But a mutex sounds like a good idea then

serg06:
Iirc the request returns data for a bunch of child comments too, the data is tree shaped.

I'll see if I can take a crack at it later today.

calops:
Is there any hope to see this merged one day? At this point I'd be OK with the performance hit if it's an opt-in feature. Threads with a huge amount of code blocks are rare anyway, and they won't all be loaded at the same time.

JuhaJGamer:
> Is there any hope to see this merged one day? At this point I'd be OK with the performance hit if it's an opt-in feature. Threads with a huge amount of code blocks are rare anyway, and they won't all be loaded at the same time.

Please. It's been two years. Add it as an opt-in, completely gut it and just transform any triple-backticks to code tags, I don't care, I have infinite data. The code is right there, it works, it's inefficient, but I want it on my computer without having to custom build RES.

BenMcGarry:
This PR is incomplete, if you want it merged feel free to raise a new one with the actions completed.

calops:
Is the only pending issue the one about performance (sending too many ajax requests)? If so, I'm willing to work on this.

magnus-ISU:
I stopped using reddit for a while. I suspect fable can one-shot this, so I guess I'll try to get it fixed up.

PR #5444, opened 2022-12-23
Include captions on i.redd.it and v.redd.it media

brookst:
Relevant issue: https://www.reddit.com/r/RESissues/comments/zqvuis/inline_image_viewer_missing_text/
Tested in browser: Google chrome 108.0.5359.125 (Official Build) (64-bit)

Not really a web dev so I may well have done something stupid here. This seems to bring in the Reddit selftext correctly and still looks right if there's no selftext. I'm using selftext_html which seems to double up on formatting, but includes anchors for links such as on the post in the issue.

Would appreciate any feedback on this but hopefully it's a clean fix.

### DIFF
--- a/lib/modules/hosts/ireddit.js
+++ b/lib/modules/hosts/ireddit.js
@@ -15,7 +15,7 @@ export default new Host('ireddit', {
 		if (preview.variants.mp4) {
 			return {
 				type: 'VIDEO',
-				caption: postMetadata.selftext_html.replace(/<\/?p>/g, ''),
+				caption: postMetadata.selftext_html && postMetadata.selftext_html.replace(/<\/?p>/g, ''),
 				loop: true,
 				muted: true,
 				fallback: preview.variants.gif && preview.variants.gif.source.url,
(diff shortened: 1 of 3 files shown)


brookst:
The formatting weirdness seems to be due to `.res-caption` being styled with `white-space: pre-wrap` [here](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/70c22af5f4af4fbff19e8484d90e20deb6f72a68/lib/css/modules/_showImages.scss#L264). The main issue is blocks of text picking up a wrapping `<p></p>` somewhere. There are embedded newlines that don't have a matching `<br>` so stripping newlines appears worse. Just stripping out p tags seems to work for everything I could see.

E.g: [This post](https://www.reddit.com/r/Alienware/comments/zspw0p/project_area_51_r2/) has an embedded list with newlines, plus the wrapping p tags.

brookst:
Pleased with getting yarn to actually build RES, I totally missed the contributor guidance on testing and linting. A run of `yarn flow check --show-all-errors && yarn lint` is clean now. `yarn test` also passes though I don't have selenium and a driver set up to do the integration tests. Everything seems fine in browser though.

larsjohnsen:
Thanks!

PR #5465, opened 2023-04-21
Add module "Submission Shortcuts"

oSumAtrIX:
## This PR adds the module "Submission Shortcuts":

Moderation is tedious, especially if you rely on the submission buttons to approve, remove or mark submissions as spam. For that reason, this module has been added to ease moderation with shortcuts:

   image

### Note

- This PR does not consider the alphabetic order in the locales file.
- There is no extensive use of existing utilities of RES but instead of regular browser features (such as querySelector)
- The feature does not work outside of submissions on feeds
- This module does not specify on moderation, but shortcuts for moderation have been implemented for the initial release.
- Tested in browser Microsoft Edge

### DIFF
--- a/lib/modules/submissionShortcuts.js
+++ b/lib/modules/submissionShortcuts.js
@@ -1,0 +1,51 @@
+/* @flow */
+import $ from 'jquery';
+import { Module } from '../core/module';
+import { checkKeysForEvent, click } from '../utils';
+
+export const module: Module<*> = new Module('submissionShortcuts');
+
+module.moduleName = 'submissionShortcutsName';
+module.category = 'submissionsCategory';
+module.description = 'submissionShurtcutsDesc';
+module.options = {
+	markSpammed: {
+		type: 'keycode',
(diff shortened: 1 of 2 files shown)


lib/modules/submissionShortcuts.js
```diff
@@ -0,0 +1,51 @@
...
+		mustBeLoggedIn: true,
+		buttonSelector: '.icon-approve',
+	},
+
+};
+
+module.contentStart = () => {
+	$(document.body).on('keydown', (e: KeyboardEvent) => {
+		Object.values(module.options).forEach(shortcut => {
+			if (checkKeysForEvent(e, shortcut.value)) {
```
oSumAtrIX, line 42:
```suggestion
	const optionValues = Object.values(module.options);

	$(document.body).on('keydown', (e: KeyboardEvent) => {
		optionValues.forEach(shortcut => {
			if (checkKeysForEvent(e, shortcut.value)) {
```

PR #5475, opened 2023-09-03
Update showImages.js to fix google reverse image search URL

tompreuss:
Google added an additional `sbisrc` parameter to continue to access the non-Lens type of reverse image search


tompreuss:
Due to my submission of this pull request, I've been notified of the "review" by @Playaset above.

Per https://github.com/json-schema-org/website/issues/1116#issuecomment-2477593333 and my own look around their profile, their only contribution is spam and their nonsensical participation here should be disregarded.

PR #5488, opened 2024-01-12
Replace webpack with esbuild

larsjohnsen:
This PR removes the Webpack build system, replacing it with the simpler esbuild.

The main difference from the webpack setup is how resources are loaded. The build process is initiated by `build.js`, which prepares manifests, distributed files, and bundling the javascript and CSS. Resources must now be referenced explicitly where used (mainly references in `manifest.json`). Additionally, all module files and media hosts must now be explicitly referenced.

Some of the dev dependencies are removed, like `nyc`, `postcss`, and `autoprefixer` since they apparently are not needed anymore. The main `babel` dependency is kept due to`eslint` requiring it when the code is flowtype annotated, although it is not used in the main build process.

All other dependencies except for `flow-bin` and `nightwatch` are now up to date. These requires a fair amount of work to get up to date.

nodejs 21 is required to build this.

Tested in browser: Firefox 121, Chrome 122


BenMcGarry:
Can we remove nightwatch? I think saucelabs stopped working for us years ago?

larsjohnsen:
> Can we remove nightwatch? I think saucelabs stopped working for us years ago?

The test suite is still valuable and can be run locally, though many of the tests are now failing due to changes in the browsers' control interface. Probably not too hard to fix for someone familiar with Nightwatch / webdriver, so I think we should keep it.

BenMcGarry:
Is the blocker for updating nightwatch our setup? Guessing we need to re-write it all.

larsjohnsen:
From what I could figure out from attempting to run the tests locally: some changes in `nightwatch.conf.js` would be required and some of the methods used in the tests are removed. This seems to be completely independent of the build system, however.

BenMcGarry:
We still have free access to browserstack as a OSS project, i'll take this as something for me to do to try move to it. From a quick search I think we can quickly flip our tests over.

BenMcGarry:
Whats blocking on flow-bin?

larsjohnsen:
> Whats blocking on flow-bin?

Upgrading past flow-bin 84 [requires a lot of additional annotation](https://medium.com/flow-type/asking-for-required-annotations-64d4f9c1edf8).

BenMcGarry:
After this i'll try mv3 again, as i think this may make it easier to get that all updated.

BenMcGarry:
Is this all good from your side @larsjohnsen ?

larsjohnsen:
Yes, it appears to be running fine. I've mainly been running this in Firefox though.

PR #5502, opened 2024-03-22
Use commentDepth options in Follow Comments keyboard navigation

khaister:
Relevant issue: N/A
Tested in browser: Chrome

`commentDepth` options only supported when navigating to comments link by clicking on it. This PR supports commentDepth options when navigating to comments using keyboard shortcut as well (if defined under _Browsing > Keyboard Navigation > Follow Comments_)

### DIFF
--- a/lib/modules/keyboardNav.js
+++ b/lib/modules/keyboardNav.js
@@ -1,17 +1,20 @@
 /* @flow */
 
 import $ from 'jquery';
-import { once, maxBy, without, memoize } from 'lodash-es';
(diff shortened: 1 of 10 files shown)


BenMcGarry:
Can you please remove the language additions to all languages except for en.json. As the build process replicates to all others.

khaister:
> Can you please remove the language additions to all languages except for en.json. As the build process replicates to all others. I'm also not sure what this is trying to solve/fix - can you detail what this change is some more please?

@BenMcGarry done 👍

Issue #5511, opened 2024-03-29
MV3 CSP Blocking Reddit media player


BenMcGarry:
`Refused to load the script 'https://www.redditstatic.com/videoplayer.XCrwE8Bi5A4.js' because it violates the following Content Security Policy directive: "script-src 'self' 'wasm-unsafe-eval' 'inline-speculation-rules'". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.`

https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/utils/pageContextScript.js

Stack trace points here: https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/580e8eafa9bdd1d4bd2ff9db1cd2f6e53c783a51/lib/modules/showImages.js#L460-L476

BenMcGarry:
Appears https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/lib/modules/hosts/vreddit.js is pulling in a external script and MV3 doesnt allow that anymore.

larsjohnsen:
No, but it it is added directly via `build.js`.

The `onInit` in showImages block had, as far as I can remember, no side effects other than some performance improvements, so I think we might as well remove it.

BenMcGarry:
Removing that `onInit `block may have fixed it... Not able to repro it anymore and no CSP error. It looks like Reddits advert block might have been messing with it.

BenMcGarry:
Looking at https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/507c0b2e89a839a9c08fff90c28f3dcf826b2a03

I think its because theyve started to run ads through the same domain with a different player.

PR #5512, opened 2024-03-29
Remove showImages onInit

BenMcGarry:
Appears to resolve https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/5511

Tested in Chrome and Firefox.

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -18,14 +18,12 @@ import type {
 	GenericMedia,
 } from '../core/host';
 import { Host } from '../core/host';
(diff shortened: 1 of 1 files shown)


BenMcGarry:
@larsjohnsen as this is rolling back your original commit somewhat (https://github.com/honestbleeps/Reddit-Enhancement-Suite/commit/507c0b2e89a839a9c08fff90c28f3dcf826b2a03) - can you see any issues with this? From testing in Chrome/Firefox seems to fix the issue with no obvious issues.

PR #5521, opened 2024-05-02
showImages: Fix audio being played in the background

### LINKED ISSUE #5519: Video expandos now break in comments causing double audio
I cant reproduce but: https://www.reddit.com/r/RESissues/comments/1cha7px/videos_autoplay_despite_expando_being_collapsed/

Cause: https://github.com/honestbleeps/Reddit-Enhancement-Suite/pull/5512

larsjohnsen:
It seems like the issue resolved in #5512 did more than just improving performance. In some cases it also prevented Reddit's video player from being initialized in the background, causing audio being played.

This PR reworks the reverted code by caching the setting value in localStorage and thereby letting us avoid having the `undo` logic (which is incompatible with manifest V3).

Relevant issue: Fixes #5519 
Tested in browser: Chrome 124

### DIFF
--- a/lib/modules/showImages.js
+++ b/lib/modules/showImages.js
@@ -18,12 +18,14 @@ import type {
 	GenericMedia,
 } from '../core/host';
 import { Host } from '../core/host';
+import { loadOptions } from '../core/init';
 import { Module } from '../core/module';
 import {
 	positiveModulo,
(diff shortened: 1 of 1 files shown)


BenMcGarry:
@larsjohnsen are you able to lint?

PR #5536, opened 2024-09-14
showImages/hosts: Update Imgur

larsjohnsen:
- Fixes video retrieval (the API no longer contains the property `gifv`)
- Removes option `useGifOverGifv` which seemingly is broken
- Removes redundant protocol replacement, as Imgur now always responds with https

Relevant issue:  https://www.reddit.com/r/RESissues/comments/1exm5b7/imgur_gifgifvmp4_expandos_broken/
Tested in browser: Chrome 128

### DIFF
--- a/lib/modules/hosts/imgur.js
+++ b/lib/modules/hosts/imgur.js
@@ -15,12 +15,6 @@ export default new Host('imgur', {
 			value: true,
 			type: 'boolean',
 		},
-		useGifOverGifv: {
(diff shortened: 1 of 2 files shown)


benmcgarry:
This good to merge? If so can and i'll push a release.

larsjohnsen:
> This good to merge? If so can and i'll push a release.

Yes, it seems good.

Issue #5531, opened 2024-08-07
Expando permission prompts do not appear when 1Password is installed on Chromium


imagebird:
- My browser (copy from https://www.whatismybrowser.com): Chrome 127 on Windows 19
- My RES version: v5.24.6
- Am I in reddit beta (yes/no): no

*What's the problem?*

When expanding embeds, permission prompts do not appear when 1Password's extension is enabled.

*Are there any related posts in https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues or /r/RESissues?*

https://old.reddit.com/r/RESissues/comments/1eeeo76/cant_grant_expando_permissions_in_chromium_browser/

*What other browser extensions are installed?*

![image](https://github.com/user-attachments/assets/61ae3cc9-60fb-4fad-8a55-77a00c5855e8)

yyli:
I can confirm that I have the same issue. If I disable 1password the expando permission prompt shows up.

benmcgarry:
This appears to be a 1PW issues, i've reached out to them.

jef:
Just confirmed this information as well. Thanks for reaching out to them @benmcgarry; hopefully we'll get an update!

benmcgarry:
As an update to this, 1Password have confirmed its an issue on their side. However they have given no timeframe for it to be fixed or any commitment that it will be fixed. 

Will keep this open to track with whatever happens. Going to see about adding a dialog to detect 1PW and throwing an alert telling them to temporarily disable it for RES to work fully.

ncla:
Do you know if there is a way to track this issue on 1Password side (forum, issue tracker etc.)?

My team are developing our own browser extension and experienced permissions pop-ups not appearing after calling [browser.permissions.request](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/request). I am MacOS and Brave myself, and I use 1Password browser extension as well, and after disabling 1Password extension, the permissions pop-ups appeared after _user click_.

A workaround I found is if you can navigate to the button that triggers the permission request with just keyboard, by pressing TAB and SPACE button, you can trigger the browser permission modal that way.

It's strange that this is caused by 1Password, but reading the WebExtension docs I am thinking this is something to do with what browser considers an [user action](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/User_actions).

benmcgarry:
[...]

PR #5538, opened 2024-09-16
environment/permissions: Fix permission popup not showing

larsjohnsen:
`window.screen` is not available in manifest v3, causing the prompt window fallback method to fail. This uses the web extension API to find the coordinates to place the popup.

The fallback method is ordinarily not needed in Chrome, but other extensions may interfere (see #5531) Chrome's detection of user initiated permission requests.

Relevant issue: #5531 
Tested in browser: Firefox 130, Chrome 128

### DIFF
--- a/lib/environment/background/permissions.js
+++ b/lib/environment/background/permissions.js
@@ -24,9 +24,11 @@ async function makePromptWindow({ permissions, origins }) {
 
 	const width = 630;
 	const height = 255;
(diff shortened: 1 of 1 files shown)


benmcgarry:
lmk if good to merge and will do a release this weekend.

larsjohnsen:
> lmk if good to merge and will do a release this weekend.

Should be good to go!

PR #5556, opened 2025-01-09
Fix: reddit poll expando redirects to comments

Luc-Mcgrady:
Relevant issue: Poll expandos currently redirect to the comments.
image

Tested in browser: Firefox


arye321:
polls are broken again. 

maybe should consider redirecting to sh.reddit.com

https://sh.reddit.com/r/polls/comments/1j0bpq3/you_cant_change_country_or_class_who_would_you/

benmcgarry:
Sh.Reddit is in the process of being removed as a endpoint. So the same thing will happen. Unfortunately this looks to be a change out of RES control and most likely won't look to fix further.

PR #5558, opened 2025-01-22
widget edit cancel button fix

### LINKED ISSUE #5557: Widget editing cancel button text is changeable
If you focus the cancel button when editing a widget without clicking it, e.g. by clicking the display name field and pressing tab twice, the text can be changed and its value will be saved and used for all widgets' cancel buttons if save or even the cancel button itself is clicked.

The button's input type is set to "cancel", which is not a valid type. Changing it to button fixes the issue. This also causes some of its CSS to inherit other values which overrides its intended look, but can be fixed with !important.

shanktank:
Relevant issue: 5557
Tested in browser: Firefox

Fix #5557 by changing input type to button

### DIFF
--- a/lib/css/modules/_dashboard.scss
+++ b/lib/css/modules/_dashboard.scss
@@ -208,14 +208,14 @@
 .cancelButton {
 	width: 50px;
 	text-align: center;
-	cursor: pointer;
+	cursor: pointer !important;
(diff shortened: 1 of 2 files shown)


shanktank:
Problem also applies to Chrome and is fixed by these changes

PR #5560, opened 2025-01-23
Feature: Add support for nitter expandos

FlaminSarge:
Relevant issue: 
Tested in browser: Chrome

This goes directly to the given nitter instance instead of using the existing twitter/x handling.
Only includes the 2 nitter instances that seem to allow embeds.

Remaining in draft until discussion on whether to even do this concludes.

### DIFF
--- a/lib/modules/hosts/index.js
+++ b/lib/modules/hosts/index.js
@@ -39,6 +39,7 @@ import makeameme from './makeameme';
 import memecrunch from './memecrunch';
 import memedad from './memedad';
 import navertv from './navertv';
+import nitter from './nitter';
 import onedrive from './onedrive';
 import pastebin from './pastebin';
 import peertube from './peertube';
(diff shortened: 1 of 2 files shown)


FlaminSarge:
Not sure if this should even be included given how liable the hosting services are to break...

benmcgarry:
Not entirely against it, but with everything happening on Reddit right now im wary of flooding an endpoint with requests (we've done this before and got blocked) - How stable is this?

FlaminSarge:
> How stable is this?

From my understanding: not stable at all at the moment. privacydev.net returns 429s after like the 2nd request, and poast.org load-loops. If xcancel or lightbrd were on board, those would be better, but they don't allow being used in iframes at all in their CSP.
Other instances: https://github.com/zedeus/nitter/wiki/Instances

However, under the assumption that folks will move away from Twitter links and at worst use these alternatives, it helps to be prepared with this draft PR.

benmcgarry:
Yeah, Probably best to sit on this for a week or two and see if more appear that are stable and see how it goes. We have the release with Bluesky rolling out now so if this picks up we can quickly push one out.

PR #5577, opened 2025-09-10
Switch to using "all" instead of "unread" as the default inbox page

MonterraByte:
Reddit broke the inbox by making every message be marked as read automatically, so the unread page is now useless. This PR makes the inbox button take you to the barely useful inbox page instead.


MonterraByte:
Seems like the inbox is no longer broken, I'll close this for now.

